PR #20, opened 2021-10-29
[Port] A proposal of an all_simple_paths function implementation

etiennedeg:
this is a port of [#1540](https://github.com/sbromberger/LightGraphs.jl/pull/1540)

### DIFF
--- a/src/traversals/allsimplepaths.jl
+++ b/src/traversals/allsimplepaths.jl
@@ -1,0 +1,185 @@
+using DataStructures
+
+"""
+    all_simple_paths(g, source, targets, cutoff=nothing)
+
+Returns an iterator that generates all simple paths in the graph `g` from `source` to `targets`.
+If `cutoff` is given, the paths' lengths are limited to equal or less than `cutoff`.
+Note that the length of a path is defined as the number of edges, not the number of elements.
+ex. the path length of `[1, 2, 3]` is two.
+Internally, a DFS algorithm is used to search paths.
+
+# Examples
+
+```jldoctest
+julia> using LightGraphs
+julia> g = complete_graph(4)
+julia> collect(all_simple_paths(g, 1, [4]))
+5-element Array{Array{Int64,1},1}:
+ [1, 4]
+ [1, 3, 4]
+ [1, 3, 2, 4]
+ [1, 2, 4]
+ [1, 2, 3, 4]
+ ```
+"""
+function all_simple_paths(g::AbstractGraph, source::T, targets::Vector{T}; cutoff::Union{Int,Nothing}=nothing) where T <: Integer
+    return SimplePathIterator(g, source, targets, cutoff=cutoff)
+end
+
+
+"""
+    all_simple_paths(g, source, target, cutoff=nothing)
+
+This function is equivalent to `all_simple_paths(g, source, [target], cutoff)`.
+This is provided for convenience.
+
+See also `all_simple_paths(g, source, targets, cutoff)`.
+"""
+function all_simple_paths(g::AbstractGraph, source::T, target::T; cutoff::Union{Int,Nothing}=nothing) where T <: Integer
+    return SimplePathIterator(g, source, [target], cutoff=cutoff)
+end
+
+
+"""
+    SimplePathIterator{T <: Integer}
+
+Iterator that generates all simple paths.
+The iterator holds the condition specified in `all_simple_path` function.
+"""
+struct SimplePathIterator{T <: Integer}
+    g::AbstractGraph
+    source::T  # Starting node
+    targets::Set{T}  # Target nodes
+    cutoff::Union{Int,Nothing}  # Max length of resulting paths
+
+    function SimplePathIterator(g::AbstractGraph, source::T, targets::Vector{T}; cutoff::Union{Int,Nothing}=nothing) where T <: Integer
+        new{T}(g, source, Set(targets), cutoff)
+    end
+end
+
+
+"""
+    SimplePathIteratorState{T <: Integer}
+
+SimplePathIterator's state.
+"""
+mutable struct SimplePathIteratorState{T <: Integer}
+    stack::Stack{Vector{T}}  # Store child nodes
+    visited::Stack{T}  # Store current path candidate
+    queued_targets::Vector{T}  # Store rest targets if path length reached cutoff.
+    function SimplePathIteratorState(spi::SimplePathIterator{T}) where T <: Integer
+        stack = Stack{Vector{T}}()
+        visited = Stack{T}()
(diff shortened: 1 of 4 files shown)


jpfairbanks:
What about making the default cutoff smaller? typemax(T) is just completely impractical right? Even in 32 bit int that is ~10^10. What about min(typemax(T), nv(g)^3) or something?

Also, I think we should throw an exception if you hit the cutoff. That way people know that they should either increase the cutoff and try again or give up on enumerating all the paths in the graph. Silently giving an incomplete enumeration isn't the right answer.

jpfairbanks:
@etienneINSA, @matbesancon, Any thoughts on raising an exception if we don't yield all the simple paths? I've never needed all the simple paths, so I don't really know the application context, but I would expect that if I got a partial result, it would come with some return code I could check, or exception that I needed to catch.

src/traversals/allsimplepaths.jl
```diff
@@ -0,0 +1,188 @@
...
+        push!(stack, [spi.source, 1])  # Add a child node with index = 1
+        new{T}(stack, visited, queued_targets)
+    end
+end
+
+"""
+    function _stepback!(state)
+
+A helper function that updates iterator state.
+For internal use only.
```
matbesancon, line 85:
```suggestion
A helper function that updates the simple path iterator state.
```
Internal use is already implied by the underscore

matbesancon:
> Any thoughts on raising an exception if we don't yield all the simple paths?

When does this happen? I didn't see this path in the code?

gdalle:
Related to #106

simonschoelly:
What is the current status of this PR? I.e. who's turn is is to review/write code?

gdalle:
> What is the current status of this PR? I.e. who's turn is is to review/write code?

I was not aware we had defined turns, I thought it was more of a "whoever has time does their bit". As far as I'm concerned, I need to hand in my thesis manuscript in a month so i'm out of the equation for now

simonschoelly:
> > What is the current status of this PR? I.e. who's turn is is to review/write code?
> 
> I was not aware we had defined turns, I thought it was more of a "whoever has time does their bit". As far as I'm concerned, I need to hand in my thesis manuscript in a month so i'm out of the equation for now

No of course there is no formal process, I was just trying to understand where we are :) 

Good look with your manuscript,, I will try to avoid pinging you any further then for now.

gdalle:
> Good look with your manuscript,, I will try to avoid pinging you any further then for now.

On the contrary, do ping me when you need a quick opinion, it's the only way you'll get my attention ^^ But I won't be able to contribute lengthy reviews or code until the summer has elapsed

mschauer:
@samurai-kant We need this

src/traversals/allsimplepaths.jl
```diff
@@ -0,0 +1,188 @@
...
+"""
+    SimplePathIterator{T <: Integer}
+
+Iterator that generates all simple paths.
+The iterator holds the condition specified in `all_simple_path` function.
+"""
+struct SimplePathIterator{T <: Integer}
+    g::AbstractGraph
+    source::T  # Starting node
+    targets::Set{T}  # Target nodes
```
mschauer, line 53: This doesn't have to be a `Set`, right? Maybe just make it `targets::T`

mschauer:
By using outneighbors, does this work for directed and undirected graphs?

i-aki-y:
Hi, I am the author of the original implementation.
Thank you for being so interested in my PR. I recently found this project!

I will answer some questions.

@mschauer
> By using outneighbors, does this work for directed and undirected graphs?

Yes. See the example below.

```jl
println("undir:", collect(all_simple_paths(cycle_graph(4), 1, [3])))
println("dir:", collect(all_simple_paths(cycle_digraph(4), 1, [3])))
> undir[[1, 2, 3], [1, 4, 3]]
> dir[[1, 2, 3]]
```

@jpfairbanks
> What about making the default cutoff smaller? typemax(T) is just completely impractical right? Even in 32 bit int that is ~10^10. What about min(typemax(T), nv(g)^3) or something?

The resulting paths do not have any loop. So the `cutoff` upper bound is: `nv(g)`, the paths that visit all nodes. Here, I intended to set no limit for the path length by using `typemax`. Note that the `cutoff` does not mean the number of outputs but the maximum length of each individual path, in other words, the max depth of the search path.

> Also, I think we should throw an exception if you hit the cutoff. That way people know that they should either increase the cutoff and try again or give up on enumerating all the paths in the graph. Silently giving an incomplete enumeration isn't the right answer.

Would you clarify the meaning of "incomplete enumeration" in your comment?
If a user set `cutoff=c`, the user expects the paths whose lengths are less than or equal to `c`, and the function returns all paths that satisfy the condition. In that sense, the enumeration is complete, I think. 
If the user wants to limit the iteration count and restart it afterward, the user can use `Iterators.Stateful` and `Iterators.take`.
See:

```jl
g = complete_graph(5)
path_iter = Iterators.Stateful(all_simple_paths(g, 1, 2))
println("first 3:", collect(Iterators.take(path_iter, 3)))
println("rests:", collect(path_iter))
> first 3:Any[[1, 2], [1, 3, 2], [1, 3, 4, 2]]
> rests:Any[[1, 3, 4, 5, 2], [1, 3, 5, 2], [1, 3, 5, 4, 2], [1, 4, 2], [1, 4, 3, 2], [1, 4, 3, 5, 2], [1, 4, 5, 2], [1, 4, 5, 3, 2], [1, 5, 2], [1, 5, 3, 2], [1, 5, 3, 4, 2], [1, 5, 4, 2], [1, 5, 4, 3, 2]]
```

If you want to increase the depth gradually, we need to implement a breadth-first path search. It is possible, but I think the bfs requires much more memory footprint.

Please let me know if there is anything I can do to help you move forward with your review.

src/traversals/allsimplepaths.jl
```diff
@@ -0,0 +1,188 @@
+using DataStructures
+
+"""
+    all_simple_paths(g, source, targets, cutoff)
```
mschauer, line 4:
```suggestion
    all_simple_paths(g, source, targets; cutoff)
```

mschauer:
Thank you, and sorry that this got stalled, @i-aki-y  . I think this is fine, setting the cutoff to typemax(Int) just means there is no cut-off which is a reasonable default even if unsuitable for large graphs.

mschauer:
I have no push access, can't proceed here (e.g. resolving the conflicts).

gdalle:
It's weird, cause as a member of JuliaGraphs you should be able to work on this PR. Maybe @etiennedeg can give you access to his branch?

gdalle:
Superseded by #353

PR #22, opened 2021-10-29
[Port] fix induced_subgraph indexing with vector of Bools

etiennedeg:
This is a port of [#1573](https://github.com/sbromberger/LightGraphs.jl/pull/1573)

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -659,6 +659,10 @@ function induced_subgraph(g::T, vlist::AbstractVector{U}) where T <: AbstractGra
     return h, vmap
 end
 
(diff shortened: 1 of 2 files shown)


simonschoelly (approved):
I am merging this now. This PR would probably create some issues with graphs that use `Bool` as eltype, but then such a graph does not really make sense anyway.

Another thing that could go wrong has to do with the method signature, as it is if of the form 
```julia
induced_subgraph(g::AbstractGraph, vlist::AbstractVector{Bool})
```
so I guess it could lead to some issues with some downstream packages that implement
```julia
induced_subgraph(g::SomeSpecificGraph, vlist)
```
but I think in such a case, we could just try to figure out a better method signature later.

PR #24, opened 2021-10-29
[Port] Have added Wilson's algorithm and loop erased random walks

etiennedeg:
this is a port of [#1576](https://github.com/sbromberger/LightGraphs.jl/pull/1576)

### DIFF
--- a/src/traversals/randomwalks.jl
+++ b/src/traversals/randomwalks.jl
@@ -167,7 +167,7 @@ function loop_erased_randomwalk(
         else
             cur_pos += 1
             if length(visited) < cur_pos
-                resize!(visited, max(2*cur_pos, nv(g)))
+                resize!(visited, min(2*cur_pos, nv(g)))
             end
             visited[cur_pos] = v
             visited_view = view(visited, 1:cur_pos)


jpfairbanks:
@nassarhuda, are you familiar with Wilson's algorithm? Could you take a look here?

src/spanningtrees/wilson.jl
```diff
@@ -0,0 +1,58 @@
...
+e.g.
+
+tree = wilson_rst(g)
+probability_of_tree = prod([weights(g)[src(e), dst(e)] for e in tree])
+probability_of_tree /= det(laplacian_matrix(g)[2:nv(g), 2:nv(g)]))
+"""
+function wilson_rst end
+@traitfn function wilson_rst(g::AG::(!IsDirected),
+    distmx::AbstractMatrix{T}=weights(g);
+    seed::Int=-1,
```
simonschoelly, line 23: Do we really need a seed, when we can pass in a random generator?

src/spanningtrees/wilson.jl
```diff
@@ -0,0 +1,58 @@
...
+The exact probability of the tree may be determined by computing the 
+normalization constant of the distribution via [Kirchoff's](https://en.wikipedia.org/wiki/Kirchhoff%27s_theorem)
+or [Tutte's Matrix Tree](https://personalpages.manchester.ac.uk/staff/mark.muldoon/Teaching/DiscreteMaths/LectureNotes/MatrixTreeProof.pdf)
+theorem, which is equivalent to the determinant of any minor of the Laplacian matrix.
+
+e.g.
+
+tree = wilson_rst(g)
+probability_of_tree = prod([weights(g)[src(e), dst(e)] for e in tree])
+probability_of_tree /= det(laplacian_matrix(g)[2:nv(g), 2:nv(g)]))
```
simonschoelly, line 18: Can we use a `julia` or `jldoctest` block here?

src/spanningtrees/wilson.jl
```diff
@@ -0,0 +1,58 @@
...
+    seed::Int=-1,
+    rng::AbstractRNG=GLOBAL_RNG
+) where {T <: Real, U, AG <: AbstractGraph{U}}
+
+    if seed >= 0
+        rng = getRNG(seed)
+    end
+
+    start1 = rand(rng, 1:nv(g))
+    start2 = rand(rng, 1:nv(g))
```
simonschoelly, line 32:
```suggestion
    start1 = rand(rng, vertices(g))
    start2 = rand(rng, vertices(g))
```
might make more sense, although at least for `SimpleGraph`, this will not be of type `Int` but of type `eltype(g)`

src/traversals/randomwalks.jl
```diff
@@ -118,3 +118,64 @@ function self_avoiding_walk(g::AG, s::Integer, niter::Integer; seed::Int=-1) whe
...
+
+Return a vector of vertices visited in order.
+"""
+function loop_erased_randomwalk(
+    g::AG, s::Integer, 
+    niter::Integer=max(100, nv(g)^2);
+    distmx::AbstractMatrix{T}=weights(g),
+    f::Union{Set{V},Vector{V}}=Set{Integer}(), 
+    seed::Int=-1,
+    rng::AbstractRNG=GLOBAL_RNG
```
simonschoelly, line 142: Seem here, do we need `seed`?

src/traversals/randomwalks.jl
```diff
@@ -118,3 +118,64 @@ function self_avoiding_walk(g::AG, s::Integer, niter::Integer; seed::Int=-1) whe
...
+)::Vector{Int} where {T <: Real, U, AG <: AbstractGraph{U}, V <: Integer}
```
simonschoelly, line 143: If I remember `::SomeType` after a function does actually force `convert` to be used on the result of that function.

simonschoelly:
@etiennedeg @gjh6 I think this is something useful to have. Given that this PR has been lying dormant for quite a while, there might be some things that you would do differently? Then maybe it is better, if you would refactor this before me or someone else continues with reviewing.

src/spanningtrees/wilson.jl
```diff
@@ -0,0 +1,58 @@
...
+
+    walk = loop_erased_randomwalk(g, start1, distmx=distmx, f=[start2], rng=rng)
+
+    tree = SimpleGraph(nv(g))
```
etiennedeg, line 36: `tree` is only used to together with `add_edge!`, which is a costly operation. It is more efficient to directly store the unique edges.

src/spanningtrees/wilson.jl
```diff
@@ -0,0 +1,58 @@
...
+    for i = 1:length(walk)-1
+        add_edge!(tree, walk[i], walk[i+1])
+    end
+
+    visited_vertices = Set(walk)
+    unvisited_vertices = setdiff(Set([i for i = 1:nv(g)]), visited_vertices)
```
etiennedeg, line 42: It is more efficient to just flag visited vertices (with a linear pass over `walk`) than to do complex set operations (which I think are not even linear...). We just maintain a `BitVector` `is_visited`of size `nv(g)`.
In `loop_erased_randomwalk`, `e in f` becomes `is_visited[e]`, which is also more efficient.

etiennedeg, line 42:
```suggestion
    unvisited_vertices = setdiff(Set(vertices(g)), visited_vertices)
```

Krastanov:
This has bitrotten quite a bit since last time it was tackled. I will mark it as a draft to keep my review queue a bit more organized, but do not hesitate to convert it back to not a draft if you would like to engage with it again.

PR #28, opened 2021-10-29
[Port] New s_metric method

etiennedeg:
This is a port of [#1548](https://github.com/sbromberger/LightGraphs.jl/pull/1548)

### DIFF
--- a/src/community/s_metric.jl
+++ b/src/community/s_metric.jl
@@ -1,5 +1,5 @@
 """
-    s_metric(g;norm=true)
+    s_metric(g; norm=true)
 
 Return the normalised s-metric of `g`.
 
@@ -21,10 +21,10 @@ julia> s_metric(star_graph(4))
 ```
 """
 
-function s_metric(g::AbstractGraph{T};norm=true) where T
+function s_metric(g::AbstractGraph{T}; norm=true) where T
     s = zero(T)
     for e in edges(g)
-        s += outdegree(g,src(e)) * indegree(g,dst(e))
+        s += outdegree(g, src(e)) * indegree(g, dst(e))
     end
     if norm
         sm = sum(degree(g).^3)/2
--- a/test/community/s_metric.jl
+++ b/test/community/s_metric.jl
@@ -3,19 +3,19 @@ using Random, Statistics
 @testset "S-metric" begin
     @testset "Directed ($seed)" for seed in [1, 2, 3], (_n, _ne) in [(14, 18), (10, 22), (7, 16)]
         g = erdos_renyi(_n, _ne; is_directed=true, seed=seed)
-        sm = s_metric(g,norm=false)
-        sm2 = sum([outdegree(g,src(d)) * indegree(g,dst(d)) for d in edges(g)])
+        sm = s_metric(g, norm=false)
+        sm2 = sum([outdegree(g, src(d)) * indegree(g, dst(d)) for d in edges(g)])
         @test @inferred sm ≈ sm2
-        sm = s_metric(g,norm=true)
+        sm = s_metric(g, norm=true)
         sm2 /= sum(degree(g).^3)/2
         @test @inferred sm ≈ sm2
     end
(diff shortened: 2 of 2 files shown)


src/community/s_metric.jl
```diff
@@ -0,0 +1,35 @@
...
+```jldoctest
+julia> using LightGraphs
+
+julia> s_metric(star_graph(4))
+0.6
+```
+"""
+
+function s_metric(g::AbstractGraph{T}; norm=true) where T
+    s = zero(T)
```
simonschoelly, line 25: Maybe `zero(T)` is not the right type here. For small graphs it might be something like `Int8` that would overflow quite soon. Maybe it is best, to use `Int` here, or `Float64`, then the resulting type would be the same no matter if `norm` is set to `true` or `false`.

src/community/s_metric.jl
```diff
@@ -0,0 +1,35 @@
...
+# Examples
+```jldoctest
+julia> using LightGraphs
+
+julia> s_metric(star_graph(4))
+0.6
+```
+"""
+
+function s_metric(g::AbstractGraph{T}; norm=true) where T
```
simonschoelly, line 24:
```suggestion
function s_metric(g::AbstractGraph{T}; norm::Bool=true) where T
```

src/community/s_metric.jl
```diff
@@ -0,0 +1,35 @@
...
+    s = zero(T)
+    for e in edges(g)
+        s += outdegree(g, src(e)) * indegree(g, dst(e))
```
simonschoelly, line 27: Do we need some special handling for self-loops here?

test/community/s_metric.jl
```diff
@@ -0,0 +1,22 @@
...
+    @testset "Undirected ($seed)" for seed in [1, 2, 3], (_n, _ne) in [(14, 18), (10, 22), (7, 16)]
+        g = erdos_renyi(_n, _ne; is_directed=false, seed=seed)
+        sm = s_metric(g, norm=false)
+        sm2 = sum([degree(g, src(d)) * degree(g, dst(d)) for d in edges(g)])
+        @test @inferred sm ≈ sm2
+        sm = s_metric(g, norm=true)
+        sm2 /= sum(degree(g).^3)/2
+        @test @inferred sm ≈ sm2
+    end
+end
```
simonschoelly, line 22: Can we have some tests here for graphs different from random graphs? I am especially thinking about graphs with self-loops and isolated vertices.

Also, it would be good to test with graphs of eltype different than `Int`.

test/community/s_metric.jl
```diff
@@ -0,0 +1,22 @@
+using Random, Statistics
+
+@testset "S-metric" begin
+    @testset "Directed ($seed)" for seed in [1, 2, 3], (_n, _ne) in [(14, 18), (10, 22), (7, 16)]
+        g = erdos_renyi(_n, _ne; is_directed=true, seed=seed)
+        sm = s_metric(g, norm=false)
+        sm2 = sum([outdegree(g, src(d)) * indegree(g, dst(d)) for d in edges(g)])
+        @test @inferred sm ≈ sm2
+        sm = s_metric(g, norm=true)
```
simonschoelly, line 9:
```suggestion
        @test  sm ≈ sm2
        sm = @inferred s_metric(g, norm=true)
```
The [inferred macro](https://docs.julialang.org/en/v1/stdlib/Test/#Test.@inferred) is used to check if a function is type stable, i.e. the compiler can infer its resulting argument. What you are testing here is, if `≈` is type stable, which is probably not what you want.

simonschoelly:
@etiennedeg @vboussange This PR already looks quite good, but I think we need to fix a few things before we can merge it.

Krastanov:
This has bitrotted a bit, with a bunch of merge conflicts. I will mark it as a draft to keep my review queue more organized, but do not hesitate to convert it back to not a draft if you have the energy to push it forward.

PR #37, opened 2021-11-11
Update codecov patch to 90%

ViralBShah:
Just to get more greens on PRs. 97% almost always gives us PRs failing because of codecov patch.


ViralBShah:
@jpfairbanks @simonschoelly I am merging this but easy to change if any disagreement.

PR #70, opened 2021-11-15
Move Project status lower in the README

ViralBShah:
@jpfairbanks @simonschoelly Is this ok? I thought it is best to have focus on the package capabilities, and move the transition stuff towards the end.


ViralBShah:
I'm going ahead and merging this. If we don't like it, we can always revert.

PR #71, opened 2021-11-17
adjacency_matrix: Collect neighbors before sorting them

Keno:
The neighbors list may be an iterator that is not sortable. Make
sure to collect it first to get an array back.

### DIFF
--- a/src/linalg/spectral.jl
+++ b/src/linalg/spectral.jl
@@ -59,7 +59,7 @@ function _adjacency_matrix(g::AbstractGraph, T::DataType, neighborfn::Function, 
                 nz -= 1
             end
         end
-        dsts = sort(collect(neighborfn(g, j))) # TODO for most graphs it might not be necessary to sort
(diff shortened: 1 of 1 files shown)


src/linalg/spectral.jl
```diff
@@ -59,7 +59,7 @@ function _adjacency_matrix(g::AbstractGraph, T::DataType, neighborfn::Function,
                 nz -= 1
             end
         end
-        dsts = sort(collect(neighborfn(g, j))) # TODO for most graphs it might not be necessary to sort
+        dsts = sort!(collect(neighborfn(g, j))) # TODO for most graphs it might not be necessary to sort
```
Keno, line 62: In place `sort!` can be algorithmically a fair bit slower than the out of place version.

etiennedeg: From my benchmark, it seemed to be rather equivalent and to allocate less, but you are probably more knowledgeable than me

PR #72, opened 2021-11-24
Faster in the common case of non-huge graphs

PallHaraldsson:


PallHaraldsson:
Line 269, was, and probably still is a bug, assuming Int/Int64.

I know "huge graphs" are a possibility, just not how common. I tried to google it to see if how common, and how large is considered huge:

https://towardsdatascience.com/large-graph-visualization-tools-and-approaches-2b8758a1cd59

> I used to work with large graphs (hundreds of millions of nodes and vertices) [..]
> It is a great savior when you need to draw a really huge graph.

simonschoelly:
Thanks for your contributions.

If I would create Graphs.jl from scratch, I would probably also use `Int32` as default elttype. But as this would be a breaking change, I don't think we should do that. After all it is still very easy to use a different eltype. Also, it is indeed a bit weird, that we get some different behavior on 32 bit system - but then it would also not be possible to have more vertices on such a system.

For some examples of large graphs, see here: <https://snap.stanford.edu/data/index.html>

There are some downsides when using unsigned integers:
- One has to be careful when doing subtracting with these integers, as they wrap around if the go into the negative range.
- Unsigned integers are displayed as hexadecimal numbers in the Julia Repl, e.g.
    ```julia
        julia> UInt16(11)
       0x000b
    ```
  which make them a bit harder to read.
  
Also, there is currently no realistic point to use `UInt64` instead of `Int64`, as it would not be possible to address such a large range of memory.

PallHaraldsson:
>I would probably also use Int32 as default elttype. But as this would be a breaking change, I don't think we should do that.

Thanks for the list. The largest graph in the list has only 3% of the budget allowed for Int32 (for nodes, for number of edges I believe that doesn't factor into this, but then it's at most 84%).

Int32 seems to be enough for all the graphs out there (or at least in the list), so I would consider the change. I was just looking into how implemented, I'm not a user of Graphs.jl, and while the PR untested, wouldn't it speed up by up to 2x?

>After all it is still very easy to use a different eltype.

Right, I'm not sure about your users, but scientists in general, might not think about such CompSci issues, or use defaults. The filp-side is then when this turns out too small, would they know how to debug. Someone should test the PR to see what happens then.

I think most of your users would be grateful, also for now, those with larger graphs (by number of nodes).

The largest graph in the list:
com-Friendster
65,608,366	1,806,067,135

[Facebook is also there, all numbers much smaller... https://snap.stanford.edu/data/gemsec-Facebook.html for 2018 data]

Seemingly next largest graph:
twitter7
17,069,982 users	476,553,560 tweets (edges)

Yes, as subset:
>A collection of 476 million tweets collected between June-Dec 2009

All of it, or a larger subset, would blow up Int32. I suspect Twitter engineers may be competent, and would be careful to use non-default then. Or some import of data could warn or error.

See also there higgs-twitter for seemingly one-day activity: "Higgs boson on 4th July 2012"

Other large, but not really:
sx-stackoverflow
2,601,977	63,497,050 (temporal edges)

ViralBShah:
We can't do it if it is breaking. Can always do it in a major version bump - and otherwise adequately and prominently document how to use a different eltype for better performance.

PR #81, opened 2021-12-23
Update developing.md

mschauer:


mschauer:
Ref. https://discourse.julialang.org/t/is-there-a-lazy-version-of-graph/73119/5

etiennedeg:
Thanks for your PR, I also completed The `AbstractSimpleGraph` API, because the vertices forming a UnitRange is also an assumption of this type.

PR #82, opened 2021-12-23
Faster all_neighbors for SimpleDiGraphs

etiennedeg:
Credits to @abhinavmehndiratta
This was part of the 2.0.0-dev branch on LightGraphs but never made it into master.
I changed it to make it even faster, so a quick review but that should be ok.


simonschoelly (approved):
Thanks, this one makes totally sense. There is a weak assumption that vertices are always sorted incrementally for any kind of graph type, so in the future we could also discuss extending this to other types of directed graphs.

Another idea would be to use a non-allocating iterator here, although then calculating `length` would be a  bit more difficult - either we don't calculate it and might break some user code or we precalculate `length` by by doing a a pass over all vertices when constructing the iterator.

etiennedeg:
Making it an iterator was discussed but @sbromberger thought it would be inconsistent with other neighbor functions, so I kept it like this. I added the reference to the previous commit.

PR #84, opened 2021-12-24
Some normalized_cut() fixes

alyst:
Fixes a few issues I have encountered when running `normalized_cut()`:
  * sometimes `eigs()` returns less eigenvectors than requested, even less than 2. In the latter case the algorithm now would put all vertices to separate modules instead of throwing *OutOfBounds* exception.
  * isolated or nearly isolated vertices result in degenerated *W* matrix and zeros on the *D* diagonal. Now the algorithm would check for that, put the corresponding vertices to separate modules and continue cutting with these vertices excluded (instead of throwing exception upon `inv(D)`). (Such situations should have been handled by the generalized `eigs()`, but it is not implemented in *ArnoldiMethod.jl*, and IIUC using *Arpack.jl* is problematic).
  * some cleanups to `eigs()` (handling real and complex cases uniformly)


etiennedeg:
Is someone able to review this ? It seems good to me, but I'm not very knowledgeable in spectral graph theory

gdalle (approved):
I don't understand the algorithm in detail but the changes are coherent with the PR description and they look good to me

PR #87, opened 2022-01-09
Add more Julia versions to ci/cd workflow

simonschoelly:
This workflow adds the nightly version of Julia and v1.3 to the test workflow. I am currently not exactly sure, which minimal Julia version we should support, but v1.3 is certainly a good start.

I also had to do some changes to ensure that this package still works on Julia v1.3.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,9 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - 'nightly'
           - '1'
+          - '1.3'
         os:
           - ubuntu-latest
           - macos-latest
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -17,7 +17,9 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - 'nightly'
           - '1'
+          - '1.3'
```
hexaeder, line 22:
```suggestion
          - '1.6'
          - '1.3'
```
maybe also test 1.6 since it is the lts?

simonschoelly: I will add it for now - maybe it is not necessary as testing 1.3 and 1.7 should catch everything in between, but then this package is used quite a lot so maybe better to over test. As far as I know, we can have up to 20 concurrent Github actions. If that becomes an issue we could still restrict the number operating system that we test for some versions.

SimonDanisch (approved):

PR #91, opened 2022-01-10
Make minimum Julia version v1.6

simonschoelly:
This PR makes Julia v1.6 the minimum required version for Graphs.jl.

I also would bump the version of this package to v1.6, so that we could still back port fixes for older Julia versions into the Graphs.jl v1.5.x space, and because it is just nice that the version of Julia then aligns with the of Graphs.jl

If any downstream package has some issues with that, please raise your objections here, but in any case, it should still be possible for you to use Graphs.jl at version v1.5.1.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,6 @@ jobs:
           - 'nightly'
           - '1'
           - '1.6'
-          - '1.3'
         os:
           - ubuntu-latest
           - macos-latest
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Graphs"
 uuid = "86223c79-3864-5bf0-83f7-82e725a168b6"
-version = "1.5.1"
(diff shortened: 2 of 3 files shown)


etiennedeg:
1.6 is LTS (see discussion on Slack#graphs)

ViralBShah (approved):

simonschoelly:
Thanks for approving, as there was no resistance I think this is indeed the way go go.

Currently I would like to get as many bug fixes as possible in Graphs.jl, then tag a final `1.5.2` version that still works with older Julia versions, and then merge in this PR.

ViralBShah:
I'll reiterate some of the reasons from slack (from memory):

1. 1.6 is the new LTS.
2. Lots of threading related fixes from 1.3 to 1.6 and we want to use that as the baseline for new features.
3. It is best not to spend limited developer bandwidth on backporting to very old versions of Julia.

Issue #88, opened 2022-01-09
Squash does not work correctly for AbstractGraph


simonschoelly:
The way the `squash` is currently implemented is by either calling `SimpleGraph{T}(g)` or `SimpleDiGraph{T}(g)` for some integer type `T`. This fails if `g` is a subtype of `AbstractGraph` for which these functions are not implemented. There are multiple ways to solve this:

1. Restrict `squash` to `SimpleGraph` and `SimpleDiGraph` and leave the implementation of that function for other graph types open.
2. Ensure that the constructors `SimpleGraph{T}(g)` or `SimpleDiGraph{T}(g)` work for for any kind of `g::AbstractGraph`.
3. Require that subtypes of any kind of graph  type `SomeGraph <: AbstractGraph` implement a constructor `SomeGraph{T}(g)` and adjust `squash` accordingly.

In my opinion, 1. is the way to go. 2. could lead to some information (such as metadata) being thrown away by `squash` which is probably not what the user wants. 3. would make the interface more complicated and would make existing packages not correspond to the interface anymore.

etiennedeg:
I also think 1. is the way to go. It should nonetheless be possible to implement a conversion from `AbstractGraph` to `SimpleGraph` by collecting neighbors.

simonschoelly:
Yes, not with the squash function though. One way to do that is to implement functions
```julia
SimpleGraph(::AbstractGraph)
SimpleGraph{T}(::AbstractGraph)
```
 or with a different function name, to keep the number of constructs a bit low. We would also need to figure out what to do in cases when the argument is a directed graph. Currently there is already a constructor `SimpleGraph(::SimpleDiGraph)` which adds an undirected edge, if there is one edge in either direction in the directed graph.

simonschoelly:
Solved by #93

PR #93, opened 2022-01-15
Make squash specialized on simple graphs

simonschoelly:
Related issue: #88 

I did not remove `squash(::AbstractGraph)` as this has the potential of breaking something, even though it is not a good idea to use that  method.

I also changed `squash` for  simple graphs slightly so that it returns the original graph instead of a copy if squashing would not make the graph smaller.

### DIFF
--- a/src/SimpleGraphs/specializations.jl
+++ b/src/SimpleGraphs/specializations.jl
@@ -11,18 +11,18 @@ function Graphs.squash(g::Union{SimpleGraph, SimpleDiGraph}; alwayscopy::Bool=tr
     G = is_directed(g) ? SimpleDiGraph : SimpleGraph
     T = eltype(g)
 
-    (!alwayscopy && T <: Union{Int8, UInt8}) && return g
-    nv(g) < typemax(Int8)     && return G{Int8}(g)
-    nv(g) < typemax(UInt8)    && return G{UInt8}(g)
+    (!alwayscopy && T <: Union{Int8, UInt8})   && return g
+    nv(g) < typemax(Int8)                      && return G{Int8}(g)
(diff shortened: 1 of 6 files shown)


etiennedeg:
> I also changed `squash` for simple graphs slightly so that it returns the original graph instead of a copy if squashing would not make the graph smaller.

Wouldn't it be inconsistent and a breaking change ?

simonschoelly:
> Wouldn't it be inconsistent and a breaking change ?
In most use cases of `squash` one probably will not reuse the original graph. but otherwise it is breaking, yes. I will therefore add a `alwayscopy` keyword argument.

simonschoelly:
I have now added a `alwayscopy` option to `squash` for simple graphs, so by default one will always get a copy of the input graph.

src/SimpleGraphs/specializations.jl
```diff
@@ -0,0 +1,29 @@
+
+"""
+    squash(g::Union{SimpleGraph. SimpleDiGraph}; alwayscopy=true)
```
matbesancon, line 3:
```suggestion
    squash(g::Union{SimpleGraph, SimpleDiGraph}; alwayscopy=true)
```

src/SimpleGraphs/specializations.jl
```diff
@@ -0,0 +1,29 @@
...
+
+Specialised version of `Graphs::Squash` for `SimpleGraph` and `SimpleDiGraph`.
```
matbesancon, line 5:
```suggestion
Specialised version of `Graphs.squash` for `SimpleGraph` and `SimpleDiGraph`.
```

matbesancon (approved):

Issue #10, opened 2021-10-18
Mark parallel algorithms as experimental


ViralBShah:
A parallel BFS  non-deterministic issue in #9
https://github.com/JuliaGraphs/Graphs.jl/pull/9/checks?check_run_id=3932016414#step:6:227

Trace says this is where the test fails: https://github.com/JuliaGraphs/Graphs.jl/blob/master/src/Parallel/traversals/gdistances.jl#L67

The test failed in the parallel gdistances test.

ViralBShah:
I'm unable to recreate this test failure outside of CI.

Keno:
This entire function is making assumptions about Julia's threaded memory model that do not hold true. I'm not sure as to the exact cause of the CI issue, but this function is entirely unsound as is.

ViralBShah:
I was wondering if the parallel graph algorithms should be split out of Graphs.jl into a separate more experimental package.

ViralBShah:
Discussion around threaded gdistance: https://discourse.julialang.org/t/optimistic-concurrency-control/12402

As @Keno pointed out on slack, this could be correctly implemented with 1.7 atomics. But for now, it would be best to split these into a separate package to make Graphs.jl more reliable and a nice side effect would be that it will have fewer dependencies.

jpfairbanks:
Yes I think this is appropriate unless anyone wants to take on the mantle of getting this updated to correctly use the 1.7 atomics in a hurry. We might find out if anyone is using the parallel stuff when we remove it.

tkf:
FYI there's https://github.com/JuliaConcurrent/UnsafeAtomics.jl (registered) and a wrapper https://github.com/JuliaConcurrent/AtomicArrays.jl (not registered). You'd need something like this for concurrently accessed arrays like `visited`.

Also, Julia 1.7 does not help much here since we still don't have atomically accessible arrays https://github.com/JuliaLang/julia/issues/32455. However, if we only need to access small isbits types, we don't need direct support in the Julia core. I think I can provide pre-1.7-compatible API from AtomicArrays.jl. if people need it. (Or you can just directly write llvm IR)

simonschoelly:
The parallel functions are already in a `Parallel` submodule, that could be considered experimental. I haven't looked at the issues here yet, but I would rather keep these function in this package for now - maybe disable the tests and add some warnings to the documentation.

ViralBShah:
[...]

PR #94, opened 2022-01-15
Move broken Parallel.gdistances function to Experimental submodule

simonschoelly:
The `Parallel.gdistances`  functions seems to have some incorrect assumptions about threaded memory. This leads to tests failing from time to time. In order to not completely throw away the code, this PR moves it to the `Experimental.Parallel` submodule and deactivates the tests for this function.

See #10


etiennedeg (changes requested):
There is also `gdistances` tests in 
Graphs.jl/test/traversals/bfs.jl (lines 29-35)

`gdistances` should not be globally exported, but only exported in Experimental module.

We should also reflect this change in the documentation (`gdistances` appears in 
Graphs.jl/docs/src/pathing.md for example)

`period` is calling `gdistances`, what should we do about it ?

@ViralBShah  mentionned that he also saw `bfs` failing, but I couldn't find such failing test.

ViralBShah:
The `bfs` failures only happen sporadically in CI for me. I could not reproduce on my own machine.

simonschoelly:
The tests in `Graphs.jl/test/traversals/bfs.jl` are for the non-parallel version of `gdistances`  which is is not broken, so it still should be exported:
https://github.com/JuliaGraphs/Graphs.jl/blob/abdfceb0ec54990e7ce30040e1902aa3e1a1e2f3/src/traversals/bfs.jl#L126-L141

ViralBShah:
IIRC, it was parallel bfs that was broken - not the sequential one.

simonschoelly:
The testset in which the failing `gdistances` test where defined, is called
https://github.com/JuliaGraphs/Graphs.jl/blob/abdfceb0ec54990e7ce30040e1902aa3e1a1e2f3/test/parallel/traversals/gdistances.jl#L1
Is it possible, that this is the reason why you think that `bfs` fails from time to time?

ViralBShah:
Yes that would explain it. I was just going by what I was seeing in the logs.

etiennedeg:
Ok, my bad, all my remarks were on sequential `gdistances`, so I think its ok for me. (gdistances was not even mentionned in the documentation of parallel algorithms)

simonschoelly:
Thanks for reviewing :) Will merge this now, so hopefully we won't see that many failing tests anymore.

PR #95, opened 2022-01-15
Allow passing random number generator to label-propagation

### LINKED ISSUE #83: [BUG] Intermittently failing tests for `label_propagation` on mac OS
Examples of failing tests [here](https://github.com/JuliaGraphs/Graphs.jl/runs/4619930056?check_suite_focus=true) and [here](https://github.com/JuliaGraphs/Graphs.jl/runs/4336947072?check_suite_focus=true)

simonschoelly:
This adds a `rng` keyword argument to `label-propagation`, so that one pass a random generator. Should hopefully fix #83.

### DIFF
--- a/src/community/label_propagation.jl
+++ b/src/community/label_propagation.jl
@@ -9,7 +9,10 @@ the second is the convergence history for each node. Will return after
 ### References
 - [Raghavan et al.](http://arxiv.org/abs/0709.2938)
 """
-function label_propagation(g::AbstractGraph{T}, maxiter=1000; rng::AbstractRNG=GLOBAL_RNG) where T
+function label_propagation(g::AbstractGraph{T}, maxiter=1000; rng::Union{Nothing, AbstractRNG} = nothing, seed::Union{Nothing, Integer} = nothing) where T
+
+    rng = rng_from_rng_or_seed(rng, seed)
+
     n = nv(g)
     n == 0 && return (T[], Int[])
 
--- a/src/community/label_propagation.jl
+++ b/src/community/label_propagation.jl
@@ -1,5 +1,5 @@
 """
-    label_propagation(g, maxiter=1000)
+    label_propagation(g, maxiter=1000; rng=GLOBAL_RNG)
 
 Community detection using the label propagation algorithm.
 Return two vectors: the first is the label number assigned to each node, and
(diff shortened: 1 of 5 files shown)


src/community/label_propagation.jl
```diff
@@ -9,7 +9,7 @@ the second is the convergence history for each node. Will return after
 ### References
 - [Raghavan et al.](http://arxiv.org/abs/0709.2938)
 """
-function label_propagation(g::AbstractGraph{T}, maxiter=1000) where T
+function label_propagation(g::AbstractGraph{T}, maxiter=1000; rng::AbstractRNG=GLOBAL_RNG) where T
```
etiennedeg, line 12: After a quick look at the codebase, the rng seems to always be passed as an Integer seed, and then is converted to rng using `getRNG` in utils.jl. It seems it would be inconsistent to do something else. Maybe a breaking change for 2.0 ?

simonschoelly: Breaking no, inconsistent maybe. But I think there already was a conversation a while ago, about changing everything to `rng` for random functions. What we could do is to allow either of `seed` or `rng` as keyword argument and just emit a deprecation warning when someone uses `seed`.

Definitely would not wait for a 2.0 version with such a change.

etiennedeg: Ok, the deprecation road seems to be the solution.

etiennedeg: Found [this](https://github.com/sbromberger/LightGraphs.jl/pull/1324) and [this](https://github.com/sbromberger/LightGraphs.jl/pull/1328/files). Another stuff that was in the 2.0 dev and never made it to master. I will try to bring the remaining commits of this branch when I have the time, there was also interesting stuff about `astar`.

simonschoelly: The changes in these PRs are breaking though (which would not be such a big issue for 2.0), I will write something non-breaking.

simonschoelly: I have added a `rng_from_rng_or_seed`  helper function. While it is strictly not necessary in the case here, it serves as an example, and we can use it for other randomized functions in some further PR.

test/runtests.jl
```diff
@@ -4,6 +4,7 @@ using Graphs.Experimental
 using Test
 using SparseArrays
 using LinearAlgebra
+using Compat
```
matbesancon, line 7: why is Compat needed here? It would be great to be able to do without it

simonschoelly: It is there so I can use `ismutable` in `test/utils.jl`. Before Julia v1.7 that function did not exist, but on the other hand there was one with the reverse effect called `isimutable`

I had added `Compat` as a depedency a while ago for the same reason, so that some functions in in `src/shortestpaths/johnson.jl` would work with older Julia versions. So this is not a new dependency.

matbesancon (approved):

PR #97, opened 2022-01-23
Make random numbers deterministic for graph matrices tests

simonschoelly:
This PR makes the random number generator for graph matrices deterministic. This is not ideal, but at the moment these tests fail a bit too often. Maybe we can come up with some better tests in the future.


matbesancon (approved):

matbesancon:
I agree this is an acceptable workaround for now

Issue #86, opened 2022-01-09
1.5 silently dropped julia 1.3 support


SimonDanisch:
If I interpret the logs correctly, it looks like 1.3 isn't supported anymore:
https://github.com/JuliaPlots/Makie.jl/runs/4753606377?check_suite_focus=true#step:6:284
Was this on purpose?

simonschoelly:
Definitely not, although it might make sense to do that at some point.. If you look at the [Project.toml](https://github.com/JuliaGraphs/Graphs.jl/blob/v1.5.0/Project.toml),we also do not specify any minor version for Julia. Apart from the version number, nothing in that file seems to have changed since v1.4.1, so I am not sure what is going on here.

simonschoelly:
Apparently we do not check for specific Julia versions in our CI/CD scripts: https://github.com/JuliaGraphs/Graphs.jl/blob/master/.github/workflows/ci.yml. Not sure why, but could explain why we might have missed some bug on older Julia versions.

simonschoelly:
Looks like a PR that was merged into the newest release uses some syntax that does not work on Julia v1.3: https://github.com/JuliaGraphs/Graphs.jl/blob/c2e41703cb403246379d82591594f6cd18d43946/src/operators.jl#L847. I will try to fix that asap.

simonschoelly:
Fixed by #87

SimonDanisch:
Thanks for the quick fix!

PR #100, opened 2022-01-31
fixes #13

### LINKED ISSUE #13: [BUG] `biconnected_components` loses some edges - should we return only vertices ?
This is a repost of [this issue](https://github.com/sbromberger/LightGraphs.jl/issues/1591), I think it deserves discussion on whether we should return edges instead of vertices.

**Summary :**
``` julia
start=[1,1,2,2,2,3,4,4,5,5,7,7,8,8,9,9,10,10,12,12,1,2,3]
destination=[2,4,4,5,3,5,7,8,7,8,9,10,9,10,11,12,12,13,13,11,11,12,13]

graph_s=SimpleWeightedGraph(start,destination,collect(1:23))

bi_components= biconnected_components(graph_s)
````
`graph_s` has 23 edges, but only 19 are returned in the unique biconnected component.

> etienneINSA : The error is [here](https://github.com/sbromberger/LightGraphs.jl/blob/2a644c2b15b444e7f32f73021ec276aa9fc8ba30/src/biconnectivity/biconnect.jl#L53) : the edge is added only if `state.low[v] > state.depth[w]`, whereas it should be always added to the stack. I can make a PR, but I have an interrogation: Why do we return the biconnected components by their edges? As the subgraphs are induced, this is more easily defined by the vertices. Is there a good reason to do that, or would it be too breaking to change this behavior?

etiennedeg:


simonschoelly:
> Bump. Is it good to merge? I can't review it, but I think if all tests pass and there is adequate test coverage, it would be nice to allow people to merge if no comments in a few days. I suppose this is more of a policy question.

Don't you think that is a bit dangerous? We just had an issue because of that recently: https://github.com/JuliaGraphs/Graphs.jl/issues/86. You re probably involved in more packages though, how do they handle such things?

ViralBShah:
I don't think there is a clear answer. It can certainly backfire - and it is a balance. I personally prefer moving things forward if major contributors have adequate tested a new feature, and if review bandwidth is not available. I fully understand it is 2 steps forward and 1 step backward in many cases - and this is only my personal view.

simonschoelly (approved):

simonschoelly:
Its a bug fix anyway, so it won't make things worse than incorrect.

simonschoelly:
Thanks!

PR #113, opened 2022-03-03
error for self loops with coloring. fixes #112

### LINKED ISSUE #112: BoundsError in perm_greedy_color
BoundsError in `perm_greedy_color`, I think the problem is the initialization of the `cols` vector using `undef` (see the index value in the screen capture). The error happens most of the time, but not always. Here there is an example:

image

You can reproduce it by running:

```julia
using GraphRecipes, Plots
using Graphs
using LinearAlgebra
n = 5
adjmat = rand(0:1, n, n)
g = SimpleGraph(Symmetric(adjmat))
c = Graphs.degree_greedy_color(g)
```

**Version information**
 ```
      From worker 11:   Julia Version 1.6.2
      From worker 11:	Commit 1b93d53fc4 (2021-07-14 15:36 UTC)
      From worker 11:	Platform Info:
      From worker 11:	  OS: Linux (x86_64-pc-linux-gnu)
      From worker 11:	  CPU: Intel(R) Core(TM) i7-10850H CPU @ 2.70GHz
      From worker 11:	  WORD_SIZE: 64
      From worker 11:	  LIBM: libopenlibm
      From worker 11:	  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)
      From worker 11:	Environment:
      From worker 11:	  JULIA_REVISE_WORKER_ONLY = 1
```

```
      From worker 11:         Status `/tmp/jl_buYAIX/Project.toml`
      From worker 11:	  [bd48cda9] GraphRecipes v0.5.9
      From worker 11:	  [86223c79] Graphs v1.6.0
      From worker 11:	  [91a5bcdd] Plots v1.25.11
      From worker 11:	  [37e2e46d] LinearAlgebra
      From worker 11:	  [44cfe95a] Pkg
```

etiennedeg:


gdalle:
Hi! Do we have to use `has_self_loops`? This probably traverses the entire graph unnecessarily. I was thinking we could just check, for each `v`, if we end up giving it a color, and error from within the loop otherwise

etiennedeg:
It traverses only the vertices, so it's ok I think. I'm not sure testing the self loops during the main loop will be more performant than doing it at the beginning, each self loop will be checked no matter what. And if a graph does not support self loops, it just gives constant overhead.

gdalle (approved):

gdalle:
OK then, I think we can disregard the failure on nightly and merge this?

PR #118, opened 2022-03-26
Update Shortest paths docs

nhz2:
I added some details to the Shortest paths section of the docs.

Maybe some of the specific `dijkstra_shortest_paths` info should be moved to `function dijkstra_shortest_paths` docs?

### DIFF
--- a/docs/src/first_steps/paths_traversal.md
+++ b/docs/src/first_steps/paths_traversal.md
@@ -31,7 +31,7 @@ The package also includes uniform random walks and self avoiding walks with the 
 The following properties always hold for shortest path algorithms implemented here:
 
 - The distance from a vertex to itself is always `0`.
-- The distance between two vertices with no connecting edge is always `Inf`.
+- The distance between two vertices with no connecting edge is always `Inf` or `typemax(Int64)`.
 
 The `dijkstra_shortest_paths`, `desopo_pape_shortest_paths`, `floyd_warshall_shortest_paths`, `bellman_ford_shortest_paths`, and `yen_shortest_paths` functions return path states (subtypes of `Graphs.AbstractPathState`) that contain various information about the graph learned during traversal.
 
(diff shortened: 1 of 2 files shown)


nhz2:
@etiennedeg  @gdalle , could you please review this?

I would like to write an edge betweenness function using the `dijkstra_shortest_paths` in  `Graphs.jl`, but its outputs are currently poorly documented.

docs/src/first_steps/paths_traversal.md
```diff
@@ -31,7 +31,7 @@ The package also includes uniform random walks and self avoiding walks with the
 The following properties always hold for shortest path algorithms implemented here:
 
 - The distance from a vertex to itself is always `0`.
-- The distance between two vertices with no connecting edge is always `Inf`.
+- The distance between two vertices with no connecting edge is always `Inf` or `typemax(Int64)`.
```
etiennedeg, line 34:
```suggestion
- The distance between two vertices with no connecting edge is always `Inf` or `typemax(eltype(distmx))`.
```

etiennedeg (commented):
Thanks for your contribution, I think we should also document this in [Algorithms API](https://juliagraphs.org/Graphs.jl/dev/algorithms/shortestpaths/), this is actually less documented than the [Paths and traversal](https://juliagraphs.org/Graphs.jl/dev/first_steps/paths_traversal/) section.
As a side note, we should also add accessors for the `AbstractPath` object, we should not need to rely on internal fields.

nhz2:
> Thanks for your contribution, I think we should also document this in [Algorithms API](https://juliagraphs.org/Graphs.jl/dev/algorithms/shortestpaths/), this is actually less documented than the [Paths and traversal](https://juliagraphs.org/Graphs.jl/dev/first_steps/paths_traversal/) section. As a side note, we should also add accessors for the `AbstractPath` object, we should not need to rely on internal fields.

Thanks for the review, I moved the `dijkstra_shortest_paths` specific info to the `dijkstra_shortest_paths` docstring,
and added links to the referenced functions in [Paths and traversal](https://juliagraphs.org/Graphs.jl/dev/first_steps/paths_traversal/).

etiennedeg (approved):

gdalle:
Thanks @etiennedeg

PR #121, opened 2022-04-28
make clipboard(g) work

anandijain:
With this PR, if I have a graph in one REPL and I `clipboard(g)`, I can paste it in another and have the same graph in both places

### DIFF
--- a/src/SimpleGraphs/SimpleGraphs.jl
+++ b/src/SimpleGraphs/SimpleGraphs.jl
@@ -50,7 +50,7 @@ An abstract type representing a simple graph structure.
 """
 abstract type AbstractSimpleGraph{T<:Integer} <: AbstractGraph{T} end
 
-function show(io::IO, g::AbstractSimpleGraph{T}) where T
+function show(io::IO, ::MIME"text/plain", g::AbstractSimpleGraph{T}) where T
     dir = is_directed(g) ? "directed" : "undirected"
     print(io, "{$(nv(g)), $(ne(g))} $dir simple $T graph")
 end
--- a/test/simplegraphs/simplegraphs.jl
+++ b/test/simplegraphs/simplegraphs.jl
@@ -30,16 +30,18 @@ import Random
     gx = SimpleGraph()
     for g in testgraphs(gx)
         T = eltype(g)
-        @test sprint(show, g) == "{0, 0} undirected simple $T graph"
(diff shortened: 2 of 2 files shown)


etiennedeg:
I agree that the current form of printing is annoying because not canonical.
I think the `show` function with `plain/text` should also be canonical, we should implement it to look something like this:
``` julia
julia> path_graph(5)
{4, 3} undirected simple Int64 graph
Edge 1 => 2, Edge 2 => 3, Edge 3 => 4, Edge 4 => 5
```
or 
```julia
julia> path_graph(5)
{4, 3} undirected simple Int64 graph
Edge 1 => 2
Edge 2 => 3
Edge 3 => 4 
Edge 4 => 5
```
or 
```julia
julia> path_graph(5)
{4, 3} undirected simple Int64 graph
1 => [2]
2 => [1, 3]
3 => [2, 4]
4 =>[3, 5]
5 => [4]
```
We should implement summary to output the old behavior (currently, it output the default `"SimpleGraph{Int64}"`, which is less satisfactory.

simonschoelly:
If I understand this pr correctly, it ensures that `show(io::IO, g::AbstractSimpleGraph)` just prints a form that is also a Julia statement that can be evaluated and creates a graph again? I guess that makes sense to have it that way then.

And the behavior in the repl is still the same, right? I think that one calls `display(g)` which in turn calls `show(io, io::IO, ::MIME"text/plain"(), g)`

> I agree that the current form of printing is annoying because not canonical. I think the `show` function with `plain/text` should also be canonical, we should implement it to look something like this:

What exactly do you mean by canonical? That it reports all the edges? That won't work for all but very small graphs,  as it would just fill the repl.

simonschoelly (approved):

etiennedeg:
Yes, from my understanding, I think it means that two different objects should have distinct canonical expressions, so the print should display all the edges. Almost all Julia objects behave the same, for example, call `show(stdout, "text/plain", collect(1:100000))`, it will print the entirety of the array. This is not that much a problem because the output is automatically cropped when calling `display`. 

Some objects don't have a strictly canonical display, for example, the "text/plain" display of a sparse matrix is the dotted representation, which is not canonical for big matrices, so I don't really know how much we need to enforce this rule.

simonschoelly:
@etiennedeg Is it ok, if we take this discussion outside of this PR and merge this one?

Issue #120, opened 2022-04-21
Error in A* implementation


gdalle:
In the implementation of the [A* algorithm](https://github.com/JuliaGraphs/Graphs.jl/blob/master/src/shortestpaths/astar.jl), there is a vector called [`closed_set`](https://github.com/JuliaGraphs/Graphs.jl/blob/da30b3b255c4fb7c4f70af136cff482cac7ff411/src/shortestpaths/astar.jl#L22) which is used to indicate whether a vertex has already been expanded.

Judging by the similar notations, this implementation is taken almost as-is from the [Wikipedia page](https://en.wikipedia.org/wiki/A*_search_algorithm), but on Wikipedia the `closed_set` does not exist. In fact, the article clearly states that

>  When the heuristic being used is admissible but not consistent, it is possible for a node to be expanded by A* many times, an exponential number of times in the worst case.

Therefore, I think the presence of `closed_set` in the code is a mistake and should be corrected.

gdalle:
@matbesancon @etiennedeg

matbesancon:
if a vertex is expanded several times, are the following ones redundant with the first time this vertex is encountered? From your message I assume not

gdalle:
Probably not, because this means we have found a cheaper path to said vertex, which may reveal a cheaper path to its neighbors

matbesancon:
I'm ok with removing this. Even better if there is an example that fails to add to the tests

captchanjack:
Hello! Just looking into this as well

Looks like the Wikipedia page you referenced **used** to have the closed_set documented in its algorithm, found a reference to an [older version](https://stackoverflow.com/questions/13578287/a-a-star-algorithm-clarification#:~:text=The%20%22Closed%20set%22%20is%20the,named%20HasBeenVisited%20or%20something%20similar.) of the page

Pretty much all the implementations I've found online use a closed_set, e.g. [python](https://github.com/jrialland/python-astar/blob/master/astar/__init__.py), [c++](https://github.com/justinhj/astar-algorithm-cpp/blob/master/cpp/stlastar.h), even this [youtube video](https://www.youtube.com/watch?v=eSOJ3ARN5FM&t=890s)

I don't really know what is correct, sources seem inconsistent...

gdalle:
[...]

Issue #59, opened 2021-11-11
[Port] [BUG] `a_star` fails on `SimpleWeightedDiGraph`


gdalle:
See https://github.com/sbromberger/LightGraphs.jl/issues/1579

gdalle:
Fixed by #125

PR #125, opened 2022-05-13
Fix A* implementation

gdalle:
This PR resolves two issues with the A* algorithm:

- #59 To make A* work with custom edge types `E` that do not define a constructor `E(s,d)`, I created an undocumented method `make_edge(g,s,d)` that falls back on `SimpleEdge`. This follows a lengthy discussion in #116 where several approaches were debated, some of them breaking. 
- #120 To make A* correct, I removed the `closed_set` variable which prevented it to be exponential in the worst-case scenario (see [this Wikipedia paragraph](https://en.wikipedia.org/wiki/A*_search_algorithm#Optimality_and_Consistency) from the article which inspired our implementation). However, I see no obvious way to test this pathological behavior so I'm submitting the PR as is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,4 @@ docs/site/
 benchmark/.results/*
 benchmark/.tune.jld
 *.cov
-Manifest.toml
+/Manifest.toml
--- a/src/shortestpaths/astar.jl
+++ b/src/shortestpaths/astar.jl
@@ -69,7 +69,7 @@ function a_star(g::AbstractGraph{U},  # the g
     # if we do checkbounds here, we can use @inbounds in a_star_impl!
     checkbounds(distmx, Base.OneTo(nv(g)), Base.OneTo(nv(g)))
 
-    open_set = PriorityQueue{Integer, T}()
+    open_set = PriorityQueue{U, T}()
     enqueue!(open_set, s, 0)
 
     g_score = fill(Inf, nv(g))
--- a/src/shortestpaths/astar.jl
+++ b/src/shortestpaths/astar.jl
@@ -18,6 +18,7 @@ end
 function a_star_impl!(g, # the graph
     goal, # the end vertex
     open_set, # an initialized heap containing the active vertices
+    closed_set, # an (initialized) color-map to indicate status of vertices
     g_score, # a vector holding g scores for each node
     came_from, # a vector holding the parent of each node in the A* exploration
     distmx,
@@ -33,7 +34,11 @@ function a_star_impl!(g, # the graph
             return total_path
         end
 
+        closed_set[current] = true
+
         for neighbor in Graphs.outneighbors(g, current)
+            closed_set[neighbor] && continue
+
             tentative_g_score = g_score[current] + distmx[current, neighbor]
 
             if tentative_g_score < g_score[neighbor]
@@ -57,20 +62,22 @@ Compute a shortest path using the [A* search algorithm](http://en.wikipedia.org/
 - `s::Integer`: the source vertex
 - `t::Integer`: the target vertex
 - `distmx::AbstractMatrix`: an optional (possibly sparse) `n × n` matrix of edge weights. It is set to `weights(g)` by default (which itself falls back on [`Graphs.DefaultDistance`](@ref)).
-- `heuristic`: an optional callable mapping each vertex to a lower estimate of the remaining distance from `v` to `t`. It is set to `v -> 0` by default (which corresponds to Dijkstra's algorithm)
+- `heuristic::Function`: an optional function mapping each vertex to a lower estimate of the remaining distance from `v` to `t`. It is set to `v -> 0` by default (which corresponds to Dijkstra's algorithm)
 - `edgetype_to_return::Type{E}`: the eltype `E<:AbstractEdge` of the vector of edges returned. It is set to `edgetype(g)` by default. Note that the two-argument constructor `E(u, v)` must be defined, even for weighted edges: if it isn't, consider using `E = Graphs.SimpleEdge`.
 """
 function a_star(g::AbstractGraph{U},  # the g
     s::Integer,                       # the start vertex
     t::Integer,                       # the end vertex
     distmx::AbstractMatrix{T}=weights(g),
-    heuristic=n -> zero(T),
+    heuristic::Function=n -> zero(T),
     edgetype_to_return::Type{E}=edgetype(g)) where {T, U, E<:AbstractEdge}
     # if we do checkbounds here, we can use @inbounds in a_star_impl!
     checkbounds(distmx, Base.OneTo(nv(g)), Base.OneTo(nv(g)))
 
     open_set = PriorityQueue{U, T}()
     enqueue!(open_set, s, 0)
+
+    closed_set = zeros(Bool, nv(g))
 
     g_score = fill(Inf, nv(g))
     g_score[s] = 0
@@ -79,6 +86,14 @@ function a_star(g::AbstractGraph{U},  # the g
     came_from[s] = s
 
     a_star_impl!(
-        g, t, open_set, g_score, came_from, distmx, heuristic, edgetype_to_return
+        g,
+        t,
+        open_set,
+        closed_set,
+        g_score,
+        came_from,
+        distmx,
+        heuristic,
+        edgetype_to_return
     )
 end
--- a/src/shortestpaths/astar.jl
+++ b/src/shortestpaths/astar.jl
@@ -6,12 +6,11 @@
 function reconstruct_path!(total_path, # a vector to be filled with the shortest path
     came_from, # a vector holding the parent of each node in the A* exploration
     end_idx, # the end vertex
-    g) # the graph
-
-    E = edgetype(g)
+    g,  # the graph
+    edgetype_to_return::Type{E}=edgetype(g)) where {E<:AbstractEdge}
     curr_idx = end_idx
     while came_from[curr_idx] != curr_idx
-        pushfirst!(total_path, E(came_from[curr_idx], curr_idx))
+        pushfirst!(total_path, edgetype_to_return(came_from[curr_idx], curr_idx))
         curr_idx = came_from[curr_idx]
     end
 end
@@ -19,29 +18,22 @@ end
 function a_star_impl!(g, # the graph
     goal, # the end vertex
     open_set, # an initialized heap containing the active vertices
-    closed_set, # an (initialized) color-map to indicate status of vertices
     g_score, # a vector holding g scores for each node
-    f_score, # a vector holding f scores for each node
     came_from, # a vector holding the parent of each node in the A* exploration
     distmx,
-    heuristic)
-
-    E = edgetype(g)
-    total_path = Vector{E}()
+    heuristic,
+    edgetype_to_return::Type{E}) where {E<:AbstractEdge}
+    total_path = Vector{edgetype_to_return}()
 
     @inbounds while !isempty(open_set)
         current = dequeue!(open_set)
 
         if current == goal
-            reconstruct_path!(total_path, came_from, current, g)
+            reconstruct_path!(total_path, came_from, current, g, edgetype_to_return)
             return total_path
         end
 
-        closed_set[current] = true
-
         for neighbor in Graphs.outneighbors(g, current)
-            closed_set[neighbor] && continue
-
             tentative_g_score = g_score[current] + distmx[current, neighbor]
 
             if tentative_g_score < g_score[neighbor]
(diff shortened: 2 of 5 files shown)


etiennedeg:
Why name it `make_edge(g,s,d)` when it gathers an existing edge?

gdalle:
@etiennedeg it doesn't necessarily gather an existing edge, it recreates an edge from `s` to `d`. If you look at the A* code I use it on `(s, d) = (0, 0)` to get the right type for the vector of edges returned by the method. This edge obviously doesn't exist, but that's the workaround I thought I needed

simonschoelly:
I was not involved in that discussion but I still wonder, if we can't solve that issue without adding another method to the interface?

gdalle:
@simonschoelly it was a lengthy discussion (see [this previous PR](https://github.com/JuliaGraphs/Graphs.jl/pull/116), but the conclusion I drew is that there is no clean way out of this mess. @matbesancon suggested what I think is the only non-breaking fix, but the real solution would be to change the behavior of A* and return a vector of vertices instead of edges. 

Please feel free to weigh in though! In my view it is a really important issue: if A* doesn't work on the standard library for weighted graphs, this kinda looks bad ^^

jpfairbanks:
I think that switching to returning a sequence of vertices precludes support for multigraphs in the future. I would rather return a sequence of SimpleEdge for all graph types or have it give a copy of the actual edges in the input graph.  I think the latter option would avoid needing a way to generically construct an edge of arbitrary type. 


We could add a flag to the function signature to enable this behavior. If avoiding a SemVer bump is important.

gdalle:
Fair point, I hadn't though of multigraphs.
Returning the actual edge object was the idea behind the proposal by @etiennedeg : he suggested we add a `get_edge(g, s, d)` method to the interface (https://github.com/JuliaGraphs/Graphs.jl/pull/116#issuecomment-1102782513). The only issue is that we would need to enforce the existence of this method in all `AbstractGraph` implementations, which (to me) seems just as breaking as changing the return type of `a_star`.
Of course, if we don't care about semver bumps, then we can go ahead and tag 2.0 for this, but to me it seems like a waste of a good major version, because there surely are many more things that we would like to break.

gdalle:
Another option would be to introduce a new method with a different name, say `astar` instead of `a_star`, and to clearly explain the difference between them in the docs. That way we ensure there is no loss of functionality, while providing a new working implementation to our liking

simonschoelly:
I think a `get_edge` would be something we need in the future (actually `get_edges` so that we can support multigraphs) - but is probably a 2.0 thing.

The way `make_edge` is implemented here is also sort of breaking - if another graph type had an edgetype `E` that supported `E(s, d)` then this change will make a-star stop working.

My proposal is to somehow fix this downstream, i.e. in `SimpeWeightedGraphs` and other packages. As we probably do not to copy the whole a-star code into that package, we use some helper function similar to yours, lets call it `_make_edge`, that works somehow like that
```julia
   function _make_edge(g, s, d) 
      return edgetype(g)(s, d)
      # TODO maybe catch exception and output a friendly message to show what to implement
    end
```
then instead of exporting that function in the interface, we keep it somewhere near the astar function. We can then import and override that function in `SimpeWeightedGraphs.jl`.

This is of course also not ideal, but I think we should avoid making the interface more complicated than it already is, as otherwise we will scare people away from creating their own graph types.

Apart from that, I also like James idea of adding a keyword argument to the function to control behavior.

simonschoelly:
> Another option would be to introduce a new method with a different name, say `astar` instead of `a_star`, and to clearly explain the difference between them in the docs. That way we ensure there is no loss of functionality, while providing a new working implementation to our liking

This could also be in a submodule where we can stage v2.0 functions without respecting semver (although I am not sure about name collisions if we use the same name).

gdalle:
> I think a `get_edge` would be something we need in the future (actually `get_edges` so that we can support multigraphs) - but is probably a 2.0 thing.

Totally agree

> The way `make_edge` is implemented here is also sort of breaking - if another graph type had an edgetype `E` that supported `E(s, d)` then this change will make a-star stop working.

Yeah, that is what I realized this afternoon. Sorry about that

> My proposal is to somehow fix this downstream, i.e. in `SimpleWeightedGraphs` and other packages. As we probably do not to copy the whole a-star code into that package, we use some helper function similar to yours, lets call it `_make_edge`

Just to be clear, cause I'm new to this whole versioning stuff: if we require downstream users to add something to their packages for `a_star` to keep working, that's breaking, right?

> Apart from that, I also like James idea of adding a keyword argument to the function to control behavior.

To me this sounds like a good compromise waiting for 2.0, and it keeps the interface just as simple. To enable static dispatch I think the additional argument could be the type of the edges we want in the returned path:

```julia
function a_star(g::AbstractGraph, s::Integer, d::Integer, ::Type{E}) where {E <: AbstractEdge}
    # Returns a list of edges of type `E`. The constructor `E(s, d)` should exist.
end
```

jpfairbanks:
I think we have a broader problem here. When all graphs are SimpleGraphs then there is an equivalence between lists of vertices and paths in the graph. All of the algorithms that compute paths use this equivalence at some point. When the graphs have either metadata attached to the edges (weights, properties, etc.) or multiedges, this equivalence is no longer canonical. You need to know additional information about the graph in order to construct it.

We could either say that Graphs.jl is really a SimpleGraph library and push the handling of this additional information to the downstream packages, or we could find generic ways to providing hooks for downstream libraries to tell us what we need to know. 

The approach used in this PR, introducing `edgetype_to_return(came_from[curr_idx], curr_idx)` as a function that is passed to the implementation makes sense as a hook for downstream libraries to tell us how to make edges. If we enforce that this function be the constructor of the Edge Type of the graph, then downstream libs have less control over what information can be required to make an edge. Having that function be an arbitrary closure makes sense because then weightedgraphs can enclose the graph variable into that closure and look up the weights that they need. This would allow our algorithms to return lists of edges of a type defined in the downstream library. 

I think the core problem is that Graphs.jl implementations assume that there is an equivalence between sequences of vertices and paths in the graph. That core problem is much bigger than `astar` or path finding algorithms in general. We are presented with a choice to expand scope to more graph types or cleanly separate our scope from the applications of graphs.

gdalle:
> I think the core problem is that Graphs.jl implementations assume that there is an equivalence between sequences of vertices and paths in the graph. That core problem is much bigger than astar or path finding algorithms in general. We are presented with a choice to expand scope to more graph types or cleanly separate our scope from the applications of graphs.

Indeed it seems like there are larger things at stake than A*, namely the fate of non-simple graphs. But that would require rethinking a large part of our package, most notably related to edge construction. For instance, the `get_edge(g, s, d)` function suggested by @etiennedeg may no longer be precise enough if there are multi-edges. The same goes for my `edgetype_to_return(s, d)` hack.

However, I think the most urgent task is is to reach a consensus on this PR so that A* finally works on weighted graphs. This is a pretty serious bug that has gone unfixed for a long long time.
In parallel, the rest of the discussion initiated by @jpfairbanks could be moved to a new issue, where it will be more visible and allow the community to weigh in.

jpfairbanks:
Ok, so to get this fixed I think we need to just return E(u,v) or SimpleEdge(u,v) as the fallback.

This delegates the handling of weights to the downstream library, which is fine because only the downstream consumer knows how to handle the additional structure. This doesn't fix or preclude multi graphs in the future. I think that so much of Graphs.jl needs to get fixed for multgraphs that we can keep the status quo until we have an interested person who wants to really dive into multigraphs.

Questions about the PR specifically:

1. We don't have any other reverse dependencies in the tests now and so I am hesitant to add one for this. It would be more consistent to have the test in SWG. There is some risk introduced by the possibility that a change in SWG would break the tests for Graphs.jl

2. We could avoid breaking the signature of `astar_impl!` just by making the `edgetype_to_return` an optional argument with a default of `edgetype(g)`. While the `_impl` functions are generally considered internal to the API, I think that some people use them for performance critical tasks we should take the opportunity to avoid the breakage. 

3. There appear to be some superficial formatting diffs that could be removed from the PR unless they are homogenizing style.

gdalle:
Yeah but the thing is, returning E(u, v) is the current (buggy) behavior, and returning SimpleEdge(u, v) would be breaking. My workaround with the additional type argument keeps the current behavior, but allows the user of eg SWG to specify that they want to return a vector of SimpleEdges. Of course this means users have to change their code to fix the bug, but it is better than us breaking the current behavior. What do you think?

As for your other remarks :

1) Good point, I’ll move the test over to SWG, it makes sense there anyway

2) I was trying to respect the typical order of arguments, as specified in the official style guide, but you think internal stability trumps that?

3) Ok I’ll revert that. But someday we need to put everything through JuliaFormatter during CI because the code base is ugly ^^

etiennedeg:
> EDIT: I just realized this is breaking too, since the constructor E(s, d) of the custom edge type will never be called even when it exists.
   
Yes, maybe we can default `get_edge` to `E(s, d)`, but it's a bit messy to default an API function to a no API function.

> For instance, the get_edge(g, s, d) function suggested by @etiennedeg may no longer be precise enough if there are multi-edges.

If we decide to support multi-graphs, we will need to have a Trait to distinguish between simple graphs and general graphs, so the `get_edge` could just be implemented for simple graphs. (And we would also have a `get_edges` in the API, which could be called by `get_edge`)

In the current PR implementation, the `edgetype_to_return` argument is not optional so it is  heavily breaking.
I'm not found of the flag implementation, even if it provides a workaround, it keeps the buggy behavior with the default arguments, which feels kind of unnecessary, when the graph carry enough information to avoid passing the edgetype. If  no solution is perfect, as you mentioned, I think the cleaner is to deprecate `a_star` to `astar`

jpfairbanks:
> 2) I was trying to respect the typical order of arguments, as specified in the official style guide, but you think internal stability trumps that?

Yes, especially here where there is a good reason for a performance-based usage of the unexpected function.

> I'm not found of the flag implementation, even if it provides a workaround, it keeps the buggy behavior with the default arguments, which feels kind of unnecessary, when the graph carry enough information to avoid passing the edgetype. If no solution is perfect, as you mentioned, I think the cleaner is to deprecate a_star to astar

It makes sense to me for SWG to define a method of `a_star` that passes the right constructor to get back a vector of weighted edges. Then a_star on a weighted graph will do the right thing via multiple dispatch. I see this patch as adding in a hook for client libraries to tell Graphs.jl how to build edges when reconstructing the path. That constructor can lookup the right weight in the input graph via a closure.

etiennedeg:
Ok, but it requires downstream packages to implement it. If we ask users to implement a hack specifically to get  `a_star` working, we can as well ask them to implement a new API method. Graphs must already implement `edgetype(g)`, so it feels weird to ask users what edge it should return when it is supposed to be non ambiguous.
Also, this approach will not work with multi graph, because the edges needs to be gathered during the search (and not after) so its really necessary to get a way to gather the edges in the API.

gdalle:
> In the current PR implementation, the edgetype_to_return argument is not optional so it is heavily breaking.

@etiennedeg I don't think so,  take a look at the default 5-argument method:

https://github.com/gdalle/Graphs.jl/blob/77334fc520dfdcbd166ab2de4b266486891ead17/src/shortestpaths/astar.jl#L97-L105

gdalle:
> It makes sense to me for SWG to define a method of a_star that passes the right constructor to get back a vector of weighted edges.

@jpfairbanks with my suggestion they can do that by specifying `edgetype_to_return` (to be faire, `a_star` will error if they don't anyway). At the moment I constrain it to be a type, but it could be any callable

gdalle:
In terms of order of arguments, I put the `edgetype_to_return` first because otherwise people have to input `distmx` and `heuristic` before getting to the edge type

etiennedeg:
> > Ok, but it requires downstream packages to implement it. If we ask users to implement a hack specifically to get a_star working, we can as well ask them to implement a new API method.
> 
> With the default method we don't, which is why this patch is non-breaking in my opinion.

This is the same if we default `get_edge` to `E(s, d)`, but as I said, it's messy to default an API function to a non API function...
To avoid that, maybe we can avoid implementing a default `get_edge` by checking in the code if `get_edge` is implemented, and if it's not the case, we use `E(s, d)`.

gdalle:
The current buggy behavior may actually be something we wish to keep, unless users explicitly declare that they want to output edges of a certain type (compatible with `E(s, d)`). It prevents surprises where a `SimpleEdge` would pop up in a code that never even mentions it.

etiennedeg:
I don't understand, my proposition does not change the actual behavior until a user redefine something (either `E(s, d)` or `get_edge`)...

gdalle:
Oh, right, I though you wanted `get_edge(g, s, d)` to fall back on `SimpleEdge(s, d)` (as I did in the first version of this PR). But that was breaking

simonschoelly:
I think the solution here is good enough enough, we should definitiely refactor this function in the future though- @gdalle could you try, if you could change SimpleWeightedGraphs, so that it works correctly with the changed function?

gdalle:
> I think the solution here is good enough enough, we should definitiely refactor this function in the future though- @gdalle could you try, if you could change SimpleWeightedGraphs, so that it works correctly with the changed function?

Yeah, I'll polish this PR during the week-end and transfer the test over to SWG.
Mental note: I should also use a concrete type in the A* queue while we're at it (see #129)

gdalle:
So I pushed a new version that adresses most of the previous concerns and provides a satisfying fix until 2.0 comes along:

- The new argument `edgetype_to_return` comes at the end to avoid breaking the signature of internal functions
- JuliaFormatter changes have been reverted
- SimpleWeightedGraphs has been deleted from the test dependencies (I'll add the test there soon)

I think this works but I have two problems left unsolved, for which I need your advice.

- `edgetype_to_return` is still constrained to be a `Type` because I need it to initialize the `total_path` vector. This prevents the use of general closures, as wished by @jpfairbanks.
- @simonschoelly pointed out in #129 that the queue we use in A* has type `PriorityQueue{Integer,T}`. This can hurt performance, but I don't know which concrete type to pick.
  - If we choose `Int`, people who want to work with `BigInt` will see errors.
  - If we use the index type `U` (from `g::AbstractGraph{U}`), I think this can cause other problems: what happens if `U = Int32` but `s::Int64`?

gdalle:
Note however that these problems are not bugs, they only impact generality / performance. As a result, I would be in favor of merging first and improving later

gdalle:
I'd be glad to get feedback on the A* test over in SWG (see [my fork branch](https://github.com/gdalle/SimpleWeightedGraphs.jl/tree/astar_test)).

etiennedeg:
> * If we use the index type `U` (from `g::AbstractGraph{U}`), I think this can cause other problems: what happens if `U = Int32` but `s::Int64`?
If `s` is in the range of `Int32`, then it will be correctly cast. If it is not, then it can't represent a vertex of the graph, so this is a wrong input. `eltype(G)` is definitely the way to go.

src/shortestpaths/astar.jl
```diff
@@ -56,38 +48,37 @@ function a_star_impl!(g, # the graph
...
 """
 function a_star(g::AbstractGraph{U},  # the g
     s::Integer,                       # the start vertex
     t::Integer,                       # the end vertex
     distmx::AbstractMatrix{T}=weights(g),
-    heuristic::Function=n -> zero(T)) where {T, U}
-
-    E = Edge{eltype(g)}
-
+    heuristic=n -> zero(T),
```
simonschoelly, line 67: Why did you remove the restriction here?

src/shortestpaths/astar.jl
```diff
@@ -56,38 +48,37 @@ function a_star_impl!(g, # the graph
...
+    edgetype_to_return::Type{E}=edgetype(g)) where {T, U, E<:AbstractEdge}
```
simonschoelly, line 68: Isn't this a bit dangerous with respect to the other optional argument `heuristic` ?

src/shortestpaths/astar.jl
gdalle: You mean the `heuristic::Function`? Because I thought we might want to use an arbitrary callable struct, not necessarily a function.

src/shortestpaths/astar.jl
gdalle: I agree that it may be dangerous, see [here](https://github.com/JuliaGraphs/Graphs.jl/pull/125#discussion_r878798977) for why I did it. We just have to weigh the pros and cons

gdalle:
Note to self: following the latest updates on #120, I should probably add back the `closed_set`

src/shortestpaths/astar.jl
gdalle: I added it back for now to facilitate merging, we can re-discuss it another time

src/shortestpaths/astar.jl
simonschoelly: Ah I didn't know that this would not for function-like objects. There is also `Base.Callable` but that is also too restrictive in that case.

.gitignore
```diff
@@ -7,4 +7,4 @@ docs/site/
 benchmark/.results/*
 benchmark/.tune.jld
 *.cov
-Manifest.toml
```
simonschoelly, line 10: Is that on purpose? I see you also commited `docs/Manifest.toml`.

gdalle:
I say we leave it as a function for now, this has waited long enough

gdalle:
Yes because a fully reproducible docs build seemed like a good idea. I don’t know whether it is the default practice

simonschoelly (approved):

PR #133, opened 2022-05-22
Add generic graphs to help with testing

simonschoelly:
Our tests currently almost everywhere use `SimpleGraph` and `SimpleDiGraph` for testing methods that work with `AbstractGraph` and `AbstractEege`.. This can lead to issues, when the  methods under test relay on methods that are defined for `SimpleGraph` and `SimpleDiGraph`, but are not part of the Graphs.jl interface.

This PR therefore adds a submodule `Graphs.Test` and the generic graph types `GenericGraph` and `GenericDiGraph` as well as the generic edge type `GenericEdge` that correspond to the interface, but nothing else. The submodule is located within the `src` and not the `test` directory, so that it can also be used in tests of downstream packages.

The names are inspired from similar structures such as `GenericArray` and `GenericOrder` in the Julia package `Test.jl`.

Mutability is currently not covered, as the interface is also not well defined there - this is something we have to tackle in the future.

I also added the generic types to a few existing tests to demonstrate their functionality - in the future we should also modify other tests to use them.

### DIFF
--- a/src/Graphs.jl
+++ b/src/Graphs.jl
@@ -528,6 +528,7 @@ include("vertexcover/degree_vertex_cover.jl")
 include("vertexcover/random_vertex_cover.jl")
 include("Experimental/Experimental.jl")
 include("Parallel/Parallel.jl")
+include("Test/Test.jl")
 
 using .LinAlg
 end # module
--- a/src/Test/Test.jl
+++ b/src/Test/Test.jl
@@ -1,0 +1,82 @@
+
+"""
+    Graphs.Test
+
+A module that provides utilities for testing functions that should work with any `Graphs.AbstractGraph`.
+"""
+module Test
+
+using Graphs
+
+export GenericEdge, GenericGraph, GenericDiGraph
+
+"""
+    GenericEdge <: Graphs.AbstractEdge
+
+An edge type that can  be used to tests functions that relay on the Graphs.jl interface.
+
+"""
+struct GenericEdge{T} <: Graphs.AbstractEdge{T}
+    e::Graphs.SimpleEdge{T}
+end
+
+Graphs.src(e::GenericEdge) = Graphs.src(e.e)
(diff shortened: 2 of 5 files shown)


simonschoelly:
@etiennedeg @gdalle can I have your opinion on that? I think it is something that would be quite useful to have.

gdalle:
@simonschoelly I think it's a really good idea, and it goes some of the way in clarifying interface specification.
If I understand correctly, these types are just wrappers around a `SimpleGraph`/`SimpleEdge`, right? So we'd only be testing whether functions accept the correct (generic) types, but not whether their behavior is correct.
That's already significant progress, but your PR makes me wonder whether we could also create a pathological graph type that satisfies the interface but will explode if we try anything fishy with it.

etiennedeg:
I also think it's good to have.
Regarding your comment on edges, I also think it would be good to have a distinction between directed on non-directed edges, but we should expect the same behavior for `reverse` on edges returned by a non-directed graph, otherwise, this will be a mess. A discussion for 2.0 ?

simonschoelly:
> @simonschoelly I think it's a really good idea, and it goes some of the way in clarifying interface specification. If I understand correctly, these types are just wrappers around a `SimpleGraph`/`SimpleEdge`, right? So we'd only be testing whether functions accept the correct (generic) types, but not whether their behavior is correct. That's already significant progress, but your PR makes me wonder whether we could also create a pathological graph type that satisfies the interface but will explode if we try anything fishy with it.

Actually, as these wrappers should correspond correctly to the interface, using them in tests does ensure that these tests work correctly. And these wrappers will fail if one does something fishy with it, for example trying to access `g.fadjlist`.

gdalle (approved):

gdalle:
I'm in favor of merging, and I opened #224 to keep track of the transition

PR #134, opened 2022-05-22
SimpleEdge comparison

CarloLucibello:
Implement `<` for `SimpleEdge`. This can be useful in different situations (in my case, sort and then compare two arrays of edges).


src/SimpleGraphs/simpleedge.jl
```diff
@@ -32,3 +32,4 @@ SimpleEdge{T}(e::AbstractSimpleEdge) where T <: Integer = SimpleEdge{T}(T(e.src)
 reverse(e::T) where T<:AbstractSimpleEdge = T(dst(e), src(e))
 ==(e1::AbstractSimpleEdge, e2::AbstractSimpleEdge) = (src(e1) == src(e2) && dst(e1) == dst(e2))
 hash(e::AbstractSimpleEdge, h::UInt) = hash(src(e), hash(dst(e), h))
+<(e1::AbstractSimpleEdge, e2::AbstractSimpleEdge) = Tuple(e1) < Tuple(e2)
```
simonschoelly, line 35: Maybe it would be better to define `isless` instead? `<` has `isless` as a fallback anyways.

I would also not relay on `Tuple` here but rather define it in terms of `src` and `dst`. The reason for that is, that we also do it this way for `==` and `hash`. Otherwise it could be, that a custom edge type overrides `Tuple`, which could lead to the weird situation that `e1 < e2` but also `e1 == e2`.

simonschoelly:
I fully support this. It has already come up in the past, but was not merged back then:
https://github.com/sbromberger/LightGraphs.jl/pull/986/files
and maybe also semi related:
https://github.com/sbromberger/LightGraphs.jl/pull/1514

I am also wondering, if we should not define `isless`, `hash` and `equals` for all `AbstractEdge`.

simonschoelly:
@CarloLucibello what is the status here?

src/SimpleGraphs/simpleedge.jl
```diff
@@ -1,4 +1,4 @@
-import Base: Pair, Tuple, show, ==, hash
+import Base: Pair, Tuple, show, ==, hash, <
```
gdalle, line 1:
```suggestion
import Base: Pair, Tuple, show, ==, hash, isless
```

src/SimpleGraphs/simpleedge.jl
gdalle:
```suggestion
isless(e1::AbstractSimpleEdge, e2::AbstractSimpleEdge) = (src(e1) < src(e2)) || ((src(e1) == src(e2)) && (dst(e1) < dst(e2)))
```

test/simplegraphs/simpleedge.jl
```diff
@@ -33,4 +33,11 @@
         @test reverse(ep1) == re
         @test sprint(show, ep1) == "Edge 1 => 2"
     end
+
+    @testset "comparison" begin
+        @test SimpleEdge(1, 2) < SimpleEdge(1, 3) 
+        @test SimpleEdge(1, 2) < SimpleEdge(2, 3)
```
gdalle, line 39:
```suggestion
        @test SimpleEdge(1, 2) < SimpleEdge(2, 3)
        @test SimpleEdge(1, 2) < SimpleEdge(2, 1)
```

gdalle (changes requested):

gdalle:
@CarloLucibello what do you think?

CarloLucibello:
Sorry I dropped this ball, the proposed changes looks fine.

gdalle:
No worries, thanks for setting this up in the first place! Will merge as soon as the CI passes

gdalle (approved):

PR #137, opened 2022-05-24
Rewrite of `edit_distance` with edge costs. fix #111

### LINKED ISSUE #111: Edit distance between 2 equal graphs is not 0.
Hi

I was using the python library networkX  to find isomorphic graphs, but I moved the code to Julia because I need efficiency. My problem is with the file edit_distance.jl.

Check this lines of code

```
graph = ((1, 2), (1, 3), (1, 4), (2, 3), (2, 5), (3, 6), (4, 5), (4, 6), (5, 6))
graph2 = ((1, 2), (1, 3), (1, 4), (2, 3), (2, 5), (3, 6), (4, 5), (4, 6), (5, 6))

G1 = SimpleGraph(6)
for edge in graph
    add_edge!(G1, edge)
end


G2 = SimpleGraph(6)
for edge in graph2
    add_edge!(G2, edge)
end

dist, _ = edit_distance(G1, G2)
```

The outputed distance is 3, but shouln't it be 0?
The two graphs are the same.

Thank you

etiennedeg:
Fix #111

I changed the default costs, as otherwise, the default algorithm as absolutely no meaning. I kept in the file the `MinkowskiCost` and  `BoundedMinkowskiCost` to avoid breakage because they are exported, but we should deprecate these methods, they are useless IMO.

I adapted the code to fix the support of directed graphs and self loops.

I kept the same type of implementation as before with a BFS , but I don't think that's a great idea, this allocates tremendously, even for rather small graphs. It should be good to rewrite this as a backtracking algorithm with some more efficient pruning, or using constraint programming or linear programming. IMO, we should at some point move it to the future GraphOptim, I don't think we have many exponential complexity functions in the code base, and this function merit a slightly more involved implementation.


src/editdist.jl
```diff
@@ -51,51 +58,174 @@ julia> edit_distance(g1, g2)
...
+            isnothing(vertex_subst_cost) &&
+            isnothing(edge_insert_cost) &&
+            isnothing(edge_delete_cost) &&
+            isnothing(edge_subst_cost) &&
+            isnothing(heuristic)
+        heuristic = default_edit_heuristic
+    end
+    if isnothing(vertex_insert_cost)
+        vertex_insert_cost = v -> 0.
+    end
```
simonschoelly, line 80:
```suggestion
    vertex_insert_cost = something(vertex_insert_cost, v -> 0.)
```
might be slightly easier to read. Or use the `@something` macro - but that needs Julia 1.7, so for 1.6 we would need to use `Compat.jl`.

src/editdist.jl
```diff
@@ -51,51 +58,174 @@ julia> edit_distance(g1, g2)
...
+                        edge_subst_cost::Function,
+                        heuristic::Function) where {T <: Integer, U <: Integer}
+
+    isdirected = is_directed(G₁) || is_directed(G₂)
 
-  # A* search heuristic
+    # compute the cost on edges due to associate u1 to v1 and u2 to v2
+    # u2 and v2 can eventually be 0
+    function association_cost(u1, u2, v1, v2)
+        cost = 0
```
simonschoelly, line 123:
```suggestion
        cost = zero(Float64)
```
Might be slightly more type save than an integer.

gdalle:
The tests fail because they now include doctests, it shouldn't be too hard to adapt?

gdalle (approved):

PR #144, opened 2022-06-11
[bugfix] `Base.eltype` works on graph types as well as instances

lmondada:
According to the [Julia documentation](https://docs.julialang.org/en/v1/base/collections/#Base.eltype), `eltype` should be defined on the `::Type{AbstractGraph}` instead of the instances themselves: 

> "The definition eltype(x) = eltype(typeof(x)) is provided for convenience so that instances can be passed instead of types. However the form that accepts a type argument should be defined for new types."

Currently, however:
```julia
julia> eltype(SimpleGraph{Int64})
Any

julia> eltype(SimpleGraph{Int64}())
Int64
```

This PR fixes this to return
```julia
julia> eltype(SimpleGraph{Int64})
Int64
```


test/simplegraphs/simplegraphs.jl
```diff
@@ -4,7 +4,7 @@ import Random
     adjmx1 = [0 1 0; 1 0 1; 0 1 0] # graph
     adjmx2 = [0 1 0; 1 0 1; 1 1 0] # digraph
     # specific concrete generators - no need for loop
-    @test @inferred(eltype(SimpleGraph())) == Int
+    @test @inferred(eltype(SimpleGraph)) == Int
```
etiennedeg, line 7: `SimpleGraph` is not a type, but a convenience constructor for `SimpleGraph{Int64}`, so `eltype` is not changed for this, causing test to fail. Maybe we should define `SimpleGraph = SimpleGraph{Int64}` instead of just defining the constructor...

etiennedeg (commented):
It makes me wonder if it's okay to call this method `eltype` since `AbstractGraph` is not iterable. I don't think it is too bad, we should better not break anything.

test/simplegraphs/simplegraphs.jl
simonschoelly: Indeed, I would also support `Any`. One could maybe argue for `Integer` instead of `Any` though...

The issue here is, that we have defined `eltype(::Type{AbstractGraph}) = _NI("eltype")` , a decision which I am not sure if I like. So maybe we should just check here if an error is being thrown. Or we just delete the test at all, as we are only concerned with concrete subtypes of `AbstractGraph`.

src/SimpleGraphs/simplegraph.jl
```diff
@@ -23,7 +23,7 @@ function SimpleGraph(
     return SimpleGraph{T}(ne, fadjlist)
 end
 
-eltype(x::SimpleGraph{T}) where T = T
+eltype(::Type{SimpleGraph{T}}) where T = T
```
simonschoelly, line 26:
```suggestion
eltype(::Type{<: AbstractSimpleGraph{T}}) where T = T
```

should also work in my opinion. But then we should move that definition to `src/SimpleGraphs/SimpleGraphs.jl` and also remove the similar one for `SimpleDiGraph`

test/simplegraphs/simplegraphs.jl
gdalle: I think we should do
```julia
eltype(::Type{AbstractGraph{T}) where T = T
```
as @simonschoelly suggested above

gdalle:
I removed the implementations for `Simple(Di)Graph` and moved the fallback to `AbstractGraph`. As a result, `eltype` does not need to be part of the interface anymore

src/SimpleGraphs/simplegraph.jl
gdalle: Moved it to AbstractGraph instead

gdalle (approved):

Issue #145, opened 2022-06-13
Add support for induced bipartite subgraph


Codsilla:
As part of a project, I had to use bipartite induced subgraphs of a graph.

Let $G=(V,E) $ be a graph and let $X,Y\subseteq V$ such that $X \cap Y = \emptyset$. The bipartite subgraph of G induced by $X$ and $Y$ is $$(X \cup Y, X \times Y \cap E) $$
In other words, it's the bipartite subgraph made of all the edges of $G$ that are between $X$ and $Y$.
Such a construction is useful in, for example, the search of [vertex separators](https://en.wikipedia.org/wiki/Vertex_separator).

I've made a simple implementation (30 lines or so) for it and I think it might be worth including it.
Before making a pull request, I wanted to see if this is something people might find interesting to have in the package.

etiennedeg:
A function which more generally returns the induced subgraph of a set of vertices could be useful. It could also be a good idea to define views on graphs where we could filter the edges and vertices (by filtering vertices, we would get the induced subgraph), see https://github.com/JuliaGraphs/Graphs.jl/issues/128#issuecomment-1129995566.

gdalle:
I agree that it would be useful functionality. @Codsilla do you want to get started on a PR so we can help?

Codsilla:
> A function which more generally returns the induced subgraph of a set of vertices could be useful.

The function induced_subgraph does just that :)
Views would be nice but are not exactly in the scope of what is proposed.



> I agree that it would be useful functionality. @Codsilla do you want to get started on a PR so we can help?

Just did!

PR #147, opened 2022-06-20
Induced bipartite

Codsilla:
Added support for induced bipartite subgraphs as discussed in this [issue](https://github.com/JuliaGraphs/Graphs.jl/issues/145).

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -728,13 +728,13 @@ function induced_bipartite_subgraph(g::T,X::AbstractVector{U},Y::AbstractVector{
     newIndex = Dict(reverse.(collect(enumerate([X;Y]))))
     
     for x in X
-        for y in neighbors(g,x) ∩ Y
+        for y in outneighbors(g,x) ∩ Y
             add_edge!(G,newIndex[x],newIndex[y])
         end
     end
     
     for y in Y
-        for x in neighbors(g,y) ∩ X
+        for x in outneighbors(g,y) ∩ X
             add_edge!(G,newIndex[y],newIndex[x])
         end
     end
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -708,6 +708,40 @@ egonet(g::AbstractGraph{T}, v::Integer, d::Integer, distmx::AbstractMatrix{U}=we
     g[neighborhood(g, v, d, distmx, dir=dir)]
 
 
+"""
+    induced_bipartite_subgraph(G,X,Y)
+
+Return the bipartite subgraph of `g` induced by the disjoint subsets of vertices X and Y.
+
+"""
+function induced_bipartite_subgraph(g::T,X::AbstractVector{U},Y::AbstractVector{U})  where T <: AbstractGraph where U <: Integer
+    
+    X ∩ Y != [] && throw("X and Y sould not intersect!")
+    !(X ⊆ 1:nv(G) && Y ⊆ 1:nv(g)) && throw("X and Y sould be subsets of the vertices")
+    
+    unique!(X)
+    unique!(Y)
+
+    n = length(X) + length(Y)
+    G = T(n)
+    
+    newIndex = Dict(reverse.(collect(enumerate([X;Y]))))
+    
+    for x in X
+        for y in neighbors(g,x) ∩ Y
+            add_edge!(G,newIndex[x],newIndex[y])
+        end
+    end
+    
+    for y in Y
+        for x in neighbors(g,y) ∩ X
+            add_edge!(G,newIndex[y],newIndex[x])
+        end
+    end
+    G
+end
+
+
 
 """
     compute_shifts(n::Int, x::AbstractArray)
--- a/test/operators.jl
+++ b/test/operators.jl
@@ -308,6 +308,7 @@
         @test sort(vm) == [1:5;]
     end
 
+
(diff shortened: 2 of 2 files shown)


src/operators.jl
```diff
@@ -708,6 +708,40 @@ egonet(g::AbstractGraph{T}, v::Integer, d::Integer, distmx::AbstractMatrix{U}=we
...
 
+"""
+    induced_bipartite_subgraph(G,X,Y)
+
+Return the bipartite subgraph of `g` induced by the disjoint subsets of vertices X and Y.
+
+"""
+function induced_bipartite_subgraph(g::T,X::AbstractVector{U},Y::AbstractVector{U})  where T <: AbstractGraph where U <: Integer
+    
+    X ∩ Y != [] && throw("X and Y sould not intersect!")
```
simonschoelly, line 719: That `throw` here actually does throw a `String`, and not an `Exception` - this is a bit confusing, because Julia allows one to throw any kind of object. To demonstrate this:
```julia
julia> try
          throw("hello")
       catch e
          println(typeof(e))
       end
String
```
Instead you can either use `error("hello")` or use a subtype of `Exception` with `throw`. E.g. `throw(DomainError("hello"))`

src/operators.jl
```diff
@@ -708,6 +708,40 @@ egonet(g::AbstractGraph{T}, v::Integer, d::Integer, distmx::AbstractMatrix{U}=we
     g[neighborhood(g, v, d, distmx, dir=dir)]
 
 
+"""
+    induced_bipartite_subgraph(G,X,Y)
+
+Return the bipartite subgraph of `g` induced by the disjoint subsets of vertices X and Y.
+
+"""
+function induced_bipartite_subgraph(g::T,X::AbstractVector{U},Y::AbstractVector{U})  where T <: AbstractGraph where U <: Integer
```
simonschoelly, line 717: From convention it would be better to use `G` instead of `T` for the graph type. Furthermore, as far as I understand, this method would not work on any kind of `AbstractGraph`, only on `SimpleGraph` and `SimpleDiGraph`. So I would restrict it to these types of graphs.

test/operators.jl
```diff
@@ -318,6 +319,14 @@
         @test @inferred(ndims(g)) == 2
     end
 
+    g10 = complete_graph(10)
+    @testset "Induced bipartite Subgraphs: $g" for g in testgraphs(g10)
+        sg = @inferred(induced_bipartite_subgraph(g, [2,3],[4,5]))
+        @test nv(sg) == 4
+        @test ne(sg) == 4
+        @test is_bipartite(sg)
+    end
```
simonschoelly, line 328: While this is good, I think we need a few more test cases, e.g.
* For empty graphs
* For graphs with self-loops
* For graphs that are not complete graphs
* For directed graphs
* For cases, where an exception is being thrown
* For `U` of different eltype than `Int`

It might also be good to test more  than just some properties, namely if the subgraph is actually the one that we wanted.

src/operators.jl
```diff
@@ -708,6 +708,40 @@ egonet(g::AbstractGraph{T}, v::Integer, d::Integer, distmx::AbstractMatrix{U}=we
...
+    
+    X ∩ Y != [] && throw("X and Y sould not intersect!")
+    !(X ⊆ 1:nv(g) && Y ⊆ 1:nv(g)) && throw("X and Y sould be subsets of the vertices")
+    
+    unique!(X)
+    unique!(Y)
```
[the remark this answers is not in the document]
simonschoelly: It is also not a good practice to modify the arguments passed to a function, without explicitly telling so.

simonschoelly:
Thanks for your PR :) I added a few comments.

src/operators.jl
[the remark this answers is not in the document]
Codsilla: You're right.! I'll change it to throw an error instead

src/operators.jl
Codsilla: I copied the style of `induced_subgraph` that also uses `T` for the graph type. If this is more consistent with the convention, I can change it. Also, we could follow what's been done for `induced_subgraph` which also returns a mapping from the subgraph to the whole graph. This solution would let people use this function for any AbstractGraph (for example, using the mapping to retrieve information in a metagraph)

src/operators.jl
```diff
@@ -708,6 +708,40 @@ egonet(g::AbstractGraph{T}, v::Integer, d::Integer, distmx::AbstractMatrix{U}=we
...
+
+    n = length(X) + length(Y)
+    G = T(n)
+    
+    newIndex = Dict(reverse.(collect(enumerate([X;Y]))))
+    
+    for x in X
+        for y in outneighbors(g,x) ∩ Y
```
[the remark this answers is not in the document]
Codsilla: Good point! I updated it

src/operators.jl
Codsilla: Thank you for the explanation!

src/operators.jl
gdalle: I think following `induced_subgraph` is a good idea for the mapping, for the types I'd go with `G`

test/operators.jl
gdalle: I agree that more tests would be nice

PR #154, opened 2022-07-17
Added strong_product, disjunctive_product, lexicographical_product, h…

dstahlke:
Added most of the graph products from https://en.wikipedia.org/wiki/Graph_product

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -596,6 +596,217 @@ function tensor_product(g::G, h::G) where G <: AbstractGraph
     return z
 end
 
+"""
+    strong_product(g, h)
+
+Return the [strong product](https://en.wikipedia.org/wiki/Strong_product_of_graphs)
+of `g` and `h`.
+
+### Implementation Notes
+Preserves the eltype of the input graph. Will error if the number of vertices
+in the generated graph exceeds the eltype.
+
+# Examples
+```jldoctest
+julia> using Graphs
+
+julia> a = star_graph(3)
+
+julia> b = path_graph(3)
+
+julia> g = strong_product(a, b)
+{9, 8} undirected simple Int64 graph
+
+julia> g == union(cartesian_product(a, b), tensor_product(a, b))
+true
+```
+"""
+function strong_product(g::G, h::G) where G <: AbstractGraph
+    z = G(nv(g) * nv(h))
+    id(i, j) = (i - 1) * nv(h) + j
+    undirected = !is_directed(g)
+    for e1 in edges(g)
+        i1, i2 = Tuple(e1)
+        for e2 in edges(h)
+            j1, j2 = Tuple(e2)
+            add_edge!(z, id(i1, j1), id(i2, j2))
+            if undirected
+                add_edge!(z, id(i1, j2), id(i2, j1))
+            end
+        end
+    end
+    for e in edges(g)
+        i1, i2 = Tuple(e)
+        for j = vertices(h)
+            add_edge!(z, id(i1, j), id(i2, j))
+        end
+    end
+    for e in edges(h)
+        j1, j2 = Tuple(e)
+        for i in vertices(g)
+            add_edge!(z, id(i, j1), id(i, j2))
+        end
+    end
+    return z
+end
+
+"""
+    disjunctive_product(g, h)
+
+Return the [disjunctive product](https://en.wikipedia.org/wiki/Graph_product)
+of `g` and `h`.
+
+### Implementation Notes
+Preserves the eltype of the input graph. Will error if the number of vertices
+in the generated graph exceeds the eltype.
+
+# Examples
+```jldoctest
+julia> using Graphs
+
+julia> a = star_graph(3)
+
+julia> b = path_graph(3)
+
+julia> g = disjunctive_product(a, b)
+{9, 8} undirected simple Int64 graph
+
+julia> complement(g) == strong_product(complement(a), complement(b))
+true
+```
+"""
+function disjunctive_product(g::G, h::G) where G <: AbstractGraph
+    z = G(nv(g) * nv(h))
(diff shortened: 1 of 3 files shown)


test/operators.jl
```diff
@@ -242,6 +242,21 @@
...
+        v2 = nv(gy)
+        e1 = ne(gx)
+        e2 = ne(gy)
+        # Edge counts from https://en.wikipedia.org/wiki/Graph_product
+        @test ne(cartesian_product(gx, gy)) == v1*e2 + v2*e1
+        @test ne(tensor_product(gx, gy)) == 2*e1*e2
+        @test ne(lexicographic_product(gx, gy)) == v1*e2 + e1*v2^2
+        @test ne(strong_product(gx, gy)) == v1*e2 + v2*e1 + 2*e1*e2
+        @test ne(disjunctive_product(gx, gy)) == v1^2*e2 + e1*v2^2 - 2*e1*e2
+    end
```
aurorarossi, line 258: `g` is not used inside the loop. It should be at the place of gx. Moreover this should be done also for gy.
I suggest these changes:

```suggestion
    gx = SimpleGraph(100, 200)
    gy = SimpleGraph(123, 234)
    @testset "Graph product edge counts" for (g1,g2) in zip(testgraphs(gx),testgraphs(gy))
        v1 = nv(g1)
        v2 = nv(g2)
        e1 = ne(g1)
        e2 = ne(g2)
        # Edge counts from https://en.wikipedia.org/wiki/Graph_product
        @test ne(cartesian_product(g1, g2)) == v1*e2 + v2*e1
        @test ne(tensor_product(g1, g2)) == 2*e1*e2
        @test ne(lexicographic_product(g1, g2)) == v1*e2 + e1*v2^2
        @test ne(strong_product(g1, g2)) == v1*e2 + v2*e1 + 2*e1*e2
        @test ne(disjunctive_product(g1, g2)) == v1^2*e2 + e1*v2^2 - 2*e1*e2
    end
```

aurorarossi:
Hi! We recently added a JuliaFormatter and this pull request was made earlier, so it was not aligned. I fixed this problem and now the controls are fine except for one, since the homomorphic product is not tested. It would be great if you could provide one.
Thank you!

src/operators.jl
```diff
@@ -595,6 +595,217 @@ function tensor_product(g::G, h::G) where {G<:AbstractGraph}
     return z
 end
 
+"""
+    strong_product(g, h)
+
+Return the [strong product](https://en.wikipedia.org/wiki/Strong_product_of_graphs)
+of `g` and `h`.
```
simonschoelly, line 602: I think it might be preferable to also document in a few sentences what the strong product is, so that somehow who checks the documentation in the REPL does not have to switch to wikipedia.

src/operators.jl
```diff
@@ -595,6 +595,217 @@ function tensor_product(g::G, h::G) where {G<:AbstractGraph}
...
+julia> b = path_graph(3)
+
+julia> g = strong_product(a, b)
+{9, 8} undirected simple Int64 graph
+
+julia> g == union(cartesian_product(a, b), tensor_product(a, b))
+true
+```
+"""
+function strong_product(g::G, h::G) where {G<:AbstractGraph}
```
simonschoelly, line 623: Does this operator also work for directed graphs? And does it really make sense to define it for all graph types? Maybe we should restrict it to `SimpleGraph` and maybe also `SimpleDiGraph`.

Otherwise it might be a bit confusing if one has, for example, a graph with metadata on its vertices.

src/operators.jl
dstahlke: Done.  I also added descriptions for the existing functions cartesian_product and tensor_product.

src/operators.jl
dstahlke: Though I don't see mention in wikipedia of directed graphs in relation to any of these products, I think most of them would make sense.  The only exception would be the homomorphic product which has (h_1 \nsim h_2) in its definition and there could be controversy over how that should be interpreted for directed graphs.  Also this one is pretty obscure and I suspect nobody would ever want to use it on a directed graph.

Good point on not supporting graphs that have metadata.  What would be the appropriate way to support both SimpleGraph and SimpleDiGraph?  Should I use the common parent class AbstractSimpleGraph?  (Note that the existing functions tensor_product and cartesian_product also have this issue.)

gdalle:
Making a round of unresolved PR conversations

Metadata is not supported anywhere in Graphs.jl, that is why I don't think it's a requirement for graph products.

The real issue is that these products require adding edges, and we can't guarantee this is supported beyond `Simplegraph`s. Furthermore, there is the question of types: what should a product of two `GridGraph`s be for instance? Fall back on a `SimpleGraph` type?

dstahlke:
Is there anything needed from me?  Should I make it take only SimpleGraph arguments rather than AbstractGraph?  And if so, should I make the already existing graph product functions do the same?

gdalle:
I think restricting these new functions to `AbstractSimpleGraph` would make sense for now, but without touching the existing ones (cause that might be breaking)

dstahlke:
I changed SimpleGraph to AbstractSimpleGraph and fixed the doctests.  The doctests had some trivial problems due to lines copied from tensor_product and somehow must not have been running before.

Krastanov:
I have been sorting through stalled and bitrotten PRs over the last hour or two, and I am really pleased that this one is actually in great condition and already reviewed in detail. I gave it a quick final pass as well. 

I will go ahead and merge it when all tests pass. Thank you for the contribution (and the immense patience).

By the way, you would be unsurprised to hear that we are struggling to find enough volunteers to help with reviews. If you want to sing up as a volunteer, do not hesitate to join in reviewing -- it would be extremely valuable for the ecosystem.

PR #155, opened 2022-07-17
Added clique_number, independence_number

dstahlke:
These simple implementations use `maximal_cliques` as the computational engine.  Perhaps there is a more efficient way, but it would be great to at least have these implemented.

### DIFF
--- a/src/community/cliques.jl
+++ b/src/community/cliques.jl
@@ -137,3 +137,112 @@ function maximal_cliques end
     end
     return cliques
 end
+
+"""
+    maximum_clique(g)
+
+Return a vector representing the node indices of a maximum clique
+of the undirected graph `g`.
+
+```jldoctest
+julia> using Graphs
+
+julia> maximum_clique(blockdiag(complete_graph(3), complete_graph(4)))
+4-element Vector{Int64}:
+ 4
+ 5
+ 6
+ 7
+```
+"""
+function maximum_clique end
+# see https://github.com/mauro3/SimpleTraits.jl/issues/47#issuecomment-327880153 for syntax
+@traitfn function maximum_clique(g::AG::(!IsDirected)) where {T, AG<:AbstractGraph{T}}
+    return sort(argmax(length, maximal_cliques(g)))
+end
+
+"""
+    clique_number(g)
+
+Returns the size of the largest clique of the undirected graph `g`.
+
+```jldoctest
+julia> using Graphs
+
+julia> clique_number(blockdiag(complete_graph(3), complete_graph(4)))
+4
+```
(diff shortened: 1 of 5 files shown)


src/community/cliques.jl
```diff
@@ -137,3 +137,112 @@ function maximal_cliques end
...
+julia> maximal_independent_sets(cycle_graph(5))
+5-element Vector{Vector{Int64}}:
+ [5, 2]
+ [5, 3]
+ [2, 4]
+ [1, 4]
+ [1, 3]
+```
+"""
+function maximal_independent_sets end
```
gdalle, line 200: why is this in the cliques file?

test/community/cliques.jl
```diff
@@ -52,4 +52,7 @@
     for g in testgraphs(h)
       @test test_cliques(h, Array[[7, 4, 5], [2, 6], [3, 5], [3, 6], [3, 1]])
     end
+
+    @test clique_number(cycle_graph(11)) == 2
+    @test independence_number(cycle_graph(11)) == 5
```
gdalle, line 57: could you also add tests for the max indep set, perhaps in another file?

test/community/cliques.jl
```diff
@@ -52,4 +52,7 @@
     for g in testgraphs(h)
       @test test_cliques(h, Array[[7, 4, 5], [2, 6], [3, 5], [3, 6], [3, 1]])
     end
+
+    @test clique_number(cycle_graph(11)) == 2
```
gdalle, line 56: might be better to include in `test_cliques`?

gdalle (changes requested):
Thanks for the PR, sorry for the late review

dstahlke:
I've pushed an update that hopefully addresses these concerns.  (I rebased to master - hopefully that doesn't make the diffs too messy.)

After making a proper unit test for independent sets, I found I had to convert AbstractGraph to SimpleGraph in order to call complement.  Performance shouldn't be an issue since this data copy is negligible in comparison to the cost of the maximal_cliques call.  But I don't know if there is some other reason this would be problematic.

    @traitfn function maximum_independent_set(
        g::AG::(!IsDirected)
    ) where {T,AG<:AbstractGraph{T}}
        # Convert to SimpleGraph first because `complement` doesn't accept AbstractGraph.
        return maximum_clique(complement(SimpleGraph(g)))

gdalle:
I would add a warning to the docstring pointing to the conversion, but otherwise I'd say it's okay

Krastanov:
This looks good, and thankfully it has not bitrotten in the last two years. Thank you for the contribution!

Issue #34, opened 2021-11-02
`is_cyclic` for undirected graphs


samwycherley:
Issue previously open at [#1518](https://github.com/sbromberger/LightGraphs.jl/issues/1518) for LightGraphs.jl.

The `is_cyclic` function in LightGraphs.jl works well for directed graphs but for undirected graphs returns `true` for any nonempty graph. The usual definition for undirected graphs is that cycles don't contain repeated links, so 1->2->1 is not considered a cycle (absent multiple edges.)

Useful for me because I want a function that will test whether an (undirected) network is a tree and easiest way to check that is to confirm the graph is acyclic+connected

jpfairbanks:
@samwycherley, Are you interested in finding an algorithm for this? If not I'll close since @mtfishman solved your direct problem. The non-backtracking part makes its a bit harder.

samwycherley:
When I have time, I'll try and work on it and put up a PR. Feel free to close in the meantime.

pgrepds:
Is someone working on this Issue? If not, I would like to implement it as a first issue. If we want to add a generic function `is_tree` as suggested, we might need to consider  the disconnected case as well (a graph could have multiple trees as components -> forest).

gdalle:
@pgrepds feel free to contribute, it is a nice first issue to tackle!

etiennedeg:
[Because of this](https://github.com/sbromberger/LightGraphs.jl/issues/1518#issuecomment-768563457) that say the behavior was intended, wouldn't it be a breaking change ? To clear the confusion, maybe we could deprecate this function , and have a `is_directed_acyclic` that would work only for directed graphs, and a `is_forest` that would work only for undirected graphs ?

natema:
It seems that #168 is ready to be merged to close this. 
The pull request is by a student in [our team at INRIA UCA](https://team.inria.fr/coati/) and we hope to contribute more now that more members of the team are starting to use Julia.

etiennedeg:
closed by #168

PR #157, opened 2022-07-18
fix is_cyclic

aurorarossi:
issue #34, now `is_cyclic` works for undirected connected graphs.

### DIFF
--- a/src/digraph/transitivity.jl
+++ b/src/digraph/transitivity.jl
@@ -154,7 +154,7 @@ julia> collect(edges(transitivereduction(barbell)))
  Edge 6 => 4
 ```
 """
-function transitivereducion end
+function transitivereduction end
 @traitfn function transitivereduction(g::::IsDirected; selflooped::Bool=false)
     scc = strongly_connected_components(g)
     cg = condensation(g, scc)


src/traversals/dfs.jl
```diff
@@ -10,7 +10,7 @@ Return `true` if graph `g` contains a cycle.
 Uses DFS.
 """
 function is_cyclic end
-@traitfn is_cyclic(g::::(!IsDirected)) = ne(g) > 0
+@traitfn is_cyclic(g::::(!IsDirected)) = (ne(g)-(nv(g)-1)) > 0
```
simonschoelly, line 13: I think this not enough, consider the cyclic graph of 5 vertices
```
1  -- 2       4 -- 5
  \  /
    3
```
then `nv(g) == 5` and `ne(g) == 4`, so that `ne(g) - (nv(g) - 1)) == 0`. We therefore probably need do check every connceted component separately.

simonschoelly:
Thanks for your contribution, I think we should indeed change how this function works for undirected graphs. We probably also need to adapt the tests (that are currently failing) and the documentation for the changed behavior.

aurorarossi:
Yes, this was an improvement only for connected graphs. I'm going to extend it to unconnected ones.
Thank you!

etiennedeg:
Hmm, I'm not sure what happened here? Did you try rebasing the PR?

aurorarossi:
yes

etiennedeg:
new PR [here](https://github.com/JuliaGraphs/Graphs.jl/pull/168)

PR #158, opened 2022-07-24
Update CITATION.bib

matbesancon:
because apparently I'm breaking bibtex


gdalle:
> because apparently I'm breaking bibtex
damn right you are

gdalle (approved):

Issue #106, opened 2022-02-16
Feature Request: Iterator for Graph Traversal


kylebeggs:
For some algorithms, it is needed to visit each node and perhaps do some work there and then move on, in other words iterate through them. I implemented this myself (although a rough approach) so I just wanted to open an issue and see if this would be a welcomed addition before I clean up my code and adhere to the contribution guidelines / style.

An example of a need for this is orienting normals on a surface, see [Surface Reconstruction from Unorganized Points](https://graphics.pixar.com/library/Reconstruction/paper.pdf) section 3.3.

gdalle:
Hy @kylebeggs! This is a great idea, and we had been talking about this for a while. There is some related code in https://github.com/JuliaGraphs/Graphs.jl/tree/master/src/Experimental/Traversals, but it is not yet operational. Do you maybe want to take a look and open a PR?

kylebeggs:
Hi @gdalle, yes I will take a look and use what is there already as building blocks. Will probably start working on this in a few weeks as I'm pretty busy right now. Adding to my to-do list.

kylebeggs:
See PR #163

PR #163, opened 2022-08-16
added iterators for dfs and bfs

kylebeggs:
Adds support to iterate through the nodes in a graph in a depth-first or breadth-first manner. (see #106 )


kylebeggs:
This is failing because of other parts of the code, not what is included in this PR. I'm still kinda new to this - what is the procedure when this happens?

simonschoelly:
> The checks are failing because of other parts of the code, not what is included in this PR. I'm still kinda new to this - what is the procedure when this happens?

From the error logs, it looks as if in your PR you used some names that where already used for something else:
```
WARNING: using Traversals.tree in module Main conflicts with an existing identifier.
``

**Comment by kylebeggs** [comment]
Ah, I see. I did use the same identifier as in the Experimental module, but since the goal here is to bring this functionality forefront can I delete what is in Experimental that conflicts?

**Comment by simonschoelly** [comment]
> Ah, I see. I fixed the issue. Tests still show one error but it seems to be in a completely different module

Indeed - apparently the tests are failing in Julia v1.8 -> I will need to investigate here

**Comment by simonschoelly** [comment]
@kylebeggs If you rebase/merge the lastest master in your branch, the tests should pass now.

**Review by etiennedeg — COMMENTED** [review]
Thanks a lot for your work on this, very appreciated.

File: src/traversals/iterators.jl [hunk 1]
```diff
@@ -0,0 +1,122 @@
...
+    source::Int
+end
+
+
+"""
+    mutable struct GraphIteratorState
+
+`GraphIteratorState` is a struct to hold the current state of iteration which is need for Julia's Base.iterate() function.
+"""
+mutable struct GraphIteratorState
```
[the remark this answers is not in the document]
kylebeggs: Sure, makes sense. I'll make the changes.

src/traversals/iterators.jl
```diff
@@ -0,0 +1,122 @@
...
+
+
+"""
+    Base.iterate(t::IteratorAlgorithm)
+
+First iteration to visit each node.
+"""
+function Base.iterate(t::IteratorAlgorithm)
+    visited = falses(nv(t.graph))
+    visited[t.source] = true
```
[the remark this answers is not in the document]
kylebeggs: No, we don't guarantee that. Good catch. I will think of how to fix this and please if you have any suggestions do say. Thanks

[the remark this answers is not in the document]
kylebeggs: For now, I used a Union type. I'm sure its not the best way to do it, but its 'safe'. I was also thinking to make another abstract type of iterator that would encompass all iterators you need to supply a single source node to start. Something like

```Julia
abstract type SingleSourceIterator <: IteratorAlgorithm end
struct BFSIterator <: SingleSourceIterator
    graph::AbstractGraph
    source::Int
end
```

The issue is I'm not too well versed in graphs. I just need to use DFS for one part of my own package and am still getting up to speed with graph algorithms, so I'm not sure what are appropriate names for these types.

src/traversals/iterators.jl
```diff
@@ -0,0 +1,130 @@
+"""
+    abstract type IteratorAlgorithm
+
+`IteratorAlgorithm` is an abstract type which specifies a particular algorithm to use when iterating through a graph.
+"""
+abstract type IteratorAlgorithm end
```
simonschoelly, line 6:
```suggestion
abstract type VertexIterator end
```
I think this makes it more clear, that we iterate over vertices.

kylebeggs: Agreed, I'll update and perhaps wait on thoughts on my reply above before making another commit to possible to include that as well. Like I said in the above reply, I'm not too well verse in graphs so this helps a lot. Thanks

simonschoelly: Also, we might parametrize the Iterator in the type of vertex that we return:
```julia
abstract type VertexIterator{T <: Integer} end
```

src/traversals/iterators.jl
```diff
@@ -0,0 +1,130 @@
...
+    source::Int
+end
+
+
+"""
+    abstract type AbstractGraphIteratorState
+
+`BasicGraphIteratorState` is an abstract type to hold the current state of iteration which is need for Julia's Base.iterate() function.
+"""
+abstract type AbstractGraphIteratorState end
```
simonschoelly, line 66: Is there any advantage of declaring an abstract type here, especially when we only have one subtype?

src/traversals/iterators.jl
```diff
@@ -0,0 +1,130 @@
...
+end
+
+
+"""
+    Base.iterate(t::DFSIterator, state::BasicGraphIteratorState)
+
+Iterator to visit each node in a depth-first manner.
+"""
+function Base.iterate(t::DFSIterator, state::BasicGraphIteratorState)
+    while !isempty(state.queue)
```
simonschoelly, line 99: If I understand this correctly, if the graph is is not connected (or only weakly connected in the directed case), we might not visit all vertices? Is that on purpose?

etiennedeg: It's a good catch but I think it make sense to only visit vertices reachable from the source, and we could have a separate Iterator (with no source) that would call a DFS Iterator on unexplored vertices until all vertices are explored.

src/traversals/iterators.jl
etiennedeg: Well, he applied my review. This leaves an API for Iterators that would store more information, for example current depth of the vertex, visited edges... I think there should be some work to define a clear API for all the types of iterators we would like to support.

kylebeggs: Yeah I think we need an API because I'd also like to implement some iterators for weighted graphs to do minimum spanning trees for instance. I don't know the algorithms for that, but I'd imagine that would require some more information to be stored in the state.

src/traversals/iterators.jl
kylebeggs: The assumption when doing DFS or BFS is you only explore connected vertices, at least that's my understanding (and personal use case)

src/traversals/iterators.jl
kylebeggs: Sure. I didn't realize there are other types of vertices.. can you point me to an example so I can read up

kylebeggs:
@simonschoelly @etiennedeg  I took a new approach to this taking into account some of the comments. Let me know what you think. I converted this PR to a draft to get some feedback before writing the appropriate tests.

src/iterators/bfs.jl
```diff
@@ -0,0 +1,104 @@
...
+5
+3
+1
+4
+5
+2
+```
+"""
+struct BFSIterator{S} <: VertexIterator
+    graph::AbstractGraph
```
thchr, line 36: The type of the graph should be concrete, i.e., change to:
```suggestion
struct BFSIterator{S, T<:AbstractGraph{S}} <: VertexIterator
    graph::T
```

kylebeggs:
@thchr thanks for the suggestion. @simonschoelly @etiennedeg I am opening this back up as I have made some changes. Please review and lets all work together to come up with something to get us started with iterators in this package.

Tortar:
This is very cool, one thing I want to suggest for the bfs part is that to apply the optimization I already applied here: https://github.com/JuliaGraphs/Graphs.jl/pull/250.

src/iterators/bfs.jl
```diff
@@ -0,0 +1,61 @@
+"""
+    BFSIterator
+
+`BFSIterator` is a subtype of [`VertexIterator`](@ref) to iterate through graph vertices using a breadth-first search. A source node(s) is optionally supplied as an `Int` or an array-like type that can be indexed if supplying multiple sources. If no source is provided, it defaults to the first vertex.
```
Tortar, line 4: Let me know if I'm wrong but the `BFSIterator` doesn't seem to allow multiple sources in the current form

src/iterators/bfs.jl
```diff
@@ -0,0 +1,61 @@
...
+
+"""
+    Base.iterate(t::BFSIterator)
+
+First iteration to visit each vertex in a graph using breadth-first search.
+"""
+function Base.iterate(t::BFSIterator)
+    visited = falses(nv(t.graph))
+    visited[t.source] = true
+    return (t.source, VertexIteratorState(visited, [t.source]))
```
Tortar, line 40: In particular this part doesn't seem to work with e.g. a vector of nodes

src/iterators/bfs.jl
kylebeggs: That is correct

kylebeggs:
@Tortar Thanks for the comments. Unfortunately, I don't have time to apply your suggestions right now. Please, make the changes if you are able to. Hopefully one day we can merge this.

Tortar:
@kylebeggs I'm willing to go on with this, I think we have two options: you can list me as a collaborator of your fork so that I can update this branch, otherwise I will need to superseed this pr. I would prefer the first option so that you appear as the main author

kylebeggs:
@Tortar sounds great! I'll add you as a collaborator..

gdalle:
@kylebeggs if we give up on multiple sources for now and focus on a single source, is this in a reviewable / mergeable state?

Tortar:
Hi @gdalle working on extending to multiple sources at the moment, shouldn't be too hard, sorry for the delay, but I forgot about this

Tortar:
okay, now multi source versions for bfs and dfs should work, I don't know the Kruskal algorithm so I didn't update it. Also, tests are failing because of jet complaining about the kruskal implementation

Tortar:
my idea would be to put that algorithm in a separate PR and finish the work for the simpler cases of bfs and dfs, what do you think?

gdalle:
I agree, let's do BFS and DFS first, they're much more widely used

Tortar:
Or if someone could help with that it would be even better

Tortar:
I removed the Kruskal implementation, if you can @kylebeggs I think you should open the new PR though since you made it

kylebeggs:
@Tortar I'm sorry, I'm confused. Why do we need to make a new PR? This one reflects the removal of Kruskal?

Tortar:
I removed it from here so that we can finish the work for bfs/dfs, so maybe not to forgot about the kruskal implementation here: https://github.com/JuliaGraphs/Graphs.jl/pull/163/commits/72c6840dfd75cedc1d1a3286986bd19f3e522b48 you can open up a new pr

Tortar:
anyway @gdalle it should be ready for review!

kylebeggs:
@Tortar I see. You have done a lot here, please feel free to make that PR.

gdalle:
Thank you for the PR! I cleaned up the code a little bit and ensured type stability in the structs

src/iterators/bfs.jl
gdalle: Now it does

src/iterators/bfs.jl
```diff
@@ -0,0 +1,95 @@
...
+
+"""
+    BFSVertexIteratorState
+
+`BFSVertexIteratorState` is a struct to hold the current state of iteration
+in BFS which is needed for the `Base.iterate()` function.
+"""
+mutable struct BFSVertexIteratorState
+    visited::BitVector
+    queue::Vector{Int}
```
gdalle, line 34: Something from https://github.com/JuliaCollections/DataStructures.jl might be more appropriate

src/iterators/dfs.jl
```diff
@@ -0,0 +1,83 @@
...
+
+"""
+    DFSVertexIteratorState
+
+`DFSVertexIteratorState` is a struct to hold the current state of iteration
+in DFS which is needed for the `Base.iterate()` function.
+"""
+mutable struct DFSVertexIteratorState
+    visited::BitVector
+    queue::Vector{Int}
```
gdalle, line 34: Something from https://github.com/JuliaCollections/DataStructures.jl might be more appropriate

src/iterators/bfs.jl
```diff
@@ -0,0 +1,95 @@
...
+    source::S
+end
+
+"""
+    BFSVertexIteratorState
+
+`BFSVertexIteratorState` is a struct to hold the current state of iteration
+in BFS which is needed for the `Base.iterate()` function.
+"""
+mutable struct BFSVertexIteratorState
```
gdalle, line 32: Can you describe the fields of the struct in the docstring?

src/iterators/dfs.jl
```diff
@@ -0,0 +1,83 @@
...
+    source::S
+end
+
+"""
+    DFSVertexIteratorState
+
+`DFSVertexIteratorState` is a struct to hold the current state of iteration
+in DFS which is needed for the `Base.iterate()` function.
+"""
+mutable struct DFSVertexIteratorState
```
gdalle, line 32: Can you describe the fields of the struct in the docstring?

src/iterators/bfs.jl
```diff
@@ -0,0 +1,95 @@
...
+    state = BFSVertexIteratorState(visited, copy(t.source), 1, 1)
+    return (first(t.source), state)
+end
+
+"""
+    Base.iterate(t::BFSIterator, state::VertexIteratorState)
+
+Iterator to visit each vertex in a graph using breadth-first search.
+"""
+function Base.iterate(t::BFSIterator, state::BFSVertexIteratorState)
```
gdalle, line 67: Can you add some comments in this function to explain the steps and help with reviewing?

src/iterators/dfs.jl
```diff
@@ -0,0 +1,83 @@
...
+    state = DFSVertexIteratorState(visited, source_rev)
+    return (last(source_rev), state)
+end
+
+"""
+    Base.iterate(t::DFSIterator, state::VertexIteratorState)
+
+Iterator to visit each vertex in a graph using depth-first search.
+"""
+function Base.iterate(t::DFSIterator, state::DFSVertexIteratorState)
```
gdalle, line 66: Can you add some comments in this function to explain the steps and help with reviewing?

src/iterators/bfs.jl
```diff
@@ -0,0 +1,95 @@
...
+mutable struct BFSVertexIteratorState
+    visited::BitVector
+    queue::Vector{Int}
+    neighbor_idx::Int
+    n_visited::Int
+end
+
+BFSIterator(g::AbstractGraph) = BFSIterator(g, first(vertices(g)))
+
+Base.length(t::BFSIterator) = nv(t.graph)
```
gdalle, line 41: I added a known length but on second thought it may not work for unconnected graphs. In which case the termination criteria in the iteration also need to change

src/iterators/dfs.jl
```diff
@@ -0,0 +1,83 @@
...
+in DFS which is needed for the `Base.iterate()` function.
+"""
+mutable struct DFSVertexIteratorState
+    visited::BitVector
+    queue::Vector{Int}
+end
+
+DFSIterator(g::AbstractGraph) = DFSIterator(g, first(vertices(g)))
+
+Base.length(t::DFSIterator) = nv(t.graph)
```
gdalle, line 39: I added a known length but on second thought it may not work for unconnected graphs. In which case the termination criteria in the iteration also need to change

test/iterators/dfs.jl
```diff
@@ -0,0 +1,44 @@
+@testset "DFSIterator" begin
```
gdalle, line 1: It would be good to test edge cases: empty graph, graph with 2 connected components

test/iterators/bfs.jl
```diff
@@ -0,0 +1,44 @@
+@testset "BFSIterator" begin
```
gdalle, line 1: It would be good to test edge cases: empty graph, graph with 2 connected components

gdalle (changes requested):

src/iterators/bfs.jl
Tortar: I don't think so, removing from the start and the end is O(1) in a Vector if I'm not mistaken

src/iterators/dfs.jl
Tortar: same

src/iterators/bfs.jl
Tortar: yes, we need to assume an unknown length in all cases I guess

src/iterators/dfs.jl
Tortar: same as before

test/iterators/dfs.jl
Tortar: graph with 2 connected component is already tested

Tortar: but adding an empty graph seems good

Tortar:
Thanks @gdalle for the help! But tests are now failing somehow

test/iterators/bfs.jl
Tortar: same

src/iterators/bfs.jl
gdalle: I'm assuming the devil is in the constants, otherwise we wouldn't need stacks and queues

src/iterators/bfs.jl
```diff
@@ -0,0 +1,95 @@
...
+
+"""
+    Base.iterate(t::BFSIterator, state::VertexIteratorState)
+
+Iterator to visit each vertex in a graph using breadth-first search.
+"""
+function Base.iterate(t::BFSIterator, state::BFSVertexIteratorState)
+    graph, visited, queue = t.graph, state.visited, state.queue
+    while !isempty(queue)
+        if state.n_visited == nv(graph)
```
gdalle, line 70: @Tortar this is the termination criterion that is wrong for unconnected graphs

Tortar: mmh, why is it wrong for unconnected graphs? this terminates the iterator only if all the nodes have been visited, it is more of an optimization for connected graph than anything else

src/iterators/bfs.jl
Tortar: will look for it, but I still bet on Vectors :D

src/iterators/bfs.jl
gdalle: yeah you're right sorry

src/iterators/bfs.jl
gdalle: a fair assumption ^^

Tortar: checked...I lost the bet :D

gdalle: woops ^^

gdalle:
tests are fixed now

kylebeggs:
Great to see so much progress on this @Tortar @gdalle. What else is needed to get this merged? There has been a lot of activity, so it is a little difficult to see what exactly is left to do?

gdalle:
Essentially address my unresolved comments in the review: adding more tests and some comments / docstrings describing the logic to make reviewing easier

src/iterators/bfs.jl
Tortar: actually I checked before only on some micro-benchmarks but then I tried in the specific case on a simple graph and I actually saw a regression of 40% so in the end I think we can keep Vectors for now

test/iterators/dfs.jl
Tortar: added it

src/iterators/bfs.jl
Tortar: I tried but maybe it can be improved

Tortar:
I tried to address all review comments by @gdalle, probably docstrings can be improved but hopefully the rest should be okay

src/iterators/bfs.jl
gdalle: since this feature does not exist yet, it's better to have it sooner than to have a fully optimized version maybe never

gdalle: so let's do it this way and we can improve later if necessary

gdalle (approved):
Thank you for the contribution!
There are things we could improve, like using the `eltype` of the graph instead of `Int` or testing on directed graphs, but this has been cooking for far too long and it is a needed feature.

kylebeggs:
Huge thanks to @Tortar !!

PR #164, opened 2022-08-19
Fix loadgraph for Julia v1.8

simonschoelly:
This is a temporary workaround for a bug that only appears from Julia v1.8 on.

See: https://github.com/GunnarFarneback/Inflate.jl/issues/10


GunnarFarneback:
This looks fine for a workaround.

GunnarFarneback:
Probably better to just await https://github.com/GunnarFarneback/Inflate.jl/pull/11.

simonschoelly:
Closed in favor of #166

PR #166, opened 2022-08-20
Bump Inflate dependency to include a bugfix for Julia 1.8.

GunnarFarneback:
Use upstream bugfix instead of the #164 workaround.


simonschoelly (approved):

simonschoelly:
Great, thanks for the quick fix. I am not around today, so if you would have time, could you also trigger a new release?

GunnarFarneback:
It seems unlikely that I would have any powers to trigger a release.

PR #168, opened 2022-08-26
fix is_cyclic 

aurorarossi:
issue https://github.com/JuliaGraphs/Graphs.jl/issues/34, now is_cyclic works for undirected connected and disconnected graphs.

### DIFF
--- a/src/traversals/dfs.jl
+++ b/src/traversals/dfs.jl
@@ -16,20 +16,20 @@ function is_cyclic end
     fill!(vertex_state,unvisited)
     parent = zeros(nv(g))
     for v in vertices(g)
-        S = Vector{T}([v])
-        while !isempty(S)
-            v = pop!(S)
-            if vertex_state[v] != visited
-                vertex_state[v] = visited
-                for u in neighbors(g, v)
-                    parent[u] = v
-                    push!(S,u)
-                end
-            else
-                if length(S) != 0
-                    if v != parent[S[end]]  
-                        return true
+        if vertex_state[v] == unvisited
+            S = Vector{T}([v])
+            while !isempty(S)
+                w = pop!(S)
+                if vertex_state[w] == unvisited
+                    vertex_state[w] = visited
+                    for u in neighbors(g, w)
+                        if parent[w] != u
+                            push!(S,u)
+                            parent[u] = w
+                        end
                     end
+                else
+                    return true
                 end
             end
         end
--- a/src/traversals/dfs.jl
+++ b/src/traversals/dfs.jl
@@ -7,20 +7,34 @@
 Return `true` if graph `g` contains a cycle.
   
 ### Implementation Notes
-Undirect graph: for each connected component of the graph, the algorithm counts how many edges are in surplussurplus to a tree with the same number of nodes.
-Directed graph: uses DFS.
+The algorithm uses a DFS.
 """
 function is_cyclic end
-@traitfn function is_cyclic(g::::(!IsDirected))
-    b=false
-    for vcomponent in connected_components(g)
-        sg, _ = induced_subgraph(g,vcomponent)
-        if (ne(sg)-(nv(sg)-1)) > 0
-            b = true
-            break
+@enum Vertex_state unvisited visited
+@traitfn function is_cyclic(g::AG::(!IsDirected)) where {T, AG<:AbstractGraph{T}}
+    vertex_state = Vector{Vertex_state}(undef, nv(g)) 
+    fill!(vertex_state,unvisited)
+    parent = zeros(nv(g))
+    for v in vertices(g)
+        S = Vector{T}([v])
+        while !isempty(S)
+            v = pop!(S)
+            if vertex_state[v] != visited
+                vertex_state[v] = visited
+                for u in neighbors(g, v)
+                    parent[u] = v
+                    push!(S,u)
+                end
+            else
+                if length(S) != 0
+                    if v != parent[S[end]]  
+                        return true
+                    end
+                end
+            end
         end
     end
-    return b
+    return false
 end
 # see https://github.com/mauro3/SimpleTraits.jl/issues/47#issuecomment-327880153 for syntax
 @traitfn function is_cyclic(g::AG::IsDirected) where {T, AG<:AbstractGraph{T}}
--- a/src/traversals/dfs.jl
+++ b/src/traversals/dfs.jl
@@ -16,10 +16,11 @@ function is_cyclic end
     for v in vertices(g)
         visited[v] && continue
         visited[v] = true
-        S = [(0,v)] 
+        S = [(v,v)] 
         while !isempty(S)
             parent, w = pop!(S)
             for u in neighbors(g, w)
+                u == w && return true # self-loop
                 u == parent && continue
                 visited[u] && return true
                 visited[u] = true
--- a/src/traversals/dfs.jl
+++ b/src/traversals/dfs.jl
@@ -5,7 +5,6 @@
     is_cyclic(g)
 
 Return `true` if graph `g` contains a cycle.
-  
 ### Implementation Notes
 The algorithm uses a DFS.
 """
--- a/src/traversals/dfs.jl
+++ b/src/traversals/dfs.jl
@@ -7,7 +7,7 @@
 Return `true` if graph `g` contains a cycle.
 
 ### Implementation Notes
-The algorithm uses a DFS.
+The algorithm uses a DFS. For directed graphs, self-loops are counted as cycles.
 """
 function is_cyclic end
 @enum Vertex_state unvisited visited
--- a/src/traversals/dfs.jl
+++ b/src/traversals/dfs.jl
@@ -11,8 +11,7 @@ The algorithm uses a DFS.
 function is_cyclic end
 @enum Vertex_state unvisited visited
 @traitfn function is_cyclic(g::AG::(!IsDirected)) where {T, AG<:AbstractGraph{T}}
-    vertex_state = Vector{Vertex_state}(undef, nv(g)) 
-    fill!(vertex_state,unvisited)
+    vertex_state = fill(unvisited, nv(g))
     parent = zeros(T, nv(g))
     for v in vertices(g)
         if vertex_state[v] == unvisited
--- a/src/traversals/dfs.jl
+++ b/src/traversals/dfs.jl
@@ -5,6 +5,7 @@
     is_cyclic(g)
 
 Return `true` if graph `g` contains a cycle.
+
 ### Implementation Notes
 The algorithm uses a DFS.
 """
--- a/src/traversals/dfs.jl
+++ b/src/traversals/dfs.jl
@@ -16,7 +16,7 @@ function is_cyclic end
     parent = zeros(nv(g))
     for v in vertices(g)
         if vertex_state[v] == unvisited
-            S = Vector{T}([v])
+            S = T[v]
             while !isempty(S)
                 w = pop!(S)
                 if vertex_state[w] == unvisited
--- a/src/traversals/dfs.jl
+++ b/src/traversals/dfs.jl
@@ -13,7 +13,7 @@ function is_cyclic end
 @traitfn function is_cyclic(g::AG::(!IsDirected)) where {T, AG<:AbstractGraph{T}}
     vertex_state = Vector{Vertex_state}(undef, nv(g)) 
     fill!(vertex_state,unvisited)
-    parent = zeros(nv(g))
+    parent = zeros(T, nv(g))
     for v in vertices(g)
         if vertex_state[v] == unvisited
             S = T[v]
--- a/src/traversals/dfs.jl
+++ b/src/traversals/dfs.jl
@@ -10,7 +10,7 @@ Return `true` if graph `g` contains a cycle.
 Uses DFS.
 """
 function is_cyclic end
-@traitfn is_cyclic(g::::(!IsDirected)) = ne(g) > 0
+@traitfn is_cyclic(g::::(!IsDirected)) = (ne(g)-(nv(g)-1)) > 0
 # see https://github.com/mauro3/SimpleTraits.jl/issues/47#issuecomment-327880153 for syntax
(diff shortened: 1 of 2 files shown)


src/traversals/dfs.jl
```diff
@@ -5,12 +5,23 @@
...
-Uses DFS.
+Undirect graph: for each connected component of the graph, the algorithm counts how many edges are in surplussurplus to a tree with the same number of nodes.
+Directed graph: uses DFS.
 """
 function is_cyclic end
-@traitfn is_cyclic(g::::(!IsDirected)) = ne(g) > 0
+@traitfn function is_cyclic(g::::(!IsDirected))
+    b=false
+    for vcomponent in connected_components(g)
+        sg, _ = induced_subgraph(g,vcomponent)
```
gdalle, line 17: Is there a way to do this without calling `induced_subgraph`? Judging by the [implementation](https://github.com/JuliaGraphs/Graphs.jl/blob/826b8feac41a2164b05a62296a1f0ac5158d703d/src/operators.jl#L639) this function might lead to significant overhead, so I'm wondering if a more efficient option exists

test/traversals/dfs.jl
```diff
@@ -26,8 +27,9 @@
 
   @testset "is_cyclic" begin
     for g in testgraphs(path_graph(2))
-      @test @inferred(is_cyclic(g))
+      @test @inferred(!is_cyclic(g))
```
gdalle, line 30: Why the `!`?

test/traversals/dfs.jl
```diff
@@ -26,8 +27,9 @@
 
   @testset "is_cyclic" begin
     for g in testgraphs(path_graph(2))
-      @test @inferred(is_cyclic(g))
+      @test @inferred(!is_cyclic(g)) 
       @test @inferred(!is_cyclic(zero(g)))
+      @test @inferred(is_cyclic((gcyclic)))
```
gdalle, line 32: This should probably go outside of the loop since it is a specific graph unrelated to `path_graph(2)`. Check out how `test_graphs` work, it will become clearer

src/traversals/dfs.jl
```diff
@@ -5,12 +5,23 @@
     is_cyclic(g)
 
 Return `true` if graph `g` contains a cycle.
-
+  
 ### Implementation Notes
-Uses DFS.
+Undirect graph: for each connected component of the graph, the algorithm counts how many edges are in surplussurplus to a tree with the same number of nodes.
```
gdalle, line 10: Typo on surplus

gdalle (changes requested):
Thanks for your contribution! This is a good first step, although I have questions about the performance. Can we do better on that point? The rest is mostly details

test/traversals/dfs.jl
aurorarossi: Yes, it should be outside the loop

test/traversals/dfs.jl
aurorarossi: If g is a ```path_graph(2)``` the output of ```is_cyclic(g)``` is false. There are no cycles in a path.  
Hence the test could be: 
```Julia
@test @inferred (is_cyclic(g)==false)
```
or 
 ``` Julia
@test @inferred(!is_cyclic(g)) 
```
I used the second version because I saw it in the code.

**Comment by etiennedeg** [comment]
The most performant would be to use the same BFS than for the directed case, but without taking the edge from where we come from. This allows the algorithm to break early if it meets a cycle. By calling `connected_components`, we do a full BFS everytime...

File: src/traversals/dfs.jl (hunk 1 above)
Comment (aurorarossi) [reply src/traversals/dfs.jl:17]: I found a way. This was the performance before:
```Julia
julia> @benchmark(is_cyclic(gcyclic))
BenchmarkTools.Trial: 10000 samples with 10 evaluations.
 Range (min … max):  1.523 μs … 437.362 μs  ┊ GC (min … max):  0.00% … 99.39%
 Time  (median):     1.623 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   2.018 μs ±  10.616 μs  ┊ GC (mean ± σ):  13.89% ±  2.63%

    ▇█▅▁                                                       
  ▂▇████▇▆▅▄▃▃▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  1.52 μs         Histogram: frequency by time        2.58 μs <

 Memory estimate: 3.23 KiB, allocs estimate: 36.
 ```

With the following new version (count edges without create the subgraph):
```Julia
@traitfn function is_cyclic(g::::(!IsDirected))
    b=false
    for vcomponent in connected_components(g)
        ne=0
        l=length(vcomponent)  
        for i in 1:l
            for j in i+1:l
                if has_edge(g,vcomponent[i],vcomponent[j])
                    ne+=1
                end
            end
        end
        if (ne-(l-1)) > 0
            b = true
            break
        end
    end
    return b
end
```

The performance results:
```Julia
@benchmark(is_cyclic(gcyclic))
BenchmarkTools.Trial: 10000 samples with 160 evaluations.
 Range (min … max):  660.038 ns … 30.517 μs  ┊ GC (min … max):  0.00% … 96.61%
 Time  (median):     682.103 ns              ┊ GC (median):     0.00%
 Time  (mean ± σ):   825.723 ns ±  1.768 μs  ┊ GC (mean ± σ):  14.65% ±  6.64%

  ▅█▇▇▆▅▅▅▄▄▄▃▃▂▁▁▁▁                                           ▂
  ████████████████████▇████▇▇▇█▇▇▆▇▇▆▇▇▆▆▇██▇▇▇█▇▇▇▇▇▆▅▅▅▂▄▄▂▅ █
  660 ns        Histogram: log(frequency) by time      1.01 μs <

 Memory estimate: 1.31 KiB, allocs estimate: 15.
```

What do you think about?

src/traversals/dfs.jl
```diff
@@ -5,12 +5,37 @@
...
 function is_cyclic end
-@traitfn is_cyclic(g::::(!IsDirected)) = ne(g) > 0
+@enum Vertex_state unvisited visited
+@traitfn function is_cyclic(g::AG::(!IsDirected)) where {T, AG<:AbstractGraph{T}}
+    vertex_state = Vector{Vertex_state}(undef, nv(g)) 
+    fill!(vertex_state,unvisited)
+    parent = zeros(nv(g))
+    for v in vertices(g)
+        if vertex_state[v] == unvisited
+            S = Vector{T}([v])
```
Liozou, line 20:
```suggestion
            S = T[v]
```
The `T` specification is actually unnecessary here (but it may be easier to follow the code). But in any case, it is nicer than the `Vector{T}([v])` idiom which allocates two arrays (one for `[v]` and the other for `Vector{T}(...)` (unless the compiler is smart enough to remove one of those, which you should not rely on).

src/traversals/dfs.jl
```diff
@@ -5,12 +5,37 @@
...
+            while !isempty(S)
+                w = pop!(S)
+                if vertex_state[w] == unvisited
+                    vertex_state[w] = visited
```
Liozou, line 24:
```suggestion
                if unvisited[w]
                    unvisited[w] = false
```

Liozou (commented):
Hi and thank for your contribution! Here are some comments, mostly about code style, except for the `@inferred` note in the test which is a faulty test.
I'm not a maintainer of Graphs.jl so feel free to ignore the rest if you disagree, although I do believe it makes the code cleaner.

src/traversals/dfs.jl
```diff
@@ -5,12 +5,37 @@
...
 ### Implementation Notes
-Uses DFS.
+The algorithm uses a DFS.
 """
 function is_cyclic end
-@traitfn is_cyclic(g::::(!IsDirected)) = ne(g) > 0
+@enum Vertex_state unvisited visited
+@traitfn function is_cyclic(g::AG::(!IsDirected)) where {T, AG<:AbstractGraph{T}}
+    vertex_state = Vector{Vertex_state}(undef, nv(g)) 
+    fill!(vertex_state,unvisited)
```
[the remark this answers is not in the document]
aurorarossi: I agree with you but, in a second moment, I would like to improve the part of the code for the directed graphs. In that case there are present tree states. So I wouldn't change this in order to have the same style. What do you think?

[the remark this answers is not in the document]
Liozou: I think I see your point, but I believe having the precise same style in the two functions is not crucial though, because the involved algorithms are not exactly the same anyway.

src/traversals/dfs.jl
aurorarossi: Thank you! I didn't know this

src/traversals/dfs.jl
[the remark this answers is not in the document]
Liozou: If you want to keep it that way, you can replace the two lines by the terser expression:
```suggestion
    vertex_state = fill(unvisited, nv(g))
```

src/traversals/dfs.jl
```diff
@@ -5,12 +5,36 @@
...
+    parent = zeros(nv(g))
```
Liozou, line 16:
```suggestion
    parent = zeros(T, nv(g))
```
otherwise `parent` is a vector of `Float64`, which does not make much sense. It currently works thanks to int->float and float->int conversions, but those are unnecessary

src/traversals/dfs.jl
```diff
@@ -5,12 +5,37 @@
     is_cyclic(g)
 
 Return `true` if graph `g` contains a cycle.
-
+
```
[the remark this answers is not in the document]
Liozou: Ah sorry about that, in that comment here I meant that you should remove the whitespace, but there should still be a newline here, otherwise the `### Implementation Notes` below will not render correctly.

aurorarossi:
I noticed that sometimes some checks fail, in particular the failing is caused by test of `test/traversals/diffusion.jl:120`. It is a file that I didn't modify.

Liozou:
Thanks a lot @aurorarossi for sticking with my many comments! A PR review process can be quite lengthy...
And indeed, after thinking a bit, I believe there is an alternative implementation that bypasses the creation of the extra `parent` array by keeping the only relevant parent directly in the `S` list, like so:
```julia
@traitfn function is_cyclic(g::AG::(!IsDirected)) where {T, AG<:AbstractGraph{T}}
    visited = falses(nv(g)) 
    for v in vertices(g)
        visited[v] && continue
        visited[v] = true
        S = [(v,v)] # the first v is a dummy "parent"
        while !isempty(S)
            parent, w = pop!(S)
            for u in neighbors(g, w)
                u == parent && continue
                visited[u] && return true
                visited[u] = true
                push!(S, (w, u))
            end
        end
    end
    return false
end
```
I believe it should be faster than your proposed PR, but we should benchmark it to check. Would you know how to generate cyclic and acyclic graphs to do so?
If it's faster, the code above can be adapted with your `@enum` if that's preferred. And if it's slower, then your current PR looks good to me as it is.

Regarding the irrelevant failing test, I guess it is a good idea to open an issue to keep track of that separately (but I'm not a maintainer of Graphs.jl so my word is just my opinion).

aurorarossi:
@etiennedeg now the version of the algorithm is faster. I would like to thank @Liozou for the help!

aurorarossi:
@matbesancon

src/traversals/dfs.jl
```diff
@@ -7,10 +7,28 @@
...
+            for u in neighbors(g, w)
+                u == parent && continue
+                visited[u] && return true
+                visited[u] = true
+                push!(S, (w, u))
+            end
+        end
+    end
+    return false
+end
```
simonschoelly, line 31: This looks correct to me, but there is a small edge that we also have to consider: If an undirected graph has a self-loop (i.e. a vertex that is connected to itself), then this should also count as a cycle in my opinion. (or if not, at least we have to document this).

test/traversals/dfs.jl
```diff
@@ -26,9 +27,11 @@
 
   @testset "is_cyclic" begin
     for g in testgraphs(path_graph(2))
-      @test @inferred(is_cyclic(g))
+      @test !@inferred(is_cyclic(g))
       @test @inferred(!is_cyclic(zero(g)))
     end
+    @test @inferred(is_cyclic((gcyclic)))
+    @test @inferred(!is_cyclic(gtree))
```
simonschoelly, line 34: I think there are a few more test-cases we have to cover. First, its often good to use graphs with different eltypes (here you use only `Int`)  as a lot of errors happen when that is not the case. 

Apart from that, it is often good to have these edge cases in your test:
- A graph with zero vertices
- A graph with isolated vertices
- A graph with multiple components (this is already handled by your test case)
- A graph with self-loops (we definitely need this test case here)

simonschoelly:
Can we also add to the function documentation, what the behavior is, for undirected graphs, especially with respect to self-loops?

src/traversals/dfs.jl
```diff
@@ -7,10 +7,28 @@
...
+    visited = falses(nv(g)) 
+    for v in vertices(g)
+        visited[v] && continue
+        visited[v] = true
+        S = [(0,v)] 
+        while !isempty(S)
+            parent, w = pop!(S)
+            for u in neighbors(g, w)
+                u == parent && continue
+                visited[u] && return true
```
simonschoelly, line 24:
```suggestion
        S = [(zero(T),v)] 
        while !isempty(S)
            parent, w = pop!(S)
            for u in neighbors(g, w)
                u == parent && continue
                visited[u] && return true
```
otherwise this will lead to type instability. But also, this relies on the assumption that the value `0` is never used as vertex index. While this might indeed be the case, why not write it as

```suggestion
        S = [(v,v)] 
        while !isempty(S)
            parent, w = pop!(S)
            for u in neighbors(g, w)
                u == w && return true # self-loop
                u == parent && continue
                visited[u] && return true
```
?

simonschoelly (approved):
Awesome, thanks a lot!

simonschoelly:
@gdalle Can I merge this? Github blocks me from merging, as long you have ongoing requests.

PR #169, opened 2022-08-29
fix typo in description of watts_strogatz(n, k, β)

ChristianStehr:
The effect of β is the other way round.

This means that for β = 0 a 1-lattice is created and for β = 1 a completely random graph.

### DIFF
--- a/src/SimpleGraphs/generators/randgraphs.jl
+++ b/src/SimpleGraphs/generators/randgraphs.jl
@@ -239,7 +239,7 @@ where each vertex has exacly `div(k, 2)` neighbors on each side (i.e., `k` or
    connected to some vertex `d`, chosen uniformly at random from the entire
    graph, excluding `s` and its neighbors. (Note that `t` is a valid candidate.)
 
-For `β = 1`, the graph will remain a 1-lattice, and for `β = 0`, all edges will
+For `β = 0`, the graph will remain a 1-lattice, and for `β = 1`, all edges will
 be rewired randomly.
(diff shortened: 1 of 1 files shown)


aurorarossi (approved):

etiennedeg (approved):

etiennedeg:
Thanks!

simonschoelly:
@etiennedeg @ChristianStehr The documentation now says:
```
2. Generate a uniformly random number `r`. If `r ≥ β`, then the edge `(s, t)` is
   left unaltered. Otherwise, the edge is deleted and *rewired* so that `s` is
   connected to some vertex `d`, chosen uniformly at random from the entire
   graph, excluding `s` and its neighbors. (Note that `t` is a valid candidate.)
For `β = 0`, the graph will remain a 1-lattice, and for `β = 1`, all edges will
be rewired randomly.
```

etiennedeg:
Is there a problem about it?

simonschoelly:
Kind, of, because the description was not consistent anymore. In any case, quickly changed that here: https://github.com/JuliaGraphs/Graphs.jl/pull/175.

etiennedeg:
I'm not sure I understand, if $\beta=0$, then $r\geq \beta$ always happen, so edge $(s, t)$ is always left unaltered, then the graph will remain a 1-lattice, so the previous description was correct.

aurorarossi:
I agree (see #175) and I opened a revert pull request #176.

PR #170, opened 2022-09-02
use a stable rng from `StableRNGs` for reproducibility

t-bltg:
Is is currently very hard (impossible ?) to rely on `Graphs` **across different julia versions**: this is observed in [`Plots.jl`](https://github.com/JuliaPlots/Plots.jl)  ci through [downstream test](https://github.com/JuliaPlots/Plots.jl/runs/8154593659?check_suite_focus=true) of [`GraphRecipes.jl`](https://github.com/JuliaPlots/GraphRecipes.jl).

The reason is that the RNG across julia minors is [documented](https://docs.julialang.org/en/v1/stdlib/Random/#Reproducibility) to be unreliable in terms of reproducibility.

This makes `ci` testing via image comparison difficult unless duplicating reference files for different julia versions which is quite cumbersome.

It is advised from julia docs, to use [`StableRNGs.jl`](https://github.com/JuliaRandom/StableRNGs.jl) instead.

PR:
1. use `StableRNG`s in tests;
2. fix methods signatures using random related calls, to take a `rng` and `seed`;
3. ~~update `ci` workflows~~.


simonschoelly:
Thanks for this PR, I think StableRNG might indeed make sense for testing. But this PR would also add it outside of tests? I think we should avoid that, and only use StableRNG for the tests.

We currently might still have some random function that take only a seed instead of a random generator. These functions should be refactored so that they use `rng_from_rng_or_seed` as a temporary workaround - we then might get rid of the seed in the next Major release.

.github/workflows/ci.yml
```diff
@@ -19,41 +19,28 @@ jobs:
...
           - 'nightly'
           - '1'
-          - '1.6'
-        os:
-          - ubuntu-latest
-          - macos-latest
-          - windows-latest
-        arch:
-          - x64
+          - '1.6'  # lowest declared julia compat in `Project.toml`
```
simonschoelly, line 22: Do we need these changes for this PR? Otherwise I think a separate PR for upgrading the ci.yml file might be more appropriate, as this allows us better to spot what has changed when writing release notes.

.github/workflows/ci.yml
```diff
@@ -19,41 +19,28 @@ jobs:
...
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        arch: [x64]
     steps:
-      - uses: actions/checkout@v2
-      - uses: julia-actions/setup-julia@v1
+      - uses: actions/checkout@v3
+      - uses: julia-actions/setup-julia@latest
```
simonschoelly, line 27: I am not sure that using `latest` for an action is a good idea, as this would change the behavior of this script, whenever a new action is released - but then I also don't have much experience with that.

t-bltg: We haven't had any issues with this in repos like `Plots.jl`.

.github/workflows/ci.yml
t-bltg: Dropped workflows changes.

t-bltg:
> I think we should avoid that, and only use StableRNG for the tests.

I've restored the `GLOBAL_RNG` and used a `StableRNG` only in tests.

>  These functions should be refactored so that they use rng_from_rng_or_seed as a temporary workaround

I've propagated a `rng` and `seed` keyword argument to all functions using `rand`, `randperm` or other random methods, and replaced all `getRNG(seed)` calls with `rng = rng_from_rng_or_seed(rng, seed)`.

This took a lot of effort to fix all the tests, but `GraphRecipes` has now reproducible tests.

@simonschoelly, this PR is ready for review / comments.

src/utils.jl
```diff
@@ -56,13 +65,13 @@ always returning a random number generator.
 At least one of these arguments must be `nothing`.
 """
 function rng_from_rng_or_seed(rng::Union{Nothing, AbstractRNG}, seed::Union{Nothing, Integer})
-
     # TODO at some point we might emit a deprecation warning if a seed is specified
-
-    !(isnothing(seed) || isnothing(rng)) && throw(ArgumentError("Cannot specify both, seed and rng"))
```
simonschoelly, line 62: I think it is still better to not to allowing to set both, the `seed` and the `rng` argument - users might be confused if they set both but then the seed wont have any effect. Or do you have a specific reason why you would allow both?

src/Parallel/centrality/betweenness.jl
```diff
@@ -2,9 +2,15 @@ betweenness_centrality(g::AbstractGraph, vs=vertices(g), distmx::AbstractMatrix=
 parallel == :distributed ? distr_betweenness_centrality(g, vs, distmx; normalize=normalize, endpoints=endpoints) : 
 threaded_betweenness_centrality(g, vs, distmx; normalize=normalize, endpoints=endpoints)
     
-betweenness_centrality(g::AbstractGraph, k::Integer, distmx::AbstractMatrix=weights(g); normalize=true, endpoints=false, parallel=:distributed) = 
-parallel == :distributed ? distr_betweenness_centrality(g, sample(vertices(g), k), distmx; normalize=normalize, endpoints=endpoints) :
-threaded_betweenness_centrality(g, sample(vertices(g), k), distmx; normalize=normalize, endpoints=endpoints)
+function betweenness_centrality(
+    g::AbstractGraph, k::Integer, distmx::AbstractMatrix=weights(g);
+    normalize=true, endpoints=false, parallel=:distributed,
+    rng::Union{Nothing, AbstractRNG}=nothing, seed::Union{Nothing, Integer}=-1
```
simonschoelly, line 8:
```suggestion
    rng::Union{Nothing, AbstractRNG}=nothing, seed::Union{Nothing, Integer}=nothing
```
In the previous place where we used `rng_from_rng_or_seed` we set the default or `seed` to `nothing`, we probably should do that here too.

src/Parallel/centrality/betweenness.jl
```diff
@@ -2,9 +2,15 @@ betweenness_centrality(g::AbstractGraph, vs=vertices(g), distmx::AbstractMatrix=
...
+)
+    samples = sample(vertices(g), k; rng=rng, seed=seed)
+    parallel == :distributed ? distr_betweenness_centrality(g, samples, distmx; normalize=normalize, endpoints=endpoints) :
```
simonschoelly, line 11:
```suggestion
    return (parallel == :distributed) ? distr_betweenness_centrality(g, samples, distmx; normalize=normalize, endpoints=endpoints) :
```
We don't have any coding guidelines for Graphs.jl at the moment, but personally I perfer to explicetly use the `return` statement instead of just the last expression in a function. As this line is quite long, it might also be preferable to use an if-else statement instead of the tenary operator.

src/utils.jl
t-bltg: I wanted to be the least disruptive of the tests since there are a lot of line changes in this PR.
This is to retain compatibility with the tests setting an explicit seed (e.g. https://github.com/JuliaGraphs/Graphs.jl/blob/903f309039fa71ae2364aaec2158939f76cf5ccf/test/simplegraphs/generators/randgraphs.jl#L12).

IMO, I do think that giving an `rng` and a valid `seed` is unambiguous, meaning:
1. seed the given `rng` to `seed` (see https://github.com/JuliaGraphs/Graphs.jl/blob/903f309039fa71ae2364aaec2158939f76cf5ccf/src/utils.jl#L72);
2. return this `rng`.

src/Parallel/dominatingset/minimal_dom_set.jl
```diff
@@ -1,5 +1,5 @@
 """
-    dominating_set(g, reps, MinimalDominatingSet(); parallel=:threads, seed=-1)
+    dominating_set(g, reps, MinimalDominatingSet(); parallel=:threads, kw...)
```
simonschoelly, line 2: I think we should mention the role of `rng` (and potentially also `seed` altough we want to phase that one out)  in the documentation.

src/Parallel/dominatingset/minimal_dom_set.jl
```diff
@@ -9,6 +9,15 @@ and return the solution with the fewest vertices.
 used. This implementation is more efficient if `reps` is large.
 - If `seed >= 0`, a random generator of each process/thread is seeded with this value.
 """
-dominating_set(g::AbstractGraph{T}, reps::Integer, alg::MinimalDominatingSet; parallel=:threads, seed=-1) where T <: Integer = 
-Graphs.Parallel.generate_reduce(g, (g::AbstractGraph{T})->Graphs.dominating_set(g, alg; seed=seed), 
-(x::Vector{T}, y::Vector{T})->length(x)<length(y), reps; parallel=parallel)
+function dominating_set(
+    g::AbstractGraph{T}, reps::Integer, alg::MinimalDominatingSet;
+    parallel=:threads, kw...
```
simonschoelly, line 14: Would prefer if we mention `seed` and `rng` here, instead of `kw...`

src/Parallel/centrality/betweenness.jl
t-bltg: > We don't have any coding guidelines for Graphs.jl at the moment

Exactly, and this really a pain.

> As this line is quite long, it might also be preferable to use an if-else statement instead of the tenary operator.

Can we please not discuss cosmetic / syntax changes, as this is always proven to be a waste of everybody's time ?
I've tried to stick to the conventions used in this repo, even if I don't agree with those, and especially because of inconsistencies between source files.
I've used the style / conventions that was used in previous `src/Parallel/...` files, e.g. on master:  https://github.com/JuliaGraphs/Graphs.jl/blob/a3fb98b6a2b07b392b5c690c057f1e5bfac25c9e/src/Parallel/centrality/betweenness.jl#L6.

src/Parallel/dominatingset/minimal_dom_set.jl
t-bltg: I can restore `rng=rng, seed=seed` yes, even if IMO this only leads to code complexity.

src/Parallel/dominatingset/minimal_dom_set.jl
t-bltg: I am going to restore that. I've tried to be consistent with all the docstrings such as https://github.com/JuliaGraphs/Graphs.jl/blob/903f309039fa71ae2364aaec2158939f76cf5ccf/src/dominatingset/minimal_dom_set.jl#L6, but I might have missed some.

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -22,13 +19,16 @@ julia> SimpleGraph(5, 7)
...
+    rng::Union{Nothing, AbstractRNG}=nothing, seed::Union{Nothing, Integer}=-1
+) where T <: Integer
     tnv = T(nv)
     maxe = div(Int(nv) * (nv - 1), 2)
     @assert(ne <= maxe, "Maximum number of edges for this graph is $maxe")
-    ne > div((2 * maxe), 3)  && return complement(SimpleGraph(tnv, maxe - ne, seed=seed))
+    ne > div((2 * maxe), 3)  && return complement(SimpleGraph(tnv, maxe - ne, rng=rng, seed=seed))
 
-    rng = getRNG(seed)
+    rng = rng_from_rng_or_seed(rng, seed)
```
simonschoelly, line 31: If we switch this with the line above, we can pass this  to the `SimpleGraph` call on the line above.

t-bltg: Good point, fixed.

src/utils.jl
```diff
@@ -40,9 +43,15 @@ Sample `k` element from unit range `r` without repetition and eventually excludi
...
-sample(a::AbstractVector, k::Integer; exclude=()) = sample!(getRNG(), collect(a), k; exclude=exclude)
-
-getRNG(seed::Integer=-1) = seed >= 0 ? MersenneTwister(seed) : GLOBAL_RNG
+sample(
+    a::AbstractVector, k::Integer;
+    exclude=(), rng::Union{Nothing, AbstractRNG}=nothing, seed::Union{Nothing, Integer}=-1
+) = sample!(rng_from_rng_or_seed(rng, seed), collect(a), k; exclude=exclude)
+
+function getRNG(seed::Integer=-1)
+    seed >= 0 && seed!(GLOBAL_RNG, seed)
```
simonschoelly, line 52: What is the reason for seeding the `GLOBAL_RNG` instead of returning new random generator? This might lead to surprising as specifying a non-negative integer now changes the state of `GLOBAL_RNG` instead of creating a new independent one as before.

src/Parallel/centrality/betweenness.jl
t-bltg: Replaced all `seed=-1` values to `seed=nothing`, and fixed docstrings.

src/Parallel/dominatingset/minimal_dom_set.jl
t-bltg: This is fixed.

src/Parallel/dominatingset/minimal_dom_set.jl
t-bltg: Fixed.

src/utils.jl
t-bltg: No reason, this is a leftover from the original PR. I will restore that to what was in master.

test/runtests.jl
```diff
@@ -9,6 +9,9 @@ using DelimitedFiles
 using Base64
 using Random
 using Statistics: mean
+using StableRNGs
+
+const RNG = StableRNG(1234)
```
simonschoelly, line 14: Does this `RNG` get re-seeded for every test or testset? Otherwise the order of the tests that use `RNG` might affect the outcome of the tests. maybe it is preferable to use the constructor explicetly in ever test.

t-bltg: Yes, probably using a new rng is better in the tests instead of re-seeding the rng at the beginning of each `test`, will fix.

test/community/assortativity.jl
```diff
@@ -7,14 +7,14 @@ using Random, Statistics
         @test @inferred assortativity(wheel_graph(n)) ≈ -1/3
     end
     @testset "Directed ($seed)" for seed in [1, 2, 3], (n, ne) in [(14, 18), (10, 22), (7, 16)]
-        g = erdos_renyi(n, ne; is_directed=true, seed=seed)
+        g = erdos_renyi(n, ne; is_directed=true, rng=RNG, seed=seed)
```
simonschoelly, line 10:
```suggestion
        g = erdos_renyi(n, ne; is_directed=true, rng=StableRNG(seed))
```
otherwise iterating over `seed` has no purpose.

test/dominatingset/minimal_dom_set.jl
```diff
@@ -2,45 +2,45 @@
  
     g0 = SimpleGraph(0)
     for g in testgraphs(g0)
-        d = @inferred(dominating_set(g, MinimalDominatingSet(); seed=3))
+        d = @inferred(dominating_set(g, MinimalDominatingSet(); rng=RNG, seed=3))
```
simonschoelly, line 5:
```suggestion
        d = @inferred(dominating_set(g, MinimalDominatingSet(); rng=RNG))
```
there does not seem to be any point of specifying a seed, when we also specify `rng`.

test/community/assortativity.jl
t-bltg: It does, since when the `rng` and `seed` leads to seeding the `rng` with this value.

src/utils.jl
t-bltg: Fixed.

test/runtests.jl
t-bltg: Fixed.

test/dominatingset/minimal_dom_set.jl
t-bltg: It seeds the `rng` with this specific value, so it does, even if admittedly, we don't test the numbers stream per se.

t-bltg:
I think I have addressed all the comments. Upstream tests ✔.

src/Parallel/centrality/betweenness.jl
simonschoelly: While I also don't like too waste much time on these, I think suggestions are always allowed - but maybe I should try to make it more clear in my writing that this really was just a suggestion. 

Anyway, as it is really minor and up to taste, lets ignore it for now.

t-bltg: Thanks. This can always be done in a separate PR.

src/utils.jl
simonschoelly: I still would not do it for the following reasons:
* We want to get rid of `seed` eventually, so that seeding the rng will fall to the callers side anyway.
* While `rng` gets mutated anyway, it is still not clear to the caller, that this what is happening.
* Not all subtypes of `rng` allow one to use `seed!`, one example is `RandomDevice`.

It would need much more code to get this running - in fact I implemented that in this PR based on this branch: `https://github.com/t-bltg/Graphs.jl/pull/1` -> for some reason that PR ended up in your fork and no in the main upstream repository though.

t-bltg: Merged.

simonschoelly (approved):
Great, thanks a lot!

t-bltg:
Amazing, thanks for your valuable comments (and obviously for your time !) :tada: .

PR #172, opened 2022-09-12
Coherent return types for Kruskal and Prim

### LINKED ISSUE #130: Coherent return types for Kruskal vs Prim
See https://github.com/JuliaGraphs/SimpleWeightedGraphs.jl/issues/17

When applied on a `SimpleWeightedGraphs`, one function returns a vector of `Edge`, the other a vector of `SimpleWeightedEdge`

Related discussion on Graphs.jl:
- https://github.com/JuliaGraphs/Graphs.jl/issues/120
- https://github.com/JuliaGraphs/Graphs.jl/pull/116
- https://github.com/JuliaGraphs/Graphs.jl/pull/125

aurorarossi:
Now both  algorithms return a `Vector{Edges}`. I chose `Vector{Edges}` instead of `Vector{SimpleWeightedEdge}` because in this way the code is faster. In the case that weights are needed it is easy to query them.
Issue #130

### DIFF
--- a/src/spanningtrees/kruskal.jl
+++ b/src/spanningtrees/kruskal.jl
@@ -14,7 +14,7 @@ function kruskal_mst end
 
     connected_vs = IntDisjointSets(nv(g))
 
-    mst = Vector{edgetype(g)}()
+    mst = Vector{Edge{U}}()
     sizehint!(mst, nv(g) - 1)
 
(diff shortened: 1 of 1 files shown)


aurorarossi:
@matbesancon

aurorarossi:
Ok thanks @simonschoelly !

PR #175, opened 2022-09-25
Fix incorrect inequality in watts_strogatz documentation

simonschoelly:


aurorarossi:
I don't agree with this change. 
In the code there is the following check: 
```jl  
(rand(rng) < β && degree(g, s) < n - 1) || continue
```
which means that if `r=rand(rng)`  is less than `β` then the edge is rewired otherwise (if `r` bigger or equal to `β`) the edge is left unaltered.
Indeed, in the special cases:
- if `β=0` then r is always bigger than 0 that leads to no change in the lattice.
- if `β=1` then r is always less than 1 and the whole graph is rewired.

PR #177, opened 2022-10-02
Add `mycielski` operator

mcognetta:
This adds the `Mycielski` operator ([wiki](https://en.wikipedia.org/wiki/Mycielskian)), a unary graph operator that increases the chromatic number of a graph by 1.

This can be found in other graph libraries: [NetworkX](https://networkx.org/documentation/stable/reference/generated/networkx.generators.mycielski.mycielskian.html#networkx.generators.mycielski.mycielskian), [SageMath](https://doc.sagemath.org/html/en/reference/graphs/sage/graphs/graph_generators.html#sage.graphs.graph_generators.GraphGenerators.MycielskiGraph). 

I wasn't sure if it was best to put it here or in the generators module (which is how it is implemented in NetworkX/SageMath). Happy to move it if necessary.

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -851,3 +851,76 @@ function merge_vertices!(g::Graph{T}, vs::Vector{U} where U <: Integer) where T
 
     return new_vertex_ids
 end
+
+"""
+    mycielski(g)
+
+Returns a graph after applying the Mycielski operator to the input. The Mycielski operator
+returns a new graph with `2n+1` vertices and `3e+n` edges and will increase the chromatic
+number of the graph by 1.
+
+The Mycielski operation can be repeated by using the `iterations` kwarg. Each time, it will
+apply the operator to the previous iterations graph.
+
+For each vertex in the original graph, that vertex and a copy of it is added to the new graph.
+Then, for each edge `(x, y)` in the original graph, the edges `(x, y)`, `(x', y)`, and `(x, y')`
+are added to the new graph, where `x'` and `y'` are the "copies" of `x` and `y`, respectively.
+In otherwords, the original graph is present as a subgraph, and each vertex in the original graph
+is connected to all of it's neighbors' copies. Finally, one last vertex `w` is added to the graph
+and an edge connecting each copy vertex `x'` to `w` is added.
+
+See [Mycielskian](https://en.wikipedia.org/wiki/Mycielskian) for more information.
+
+# Examples
+```jldoctest
+julia> c = CycleGraph(5)
+{5, 5} undirected simple Int64 graph
+
+julia> m = Graphs.mycielski(c)
+{11, 20} undirected simple Int64 graph
+
+julia> collect(edges(m))
+20-element Vector{Graphs.SimpleGraphs.SimpleEdge{Int64}}:
+ Edge 1 => 2
+ Edge 1 => 5
+ Edge 1 => 7
+ Edge 1 => 10
+ Edge 2 => 3
+ Edge 2 => 6
+ Edge 2 => 8
+ Edge 3 => 4
+ Edge 3 => 7
+ Edge 3 => 9
+ Edge 4 => 5
+ Edge 4 => 8
+ Edge 4 => 10
+ Edge 5 => 6
+ Edge 5 => 9
+ Edge 6 => 11
+ Edge 7 => 11
+ Edge 8 => 11
+ Edge 9 => 11
+ Edge 10 => 11
+```
+"""
+@traitfn function mycielski(g::AbstractGraph::(!IsDirected); iterations = 1)
+    ref = g
+    out = deepcopy(g)
+    for _ in 1:iterations
+        N = nv(g)
+        add_vertices!(out, N + 1)
+        w = nv(out)
+        for e in collect(edges(g))
+            x, y = Tuple(e)
+            add_edge!(out, x, y)
+            add_edge!(out, x, y+N)
+            add_edge!(out, x+N, y)
+        end
+
+        for v in 1:N
+            add_edge!(out, v+N, w)
+        end
+        g = out
+    end
+    return out
+end
--- a/test/operators.jl
+++ b/test/operators.jl
@@ -320,4 +320,20 @@
     @testset "Length: $g" for g in testgraphs(SimpleGraph(100))
         @test length(g) == 10000
     end
+
+
+    @testset "Mycielski Operator" begin
+        g = complete_graph(2)
+
+        m = mycielski(g; iterations = 8)
+        @test nv(m) == 767
+        @test ne(m) == 22196
+
+        # check that mycielski preserves triangle-freeness
+        g = complete_bipartite_graph(10, 5)
+        m = mycielski(g)
+        @test nv(m) == 2*15 + 1
+        @test ne(m) == 3*50 + 15
+        @test all(iszero, triangles(m))
+    end
 end
--- a/src/Graphs.jl
+++ b/src/Graphs.jl
@@ -55,6 +55,7 @@ complement, reverse, reverse!, blockdiag, union, intersect,
 difference, symmetric_difference,
 join, tensor_product, cartesian_product, crosspath,
 induced_subgraph, egonet, merge_vertices!, merge_vertices,
+mycielski,
 
 # bfs
 gdistances, gdistances!, bfs_tree, bfs_parents, has_path,


src/operators.jl
```diff
@@ -851,3 +851,76 @@ function merge_vertices!(g::Graph{T}, vs::Vector{U} where U <: Integer) where T
...
+ Edge 5 => 9
+ Edge 6 => 11
+ Edge 7 => 11
+ Edge 8 => 11
+ Edge 9 => 11
+ Edge 10 => 11
+```
+"""
+@traitfn function mycielski(g::AbstractGraph::(!IsDirected); iterations = 1)
+    ref = g
```
aurorarossi, line 907: `ref` is not used in the code except in this line.

src/operators.jl
```diff
@@ -851,3 +851,76 @@ function merge_vertices!(g::Graph{T}, vs::Vector{U} where U <: Integer) where T
...
+    out = deepcopy(g)
```
aurorarossi, line 908: I think that `deepcopy(g)` could slow down and uses a lot of memory especially for big graphs.

src/operators.jl
```diff
@@ -851,3 +851,76 @@ function merge_vertices!(g::Graph{T}, vs::Vector{U} where U <: Integer) where T
...
+    for _ in 1:iterations
+        N = nv(g)
+        add_vertices!(out, N + 1)
+        w = nv(out)
+        for e in collect(edges(g))
+            x, y = Tuple(e)
+            add_edge!(out, x, y)
```
aurorarossi, line 915: This line is useless since these edges have already been copied with the deepcopy.

aurorarossi (commented):
Hello! I'm not a maintainer, I'm just contributing. I wrote some suggestions and comments.

src/operators.jl
```diff
@@ -851,3 +851,76 @@ function merge_vertices!(g::Graph{T}, vs::Vector{U} where U <: Integer) where T
...
+            add_edge!(out, x, y+N)
+            add_edge!(out, x+N, y)
+        end
+
+        for v in 1:N
+            add_edge!(out, v+N, w)
+        end
+        g = out
+    end
+    return out
```
aurorarossi, line 925: I suggest something as the following:
```suggestion
    for _ in 1:iterations
        N = nv(g)
        previous_edges= collect(edges(g))
        add_vertices!(g, N + 1)
        w = nv(g)
        for e in previous_edges
            x=e.src
            y=e.dst
            add_edge!(g, x, y+N)
            add_edge!(g, x+N, y)
        end
        for v in 1:N
            add_edge!(g, v+N, w)
        end
    end
    return g
```

src/operators.jl
mcognetta: Removed, thanks.

src/operators.jl
mcognetta: These operators are not supposed to mutate the input right? I see some have `!` variants, but not all. Perhaps there should be a `mycielski!` and `mycielski(g; iterations) = mycielski!(deepcopy(g); iterations=iterations)` method. WDYT?

So this operator will make at least a 2x larger graph so at least a deep copy's worth of data has to be initialized, and the deep copy allows us to build it in place (in the event of `iterations > 1`, and some other operators also use deep copy. Maybe a shallow copy is ok, but I am not quite sure of the graph internals.

src/operators.jl
mcognetta: Yep, thanks.

src/operators.jl
mcognetta: I used some of these suggestions, but I kept the deep copy as we don't want to mutate the input graph (see above comment).

mcognetta:
@aurorarossi thanks for the comments!

src/operators.jl
aurorarossi: Yes I agree with you. I was only thinking about the case in which we can mutate the input.

src/operators.jl
```diff
@@ -851,3 +851,74 @@ function merge_vertices!(g::Graph{T}, vs::Vector{U} where U <: Integer) where T
...
+ Edge 5 => 6
+ Edge 5 => 9
+ Edge 6 => 11
+ Edge 7 => 11
+ Edge 8 => 11
+ Edge 9 => 11
+ Edge 10 => 11
+```
+"""
+@traitfn function mycielski(g::AbstractGraph::(!IsDirected); iterations = 1)
```
simonschoelly, line 906:
```suggestion
@traitfn function mycielski(g::SimpleGraph; iterations::Int = 1)
```
The problem we have at the moment, that for graphs where we have some kind of edge weights/data, it is not clear what we mean by adding an edge. This is definitely something we should sort out in the future, but in the meantime I would suggest to restrict this function to `SimpleGraph`s.

Also, would suggest to restrict the type of iterations to  to `Int` or `Integer`.

src/operators.jl
simonschoelly: I would also support having both `mycielski!` and `mycielski`. We currently have something similar for `transitiveclosure` and `transitiveclosure!` at the moment.

src/operators.jl
```diff
@@ -851,3 +851,74 @@ function merge_vertices!(g::Graph{T}, vs::Vector{U} where U <: Integer) where T
...
+    out = deepcopy(g)
+    for _ in 1:iterations
```
simonschoelly, line 908:
```suggestion
    for _ in Base.OneTo(iterations)
```
To be honest though, I am not sure if that makes a big difference for the case where `iterations` is not of type `Int`.

src/operators.jl
```diff
@@ -851,3 +851,74 @@ function merge_vertices!(g::Graph{T}, vs::Vector{U} where U <: Integer) where T
 
     return new_vertex_ids
 end
+
+"""
+    mycielski(g)
+
+Returns a graph after applying the Mycielski operator to the input. The Mycielski operator
+returns a new graph with `2n+1` vertices and `3e+n` edges and will increase the chromatic
```
simonschoelly, line 859: I think using the symbol `e` for the number of edges is a bit unusual, when we use `n` for the number of vertices at the same time. Maybe there is  a better way?

src/operators.jl
```diff
@@ -851,3 +851,74 @@ function merge_vertices!(g::Graph{T}, vs::Vector{U} where U <: Integer) where T
...
+number of the graph by 1.
+
+The Mycielski operation can be repeated by using the `iterations` kwarg. Each time, it will
+apply the operator to the previous iterations graph.
+
+For each vertex in the original graph, that vertex and a copy of it are added to the new graph.
+Then, for each edge `(x, y)` in the original graph, the edges `(x, y)`, `(x', y)`, and `(x, y')`
```
simonschoelly, line 866: I think, the usual convention for variables that represent vertices are the symbols `u`, `v` and `w` or `s` and `t` if we talk about some path finding algorithm.

src/operators.jl
```diff
@@ -851,3 +851,74 @@ function merge_vertices!(g::Graph{T}, vs::Vector{U} where U <: Integer) where T
...
+and an edge connecting each copy vertex `x'` to `w` is added.
+
+See [Mycielskian](https://en.wikipedia.org/wiki/Mycielskian) for more information.
+
+# Examples
+```jldoctest
+julia> c = CycleGraph(5)
+{5, 5} undirected simple Int64 graph
+
+julia> m = Graphs.mycielski(c)
```
simonschoelly, line 879:
```suggestion
julia> gm = Graphs.mycielski(g)
```

We usually use `m` to represent some kind of number such as the number of vertices or the number of edges and use symbols such as `g` and `h` for graphs.

src/operators.jl
```diff
@@ -851,3 +851,74 @@ function merge_vertices!(g::Graph{T}, vs::Vector{U} where U <: Integer) where T
...
+ Edge 8 => 11
+ Edge 9 => 11
+ Edge 10 => 11
+```
+"""
+@traitfn function mycielski(g::AbstractGraph::(!IsDirected); iterations = 1)
+    out = deepcopy(g)
+    for _ in 1:iterations
+        N = nv(out)
+        add_vertices!(out, N + 1)
```
simonschoelly, line 910: This might fail, if the graph `eltype` does not support adding that many vertices. This might be especially bad, if we also create a mutating `mycielski!` function, as then this function might fail and leave `g` in some in-between state. So we probably should check at the start of the function if we have enough capacity to add that many vertices.

For the non-mutating `mycielski` function, we might also allow one to provide another larger `eltype` as an optional argument, such that we can retu

src/operators.jl
```diff
@@ -851,3 +851,74 @@ function merge_vertices!(g::Graph{T}, vs::Vector{U} where U <: Integer) where T
...
+        w = nv(out)
+        for e in collect(edges(out))
```
simonschoelly, line 912: Here we allocate an new vector for all edges, this would need quite a lot of space. Of course it is not possible to just iterate over the edges without collecting them, but maybe we can iterate of the first `N` vertices and allocate just a vector of  the outneighbors of each vertex?

test/operators.jl
```diff
@@ -320,4 +320,28 @@
...
+        g = complete_bipartite_graph(10, 5)
+        m = mycielski(g)
+        @test nv(m) == 2*15 + 1
+        @test ne(m) == 3*50 + 15
+        @test all(iszero, triangles(m))
+
+        # ensure it is not done in-place
+        @test nv(g) == 15
+        @test ne(g) == 50
+    end
```
simonschoelly, line 346: These tests are good, but I think we need to cover some additional test cases such as:
- tests for graphs with self-loops
- tests for graphs with zero vertices
- tests for graphs with isolated vertices
- tests for graphs with different eltypes - the `testgraphs` utility function can help here

simonschoelly:
This is an interesting operator, I added a few suggestions.

Krastanov:
This has stalled and needs the original author to address the reviews.

@mcognetta , if you have the time, your contribution would certainly be very much appreciated! For the moment I will mark this as a draft in order to have easier time organizing and sorting my review queue.

mcognetta:
@Krastanov Hi, thanks for bringing this to my attention.

I addressed _most_ of the comments. I am not sure of the right way to handle the potential case where we use the inplace modification version but don't have enough memory to store the result. I'm happy to update this if there is a good way to handle it, but for now it will just crash.

I updated the tests, the doc string, signature, etc. and I also split it into a `mycielski!` and `mycielski` pair. And I addressed the implementation optimization (not materializing the whole vertex/neighbor list at each iteration), but now that I am writing this comment, it occurs to me that this assumes that the graph nodes are densely labeled (like you must have the nodes be 1,2,3,...n).

I also implemented this function in networkx and there was a similar problem, which is why we just dumped the entire vertex set into a list at the start of each iteration, so we could rely on that for the vertex numbering and not the implicit index numbers.

henrik-wolf:
> I am not sure of the right way to handle the potential case where we use the inplace modification version but don't have enough memory to store the result. I'm happy to update this if there is a good way to handle it, but for now it will just crash.

I think the problem is not with memory (as in physical memory on your computer) but with the type of the node ids in the given graph. If you have a `SimpleGraph{UInt8}` it can only have 255 nodes. To me it looks like the final graph will have `2^iterations * (nv(g)+1) - 1` nodes. So you could do
```julia
@assert 2^iterations * (nv(g)+1) - 1 <= typemax(eltype(g)) "eltype too small (or a better error message)"
```

PR #179, opened 2022-10-04
added short paragraph in docs on the dangers of deleting vertices

SuperGrobi:
I recently found the solution to a bug in my code related to the persistence of vertex indices when deleting earlier vertices in the graph. Since this behaviour of "renaming" later vertices to fill the spots of deleted ones is (at least from my point of view) fairly unexpected, I decided to contribute a little paragraph in the documentation.
Let me know what you think about it.
(also... am I requesting the right pull target? Do I need to build something for this to work?)

### DIFF
--- a/docs/src/first_steps/access.md
+++ b/docs/src/first_steps/access.md
@@ -40,7 +40,7 @@ true
 julia> last_added_vertex = nv(g)
 11
 ```
-Note that this index is NOT persistent if nodes added earlier are removed. In this case, the last index of the last node (`nv(g)`) is reassigned to the index recently deleted. As edges are often indexed by vertex indices, one has to be careful when working with edges indexed by nodes as well. An edge added as `add_edge!(g, 3, 11)` can not be expected to always pass the `has_edge(g, 3, 11)` check:
+Note that this index is NOT persistent if vertices added earlier are removed. When `rem_vertex!(g, v)` is called, `v` is "switched" with the last vertex before being deleted. As edges are identified by vertex indices, one has to be careful with edges as well. An edge added as `add_edge!(g, 3, 11)` can not be expected to always pass the `has_edge(g, 3, 11)` check:
 ```julia-repl
 julia> g = SimpleGraph(10)
 {10, 0} undirected simple Int64 graph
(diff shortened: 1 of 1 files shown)


docs/src/first_steps/access.md
```diff
@@ -27,3 +27,42 @@ The following is an overview of functions for accessing graph properties.
...
+julia> g = SimpleGraph(10)
+{10, 0} undirected simple Int64 graph
+
+julia> add_vertex!(g)
+true
+
+julia> last_added_vertex = nv(g)
+11
+```
+Note that this index is NOT persistent if nodes added earlier are removed. In this case, the last index of the last node (`nv(g)`) is reassigned to the index recently deleted. As edges are often indexed by vertex indices, one has to be careful when working with edges indexed by nodes as well. An edge added as `add_edge!(g, 3, 11)` can not be expected to always pass the `has_edge(g, 3, 11)` check:
```
gdalle, line 43:
```suggestion
Note that this index is NOT persistent if vertices added earlier are removed. When `rem_vertex!(g, v)` is called, `v` is "switched" with the last vertex before being deleted. As edges are identified by vertex indices, one has to be careful with edges as well. An edge added as `add_edge!(g, 3, 11)` can not be expected to always pass the `has_edge(g, 3, 11)` check:
```

gdalle (dismissed):

gdalle:
Thanks @SuperGrobi for the suggestion, indeed it is a very frequent pain point so mentioning it in the docs will help lots of people

SuperGrobi:
Thank you very much for the review! I find the whole github-interface somewhat unintuitive, so just to be sure that I am not missing anything substantial: do I need to do something here to accept the requested changes? Or has that to be done by another reviewer with elevated access?

gdalle:
> do I need to do something here to accept the requested changes? Or has that to be done by another reviewer with elevated access?

In this instance I took care of it, but you could also have accepted the modifications because they happen on your branch, not on the main repo. The only time you need someone with merge access is to incorporate your changes into the official code

gdalle (approved):

PR #182, opened 2022-10-08
fixed typos

aurorarossi:
I fixed the following typos:
`ga` in `g`
image
and the repetition 
```Julia
e ∈ edges(g) or e ∈ edges(g)
```
in
```Julia
e in edges(g) or e ∈ edges(g)
```
image


simonschoelly (approved):
Thanks, these changes look good. 

On a side note, it is kind of incorrect, that `edges` calls `has_edge` - this is neither the case for `AbstractGraph` nor for `Simple[Di]Graph`.

It would possibly be worth to think about having an implementation `edges(g::AbstractGraph)` that relies on calls to `has_edge` by default while other types of graphs are free to implement their own more efficient `edges` function.
Also, this approach would not work for multigraphs though.

Issue #103, opened 2022-02-14
[PORT] [BUG] yen_k_shortest_paths always returns at least one path


etiennedeg:
[#1268](https://github.com/sbromberger/LightGraphs.jl/issues/1268)
The bug was fixed in [#1311](https://github.com/sbromberger/LightGraphs.jl/pull/1311) in the dev 2.0 branch of LightGraphs, but never made it  to master. It was fixed by calling Experimental methods, which is problematic. We should probably revisit this issue, and maybe we can work to move some of the Experimental shortest path algorithms to the standard algorithms if they bring improvements.

gdalle:
Fixed by #183

```julia
julia> g = cycle_digraph(10)
{10, 10} directed simple Int64 graph

julia> yen_k_shortest_paths(g, 2, 1, weights(g), 2, maxdist=2)
Graphs.YenState{Int64, Int64}(Int64[], Vector{Int64}[])
```

PR #183, opened 2022-10-14
Solved bug in yen.jl

aurorarossi:
In order to fix the bug reported in the issue  #103 I add in Dijkstra algorithm the optional argument `maxdist`  that specifies the maximum path distance beyond which all path distances are assumed to be infinite.


src/shortestpaths/dijkstra.jl
```diff
@@ -20,19 +20,21 @@ Return a [`Graphs.DijkstraState`](@ref) that contains various traversal informat
...
-`state.pathcounts` holds a vector, indexed by vertex, of the number of shortest paths from the source to that vertex. 
+`state.pathcounts` holds a vector, indexed by vertex, of the number of shortest paths from the source to that vertex.
 The path count of a source vertex is always `1.0`. The path count of an unreached vertex is always `0.0`.
 
-* `trackvertices=false`: If true, 
+* `trackvertices=false`: If true,
 
 `state.closest_vertices` holds a vector of all vertices in the graph ordered from closest to farthest.
 
+* `maxdist::Float64` (default: `Inf`) specifies the maximum path distance beyond which all path distances are assumed to be infinite (that is, they do not exist).
```
gdalle, line 36: Does it have to be a `Float64`? It seems like this could be replaced with `typemax(eltype(distmx))`?

aurorarossi: yes! I have just committed your suggestion. Thank you!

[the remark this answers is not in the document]
gdalle:
```suggestion
* `maxdist` (default: `typemax(T)`) specifies the maximum path distance beyond which all path distances are assumed to be infinite (that is, they do not exist).
```

src/shortestpaths/dijkstra.jl
```diff
@@ -73,7 +75,9 @@ function dijkstra_shortest_paths(
     distmx::AbstractMatrix{T}=weights(g);
     allpaths=false,
     trackvertices=false,
-) where {T<:Real} where {U<:Integer}
+    maxdist::T=typemax(T)
```
gdalle, line 78: Type constraint is unnecessary here

gdalle (changes requested):

gdalle (approved):

Issue #181, opened 2022-10-08
Unnecessary outputs of runtests.jl?


aurorarossi:
Hello! I was wondering if it necessary all the printed stuffs when we run the file `runtests.jl`. On my point of view, in this way is more difficult to see the summary table of the tests. I would print some information only for the failed tests.

PR #185, opened 2022-10-25
Change test outputs 

aurorarossi:
As I wrote in issue #181, the test outputs were hiding the summary table. 
With this pull request I hid the long outputs and I set `verbose = true` to have more readable information in the summary table.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -92,9 +92,9 @@ tests = [
     "experimental/experimental"
 ]
 
-@testset "Graphs" begin
+@testset verbose=true "Graphs" begin
     for t in tests
         tp = joinpath(testdir, "$(t).jl")
         include(tp)
     end
-end
+end;
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -92,7 +92,7 @@ tests = [
     "experimental/experimental"
 ]
 
-@testset verbose=true "Graphs" begin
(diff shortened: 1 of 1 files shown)


aurorarossi:
With this PR the output is the following:
image
I attach two screenshot of the current output:

image

image

aurorarossi:
Another option  removing `verbose = true`  would be:
image

gdalle (approved):

simonschoelly:
@gdalle I would have appreciated if that would have still been left open for discussion.

simonschoelly:
Ah, I realized that I only communicated on slack, so there was no trace of me reviewing

gdalle:
> @gdalle I would have appreciated if that would have still been left open for discussion.

My bad @simonschoelly, I thought it was sufficiently minor to approve directly and I didn't think of any downsides to having more detailed test outputs. I didn't see the discussion on Slack though, I'll go there to check it out. Mea culpa!

simonschoelly:
@gdalle No sorry, this is totally my fault. I had completely forgotten that I never started doing the code review here on github and was still in the progress of trying to figure out in what way the test outputs were changing. Of course it should not be necessary to double check on slack.

gdalle:
Alright then! I haven't been very Slacking recently but it should get better soon :)

PR #186, opened 2022-10-30
Implement `isdigraphical` and fix `isgraphical`

InterdisciplinaryPhysicsTeam:
Hello,

We're the developers of [MultilayerGraphs.jl](https://github.com/JuliaGraphs/MultilayerGraphs.jl).

In this PR we implement the test for graphicality in the directed without loops case (see [Fulkerson–Chen–Anstee theorem](https://en.wikipedia.org/wiki/Fulkerson%E2%80%93Chen%E2%80%93Anstee_theorem)).

The new function is called `isdigraphical`, and we tried to implement it as similarly as we could to the already existing [`isgraphical`](https://github.com/InPhyT/Graphs.jl/blob/4fce15bff67e2574843f773431d9974ad4e088e4/src/connectivity.jl#L718). Implementation differences are highlighted using comments in the code. 

We have already written the docstrings and added tests.

CC: @pitmonticone  @ClaudMor


test/connectivity.jl
```diff
@@ -256,6 +256,14 @@
     @test @inferred(!isgraphical([1, 1, 1]))
     @test @inferred(isgraphical([2, 2, 2]))
     @test @inferred(isgraphical(fill(3, 10)))
+
+    # Test directed simple graphicality
+    for i in 1:3
+      sdg = SimpleDiGraph(10, 90)
+      @test @inferred(isdigraphical(indegree(sdg), outdegree(sdg)))
+    end
+    @test @inferred(!isdigraphical([1,1,1], [1,1,0]))
```
aurorarossi, line 265: The `@inferred` macro checks  the first call, and in this case it checks whether `!` infers correctly, so you should put that method call out of the macro scope.
(also in the test of `is_graphical` is not correct).
```suggestion
    @test !@inferred(isdigraphical([1,1,1], [1,1,0]))
```

test/connectivity.jl
```diff
@@ -256,6 +256,14 @@
     @test @inferred(!isgraphical([1, 1, 1]))
     @test @inferred(isgraphical([2, 2, 2]))
     @test @inferred(isgraphical(fill(3, 10)))
+
+    # Test directed simple graphicality
+    for i in 1:3
+      sdg = SimpleDiGraph(10, 90)
+      @test @inferred(isdigraphical(indegree(sdg), outdegree(sdg)))
+    end
```
aurorarossi, line 264: Since `SimpleDiGraph(10,90)` is always complete, I suggest to remove the loop for:
```suggestion
    sdg = SimpleDiGraph(10, 90)
    @test @inferred(isdigraphical(indegree(sdg), outdegree(sdg)))
```

aurorarossi (changes requested):
Thank you for the contribution!

InterdisciplinaryPhysicsTeam:
Done @aurorarossi, 

Thanks for your suggestions. 

CC: @ClaudMor @pitmonticone

aurorarossi (approved):

aurorarossi:
This branch has conflicts because we recently introduced a Julia Formatter. Can you do a merge with Graphs.jl/master? And format the files you changed? I tried to fix it but I don't have permissions to push to your branch. Thanks!

src/connectivity.jl
```diff
@@ -733,3 +733,66 @@ function isgraphical(degs::Vector{<:Integer})
...
+```math
+\\sum_{i=1}^{r} a_i \\leq \\sum_{i=r+1}^n min(r-1,b_i) + \\sum_{i=r+1}^n min(r,b_i)
+```
+for each integer r <= n-1
+"""
+function isdigraphical(indegree_sequence::Vector{<:Integer}, outdegree_sequence::Vector{<:Integer} )
+
+    n = length(indegree_sequence)
+
+    n == length(outdegree_sequence) || return false
```
simonschoelly, line 757: Wouldn't it be more appropriate to throw an exception in this case?

src/connectivity.jl
```diff
@@ -733,3 +733,66 @@ function isgraphical(degs::Vector{<:Integer})
...
+
+
+### Implementation Notes
+According to Fulkerson–Chen–Anstee theorem, a sequence ``\\{(a_1, b_1), ...,(a_n, b_n)\\}`` (sorted in descending order of a) is graphic iff the sum of vertex degrees is even and  ``\\sum_{i = 1}^{n} a_i = \\sum_{i = 1}^{n} b_i\\}`` and the sequence obeys the property -
+```math
+\\sum_{i=1}^{r} a_i \\leq \\sum_{i=r+1}^n min(r-1,b_i) + \\sum_{i=r+1}^n min(r,b_i)
+```
+for each integer r <= n-1
+"""
+function isdigraphical(indegree_sequence::Vector{<:Integer}, outdegree_sequence::Vector{<:Integer} )
```
simonschoelly, line 753:
```suggestion
function isdigraphical(indegree_sequence::AbstractVector{<:Integer}, outdegree_sequence::AbstractVector{<:Integer} )
```

src/connectivity.jl
```diff
@@ -733,3 +733,66 @@ function isgraphical(degs::Vector{<:Integer})
...
+    @inbounds for i = 1:n
+        mindeg[i] = min(i, sorted_outdegree_sequence[i])
+    end
+    cum_min = sum(mindeg) =#
+    # Similarly for `outdegree_min_sum`.
+
+    @inbounds for r = 1:(n - 1)
+
+        indegree_sum += sorted_indegree_sequence[r]
+        outdegree_min_sum = sum([min(sorted_outdegree_sequence[i], r-1) for i in 1:r])
```
simonschoelly, line 786: Here we allocate a new Vector in each iteration of the loop. Can we do this differently?

test/connectivity.jl
```diff
@@ -256,6 +256,12 @@
     @test @inferred(!isgraphical([1, 1, 1]))
     @test @inferred(isgraphical([2, 2, 2]))
     @test @inferred(isgraphical(fill(3, 10)))
+
+    # Test directed simple graphicality
+    sdg = SimpleDiGraph(10, 90)
+    @test @inferred(isdigraphical(indegree(sdg), outdegree(sdg)))
+    @test !@inferred(isdigraphical([1,1,1], [1,1,0]))
+
```
simonschoelly, line 264: Can we also have a few edge cases here? Such as
- Empty vectors
- Graphs that only have a realization if there is a self-loop.
- Graphs where the sum of the in and out degree sequences if the same, but there is still no valid realization.

InterdisciplinaryPhysicsTeam:
> This branch has conflicts because we recently introduced a Julia Formatter. Can you do a merge with Graphs.jl/master? And format the files you changed? I tried to fix it but I don't have permissions to push to your branch. Thanks!

@aurorarossi, we have just formatted the relevant files.

src/connectivity.jl
InterdisciplinaryPhysicsTeam: Hi @simonschoelly, we've just used the same convention adopted for `isgraphical`.

src/connectivity.jl
InterdisciplinaryPhysicsTeam: Hi @simonschoelly, we've just used the same convention adopted for `isgraphical`.

src/connectivity.jl
InterdisciplinaryPhysicsTeam: Done, but that is in contradiction with the @aurorarossi's request.

src/connectivity.jl
InterdisciplinaryPhysicsTeam: We were not able to use the same strategy adopted in `isgraphical`. For further details see in the comments.

src/connectivity.jl
InterdisciplinaryPhysicsTeam: Ops sorry @aurorarossi, we've just noticed that Graphs.jl uses two different conventions for the code and its docstring. 

We're going to proceed by changing the signature according to @simonschoelly's suggestion then.

test/connectivity.jl
InterdisciplinaryPhysicsTeam: We've just noticed that `isgraphical` and `isdigraphical` don't work properly in the edge cases you mentioned. 

For example: 
- `isgraphical(Integer[])` returns `true` even if it's not possible to realise a **simple undirected** graph (no self-loops, no multi-edges). It requires a check such as `!isempty(degs) || return false`; 
- `isdigraphical(Integer[], Integer[])` returns `true` even if it's not possible to realise a **simple directed** graph (no self-loops, no multi-edges). It requires a check such as `!(isempty(indegree_sequence) || isempty(indegree_sequence)) || return false`; 
- `isgraphical([2])` returns `true` even if it's not possible to realise a **simple undirected** graph (no self-loops, no multi-edges) with just a vertex with $d(v)=2$.
- `isdigraphical([1],[1])` returns `true` even if it's not possible to realise a **simple directed** graph (no self-loops, no multi-edges) with just a vertex with $d_{in}(v)=d_{out}(v)=1$.

src/connectivity.jl
InterdisciplinaryPhysicsTeam: Yes, [done](https://github.com/JuliaGraphs/Graphs.jl/pull/186/commits/045b8b55533118795c9e21482ad954a660c9f4a1).

simonschoelly:
@InterdisciplinaryPhysicsTeam Now there is a lot of noise in this PR because of the formatting changes. Perhaps if you rebase on the latest master branch, they will go away?

InterdisciplinaryPhysicsTeam:
Hello @simonschoelly,

We rebased as you asked, updated the `isgraphical` and `isdigraphical` functions. 

Unfortunately tests seem to fail because of the formatting check, but that has nothing to do with the files we changed so we won't fix it.  

@pitmonticone @ClaudMor

InterdisciplinaryPhysicsTeam:
Hello @simonschoelly @aurorarossi @gdalle,

Since we are not maintainers, we need our fork to be on par with Graphs.jl's master branch in order to open new pull requests. Anyway, we cannot do it until this is merged.

May anyone with write permission review and merge this?

src/connectivity.jl
```diff
@@ -748,3 +760,73 @@ function isgraphical(degs::Vector{<:Integer})
...
+    isdigraphical(indegree_sequence, outdegree_sequence)
+
+Check whether the given indegree sequence and outdegree sequence are digraphical, that is whether they can be the indegree and outdegree sequence of a digraph.
+
+### Implementation Notes
+According to Fulkerson-Chen-Anstee theorem, a sequence ``\\{(a_1, b_1), ...,(a_n, b_n)\\}`` (sorted in descending order of a) is graphic iff the sum of vertex degrees is even and  ``\\sum_{i = 1}^{n} a_i = \\sum_{i = 1}^{n} b_i\\}`` and the sequence obeys the property -
+```math
+\\sum_{i=1}^{r} a_i \\leq \\sum_{i=r+1}^n min(r-1,b_i) + \\sum_{i=r+1}^n min(r,b_i)
+```
+for each integer r <= n-1.
```
simonschoelly, line 774: Do you have a source for these conditions? Because the ones one Wikipedia looks differently.

I also notices that this algorithm does return incorrect results in some cases:
```julia
# would only work with a self-loop, but we don't allow these, right?
julia> isdigraphical([1],[1])
true

# would require multi-edges
julia> isdigraphical([5],[5])
true

# but this should also be true if we allow self-loops.
julia> isdigraphical([1,0],[1,0])
false
```

simonschoelly: Also, I was wondering if latex heavy docstrings are a good idea? We don't need to change it in this PR, as we always can do it latter, the issue is just that in the REPL it looks like this:
```julia
help?> isdigraphical
search: isdigraphical

  isdigraphical(indegree_sequence, outdegree_sequence)

  Check whether the given indegree sequence and outdegree sequence are digraphical, that is whether
  they can be the indegree and outdegree sequence of a digraph.

  Implementation Notes
  ––––––––––––––––––––––

  According to Fulkerson-Chen-Anstee theorem, a sequence \{(a_1, b_1), ...,(a_n, b_n)\} (sorted in
  descending order of a) is graphic iff the sum of vertex degrees is even and \sum_{i = 1}^{n} a_i =
  \sum_{i = 1}^{n} b_i\} and the sequence obeys the property -

  \sum_{i=1}^{r} a_i \leq \sum_{i=r+1}^n min(r-1,b_i) + \sum_{i=r+1}^n min(r,b_i)

  for each integer r <= n-1.

  See also: isgraphical
```

src/connectivity.jl
```diff
@@ -728,17 +728,29 @@ According to Erdös-Gallai theorem, a degree sequence ``\\{d_1, ...,d_n\\}`` (so
...
+
+See also: [`isdigraphical`](@ref)
 """
-function isgraphical(degs::Vector{<:Integer})
+function isgraphical(degs::AbstractVector{<:Integer})
+    # Check whether the degree sequence is empty
+    !isempty(degs) || return true
+    # Check whether the degree sequence has only non-negative values
+    all(degs .>= 0) ||
+        throw(ArgumentError("The degree sequence must contain non-negative integers only."))
```
simonschoelly, line 740: Wouldn't it be more reasonable to just return `false` here? Of course if that is so, we should probably do the same thing for `isdigraphical`.

src/connectivity.jl
```diff
@@ -748,3 +760,73 @@ function isgraphical(degs::Vector{<:Integer})
     end
     return true
 end
+
+"""
+    isdigraphical(indegree_sequence, outdegree_sequence)
+
+Check whether the given indegree sequence and outdegree sequence are digraphical, that is whether they can be the indegree and outdegree sequence of a digraph.
+
```
simonschoelly, line 768: We should probably mention here that
* The digraph cannot have self-loops
* `indegree_sequence` and `outdegree_sequence`  are not independent, i.e. if entry `i` of `indegree_sequence`  has value `a_i` and entry `i` of `outdegree_sequence` has value `b_i` then there will be a vertex with indegree `a_i` and outdegree `b_i` in the realization. (maybe there is an easier way to express this).

src/connectivity.jl
```diff
@@ -748,3 +760,73 @@ function isgraphical(degs::Vector{<:Integer})
...
+    # Check whether the degree sequences have only non-negative values
+    all(indegree_sequence .>= 0) || throw(
+        ArgumentError("The indegree sequence must contain non-negative integers only.")
+    )
+    all(outdegree_sequence .>= 0) || throw(
+        ArgumentError("The outdegree sequence must contain non-negative integers only.")
+    )
+
+    n = length(indegree_sequence)
+    n == length(outdegree_sequence) || return false
```
simonschoelly, line 797: We already check this on line 783.

simonschoelly (changes requested):
> Hello @simonschoelly @aurorarossi @gdalle,
> 
> Since we are not maintainers, we need our fork to be on par with Graphs.jl's master branch in order to open new pull requests. Anyway, we cannot do it until this is merged.
> 
> May anyone with write permission review and merge this?

I am not sure,  what you mean, by not being able to open new pull requests?

In any case, I think there are still some bugs here, that need to be fixed before we can merge.

src/connectivity.jl
InterdisciplinaryPhysicsTeam: We thought that `false` was reserved for when "the degree sequence is valid but not graphical", while errors should be thrown if the degree sequence is not even valid.

If it isn't appropriate we will change the behavior.

src/connectivity.jl
InterdisciplinaryPhysicsTeam: We modified the docstring trying to include your suggestions. Thanks.

src/connectivity.jl
InterdisciplinaryPhysicsTeam: We removed the redundant check and moved the definition of the variable `n` to the top of the function.

InterdisciplinaryPhysicsTeam:
Hi @simonschoelly ,

We realized that "not being able to open new pull requests" was due to a mistake of ours, we apologize for that.

We hope that the fixes above comply with your comments.

src/connectivity.jl
```diff
@@ -748,3 +760,71 @@ function isgraphical(degs::Vector{<:Integer})
...
+    isdigraphical(indegree_sequence, outdegree_sequence)
+
+Check whether the given indegree sequence and outdegree sequence are digraphical, that is whether they can be the indegree and outdegree sequence of a simple digraph (i.e. a directed graph with no loops). This implies that `indegree_sequence` and `outdegree_sequence` are not independent, as their elements respectively represent the indegrees and outdegrees that the vertices shall have.
+
+### Implementation Notes
+According to Fulkerson-Chen-Anstee theorem, a sequence ``\\{(a_1, b_1), ...,(a_n, b_n)\\}`` (sorted in descending order of a) is graphic iff ``\\sum_{i = 1}^{n} a_i = \\sum_{i = 1}^{n} b_i\\}`` and the sequence obeys the property -
+```math
+\\sum_{i=1}^{r} a_i \\leq \\sum_{i=r+1}^n min(r-1,b_i) + \\sum_{i=r+1}^n min(r,b_i)
+```
+for each integer r <= n-1.
```
simonschoelly, line 774:
```suggestion
\\sum_{i=1}^{r} a_i \\leq \\sum_{i=1}^n min(r-1,b_i) + \\sum_{i=r+1}^n min(r,b_i)
```
for each integer 1 <= r <= n. 
```

File: src/connectivity.jl [hunk 12]
```diff
@@ -748,3 +760,71 @@ function isgraphical(degs::Vector{<:Integer})
...
+
+"""
+    isdigraphical(indegree_sequence, outdegree_sequence)
+
+Check whether the given indegree sequence and outdegree sequence are digraphical, that is whether they can be the indegree and outdegree sequence of a simple digraph (i.e. a directed graph with no loops). This implies that `indegree_sequence` and `outdegree_sequence` are not independent, as their elements respectively represent the indegrees and outdegrees that the vertices shall have.
+
+### Implementation Notes
+According to Fulkerson-Chen-Anstee theorem, a sequence ``\\{(a_1, b_1), ...,(a_n, b_n)\\}`` (sorted in descending order of a) is graphic iff ``\\sum_{i = 1}^{n} a_i = \\sum_{i = 1}^{n} b_i\\}`` and the sequence obeys the property -
+```math
+\\sum_{i=1}^{r} a_i \\leq \\sum_{i=r+1}^n min(r-1,b_i) + \\sum_{i=r+1}^n min(r,b_i)
```
simonschoelly, line 772:
```suggestion
\\sum_{i=1}^{r} a_i \\leq \\sum_{i=1}^n min(r-1,b_i) + \\sum_{i=r+1}^n min(r,b_i)
```

src/connectivity.jl
simonschoelly, line 774:
```suggestion
for each integer 1 <= r <= n-1. 
```

src/connectivity.jl
simonschoelly: I leave the decision up too you. One thing that is nice about not throwing an error, that a user does not have to make an extra check for an error, if their input might contain a negative number.

But then we probably should also check, that all degrees are less or equal than `n - 1`?

Where is would still leave the throws check in, is in the `isdigraphical` method for inputs of different lengths.

Also note that by throwing an exception in `isgraphical` we kind of introduce a breaking change - but as this is a very rare edge case, I think that should be ok.

Let me know what you decide.

simonschoelly:
Some small fixes for the formula in the docstring and some decision to make, if we should throw errors - and then I think we can finally merge this.

src/connectivity.jl
InterdisciplinaryPhysicsTeam: Thanks for your argumentation. 

We agree that returning `false` allows for simpler usage, and if the user wishes to get an error in specific circumstances, she/he may wrap the function. Thus we now return `false` when negative degrees are present, or when a degree exceeds `n-1`.

InterdisciplinaryPhysicsTeam:
We committed your suggestions and accepted your proposal to return `false` above.

simonschoelly (approved):

simonschoelly:
Merged, thanks!

PR #189, opened 2022-11-18
Add Newman-Watts-Strogatz graph generator

blwh:
See docs.


src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -281,11 +289,12 @@ julia> watts_strogatz(Int8(10), 4, 0.8, is_directed=true, seed=123)
 
 ### References
 - Collective dynamics of ‘small-world’ networks, Duncan J. Watts, Steven H. Strogatz. [https://doi.org/10.1038/30918](https://doi.org/10.1038/30918)
+- Scaling and percolation in the small-world network model, M. E. J. Newman, Duncan J. Watts. [https://doi.org/10.1103/PhysRevE.60.7332](https://doi.org/10.1103/PhysRevE.60.7332)
 - Small Worlds, Duncan J. watts. [https://en.wikipedia.org/wiki/Special:BookSources?isbn=978-0691005416](https://en.wikipedia.org/wiki/Special:BookSources?isbn=978-0691005416)
 """
 function watts_strogatz(
     n::Integer, k::Integer, β::Real;
-    is_directed=false, rng::Union{Nothing, AbstractRNG}=nothing, seed::Union{Nothing, Integer}=nothing
+    newman_watts=false, is_directed=false, rng::Union{Nothing, AbstractRNG}=nothing, seed::Union{Nothing, Integer}=nothing
```
simonschoelly, line 297:
```suggestion
    newman_watts::Bool=false, is_directed=false, rng::Union{Nothing, AbstractRNG}=nothing, seed::Union{Nothing, Integer}=nothing
```
I would add a type here, although we also don't do it for other kwargs here. But that code is ancient.

simonschoelly:
Thanks. I am not (anymore) familiar with Watts-Strogatz, so I have not looked at the implementation yet. I also don't know, how well known, Newman-Watts-Strogatz but I was wondering, if it would be better for discoverability, if we would create a separate `newman_watts_strogatz` function, that calls `watts_strogatz` with the correct parameter set which in this case, we might call `remove_edges` instead.

blwh:
> Thanks. I am not (anymore) familiar with Watts-Strogatz, so I have not looked at the implementation yet. I also don't know, how well known, Newman-Watts-Strogatz but I was wondering, if it would be better for discoverability, if we would create a separate `newman_watts_strogatz` function, that calls `watts_strogatz` with the correct parameter set which in this case, we might call `remove_edges` instead.

This sounds like a good idea. I'll look into it this evening.

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -340,6 +341,57 @@ function watts_strogatz(
...
+be rewired randomly.
+
+Note: In the original paper by Newman and Watts, self-loops and double edges were
+allowed, which is not the case here. However, for large enough networks and small
+enough `p` and `k`, this should not deviate much from the original model.
+
+### Optional Arguments
+- `is_directed=false`: if true, return a directed graph.
+- `rng=nothing`: set the Random Number Generator.
+- `seed=nothing`: set the RNG seed.
```
simonschoelly, line 373:
```suggestion
```

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -340,6 +341,57 @@ function watts_strogatz(
...
+```
+
+### References
+- Scaling and percolation in the small-world network model, M. E. J. Newman, Duncan J. Watts. [https://doi.org/10.1103/PhysRevE.60.7332](https://doi.org/10.1103/PhysRevE.60.7332)
+"""
+function newman_watts_strogatz(
+    n::Integer, k::Integer, β::Real;
+    is_directed::Bool=false, rng::Union{Nothing, AbstractRNG}=nothing, seed::Union{Nothing, Integer}=nothing
+)
+    return watts_strogatz(n, k, β, is_directed=is_directed, remove_edges=false, rng=rng, seed=seed)
```
simonschoelly, line 391:
```suggestion
    is_directed::Bool=false, rng::Union{Nothing, AbstractRNG}=nothing
)
    return watts_strogatz(n, k, β, is_directed=is_directed, remove_edges=false, rng=rng)
```

simonschoelly: Currently our random functions allow both a `seed` and `rng` argument, because we wanted to transform from functions that take `seed` to functions that take `rng` (random functions in Julia Base also take an `rng`)  argument. In order to not break anything, we left the `seed` argument in for old functions, but for new ones I think we can remove it.

simonschoelly:
I think one thing that we still need, is a few tests for the new function.

blwh:
> I think one thing that we still need, is a few tests for the new function.

I'll have a look at this. Thank you for all the support thus far!

Issue #66, opened 2021-11-11
[Port] Add JuliaFormatter template


gdalle:
See https://github.com/sbromberger/LightGraphs.jl/issues/1592

gdalle:
See #190

PR #190, opened 2022-11-20
Uniform code style

aurorarossi:
It fixes some typos and spaces.


gdalle:
Hey @aurorarossi did you do this with JuliaFormatter or manually? Because it would be great to actually add a formatting check to the test suite with a consistent style (see #66). Gonna push this on your PR if that's OK

aurorarossi:
I have the following settings on vsc:
- enabled Insert Final Newline
- enabled Trim Final Newlines
- enabled Trim Trailing Whitespace
- disabled Ignore Trim Whitespace.

Yes of course! You can add the JuliaFormatter here

gdalle:
This is a great idea, thank you for starting it! I think the VSCode settings are less reliable than an explicit JuliaFormatter configuration (see [here](https://domluna.github.io/JuliaFormatter.jl/dev/config/)) with a well-known style like BlueStyle. Let's see how much of the code base it changes

.JuliaFormatter.toml
```diff
@@ -0,0 +1 @@
+style = "blue"
```
gdalle, line 1: BlueStyle seems like the most well-established specification, whereas SciML is more recent. However JuliaFormatter cannot do it all, so we must tell contributors to adhere to the style guidelines in general. Should we add a badge to the README too?

src/Graphs.jl
```diff
@@ -10,162 +10,409 @@ using Statistics: mean
 using Compat
 
 using Inflate: InflateGzipStream
-using DataStructures: IntDisjointSets, PriorityQueue, dequeue!, dequeue_pair!, enqueue!, heappop!, heappush!, in_same_set, peek, union!, find_root!
+using DataStructures:
```
gdalle, line 13: The huge diff is due to exports being line by line, following blue style specification

simonschoelly:
I think this PR also fixes so grammatical errors - now its a bit difficult to spot these parts with also these changes here - maybe we can make two PRs out of that?

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -855,7 +921,7 @@ Create a random directed [regular graph](https://en.wikipedia.org/wiki/Regular_g
 with `n` vertices, each with degree `k`.
 
 ### Optional Arguments
-- `dir=:out`: the direction of the edges for degree parameter.
+- `dir=:out`: the direction of edges for the degree parameter.
```
simonschoelly, line 924:
```suggestion
- `dir=:out`: the direction of the edges for the degree parameter.
```

gdalle (approved):

gdalle:
> I think this PR also fixes so grammatical errors - now its a bit difficult to spot these parts with also these changes here - maybe we can make two PRs out of that?

You're right, it's hard to spot but you can see them during code review if you filter by commit (to keep only the ones by @aurorarossi) and specify that you don't want to see whitespace modifications. I reviewed all of these changes and they look okay to me.

I just wanted to do code formatting in the same PR because if we merged @aurorarossi's PR alone it would have whitespace changes that are incoherent with the BlueStyle

.JuliaFormatter.toml
simonschoelly: Any reason to prefer `BlueStyle`? We could also specify our own rules right?

simonschoelly: In any cases, most of the changes here seem to be sensible.

src/Experimental/ShortestPaths/ShortestPaths.jl
```diff
@@ -99,12 +98,17 @@ s4 = shortest_paths(g, 1, BellmanFord())
 s5 = shortest_paths(g, 1, w, DEsopoPape())
 ```
 """
-shortest_paths(g::AbstractGraph, s, alg::ShortestPathAlgorithm) =
-    shortest_paths(g, s, weights(g), alg)
+function shortest_paths(g::AbstractGraph, s, alg::ShortestPathAlgorithm)
+    return shortest_paths(g, s, weights(g), alg)
+end
```
simonschoelly, line 103: This transformation is specified by blue - do we want that?

.JuliaFormatter.toml
gdalle: I just use it because it's already well documented, so we can point contributors to the exhaustive specification by Invenia. We could make up our own rules but that would require explaining them to everyone in our docs.

Also, it's easier to contribute when you're already familiar with the style guide, which wouldn't be the case for Graphs-specific rules.

gdalle: My hot take is that the precise choice of style does not matter, as long as it's clear and easy to apply, and BlueStyle satisfies both criteria

src/Experimental/ShortestPaths/ShortestPaths.jl
gdalle: I think so, yes. When the lines are too long, Blue replaces
```julia
f(x) = y
```
with
```julia
function f(x)
    return y
end
```

src/Experimental/vf2.jl
```diff
@@ -418,129 +451,221 @@ function vf2match!(state, depth, callback::Function, problemtype::GraphMorphismP
...
-        vf2(make_callback(c), g1, g2, IsomorphismProblem(), 
+function all_isomorph(
+    g1::AbstractGraph,
+    g2::AbstractGraph,
+    alg::VF2;
+    vertex_relation::Union{Nothing,Function}=nothing,
+    edge_relation::Union{Nothing,Function}=nothing,
+)::Channel{Vector{Tuple{eltype(g1),eltype(g2)}}}
+    T = Vector{Tuple{eltype(g1),eltype(g2)}}
+    !could_have_isomorph(g1, g2) && return Channel(_ -> return, ctype = T)  # TODO: fix problem with JuliaFormatter, right now the whole file is ignored
```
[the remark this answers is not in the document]
simonschoelly: I also just learning about channels when I wrote that code.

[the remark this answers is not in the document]
simonschoelly: I think we can also just surround the code in question with comments: https://domluna.github.io/JuliaFormatter.jl/stable/skipping_formatting/

.JuliaFormatter.toml
simonschoelly: Maybe it makes sense to use some common style, as then people don't need to reconfigure their editor differently for every project.

simonschoelly: What about the default format by JuliaFormatter? Would that also be an option?

simonschoelly (approved):

src/Experimental/vf2.jl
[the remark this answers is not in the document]
gdalle: I tried, let's see what happens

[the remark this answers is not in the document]
gdalle: Added the comments to skip formatting but JuliaFormatter still throws an error, so keeping the `ignore` kwarg for now

.JuliaFormatter.toml
gdalle: The default format by JuliaFormatter is much less well-specified than BlueStyle. One of the perks of BlueStyle is that it also gives coherent writing advice to the contributors that cannot be enfored by a simple linter

gdalle: > Maybe it makes sense to use some common style, as then people don't need to reconfigure their editor differently for every project.

Another perk of JuliaFormatter is that the formatting in the Julia extension for VSCode is based on it. And it takes into account the `.JuliaFormatter.toml` file at the root of the package to format each file according to the local specification (in our case `style = "blue"`)

Issue #193, opened 2022-11-22
Bug in randbn(n,p)


aurorarossi:
The function `randbn(n,p)` that returns a binomially-distributed random number with parameters $n$ and $p$. For every $n$, if $p=1$ it should return $n$ instead it loops.

aurorarossi:
Binomial.jl test is not working  (but it is not included in runtests.jl). I suggest to change that tests.

simonschoelly:
The random graphs in general is something we have to refactor at one point - form some graphs, I think we also kind of have an incorrect distribution.

In this PR I have added to the comments some code that I wrote a long time ago but never commited, that should take of the issue here. Maybe someone with a bit more time can take that old code and create a new PR out of it?

https://github.com/JuliaGraphs/Graphs.jl/pull/150

PR #195, opened 2022-11-22
Fix randbn(n,p)

aurorarossi:
This PR fixes the issue  https://github.com/JuliaGraphs/Graphs.jl/issues/193.
It fixes the function and adds some tests.

### DIFF
--- a/test/simplegraphs/generators/randgraphs.jl
+++ b/test/simplegraphs/generators/randgraphs.jl
@@ -364,4 +364,24 @@
         @test isvalid_simplegraph(rog3)
         @test !is_cyclic(rog3)
     end
+
+    @testset "randbn" begin
+        for i in 1:10
+            @test Graphs.SimpleGraphs.randbn(i, 0.0; rng=rng) == 0
+            @test Graphs.SimpleGraphs.randbn(i, 1.0; rng=rng) == i
+        end
+        N = 30
+        s1 = zeros(N)
+        s2 = zeros(N)
+        for i in 1:N
+            s1[i] = Graphs.SimpleGraphs.randbn(5, 0.3)
+            s2[i] = Graphs.SimpleGraphs.randbn(3, 0.7)
+        end
+        μ1 = mean(s1)
+        μ2 = mean(s2)
+        sv1 = sqrt((1 / N) * sum((s1 .- μ1) .^ 2))
+        sv2 = sqrt((1 / N) * sum((s2 .- μ2) .^ 2))
+        @test μ1 - sv1 <= 0.3 * 5 <= μ1 + sv1
+        @test μ2 - sv2 <= 0.7 * 3 <= μ2 + sv2
+    end
 end
--- a/test/simplegraphs/generators/randgraphs.jl
+++ b/test/simplegraphs/generators/randgraphs.jl
@@ -381,7 +381,7 @@
         μ2 = mean(s2)
         sv1 = sqrt((1 / N) * sum((s1 .- μ1) .^ 2))
         sv2 = sqrt((1 / N) * sum((s2 .- μ2) .^ 2))
-        @test μ1 - sv1 <= 0.3 * 5 <= μ1 + sv1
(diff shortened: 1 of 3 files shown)


test/simplegraphs/generators/randgraphs.jl
```diff
@@ -364,4 +364,24 @@
...
+        s1 = zeros(N)
+        s2 = zeros(N)
+        for i in 1:N
+            s1[i] = Graphs.SimpleGraphs.randbn(5, 0.3)
+            s2[i] = Graphs.SimpleGraphs.randbn(3, 0.7)
+        end
+        μ1 = mean(s1)
+        μ2 = mean(s2)
+        sv1 = sqrt((1 / N) * sum((s1 .- μ1) .^ 2))
+        sv2 = sqrt((1 / N) * sum((s2 .- μ2) .^ 2))
```
natema, line 383: To be consistent, it's better to calculate the stdev with `Statistics` as well: 
```suggestion
        μ1 = mean(s1)
        μ2 = mean(s2)
        sv1 = std(s1)
        sv2 = std(s2)
```
Note that `std` is using the unbiased estimator (dividing by $N-1$ instead of $N$), which is a bit more accurate.

test/simplegraphs/generators/randgraphs.jl
```diff
@@ -364,4 +364,24 @@
...
+        @test μ1 - sv1 <= 0.3 * 5 <= μ1 + sv1
```
natema, line 384: I would put a comment here for those that read this test and wonder if it should fail sometime (which is possible): 
```suggestion
        @test μ1 - sv1 <= 0.3 * 5 <= μ1 + sv1 # since the stdev of μ1 is around sv1/sqrt(N), this should rarely fail
```

natema (approved):

natema (approved):

test/simplegraphs/generators/randgraphs.jl
```diff
@@ -364,4 +364,24 @@
         @test isvalid_simplegraph(rog3)
         @test !is_cyclic(rog3)
     end
+
+    @testset "randbn" begin
+        for i in 1:10
```
simonschoelly, line 369:
```suggestion
        for i in 0:10
```
while we are it, we could also test that special case.

test/simplegraphs/generators/randgraphs.jl
```diff
@@ -364,4 +364,24 @@
...
+        for i in 1:N
+            s1[i] = Graphs.SimpleGraphs.randbn(5, 0.3)
+            s2[i] = Graphs.SimpleGraphs.randbn(3, 0.7)
+        end
+        μ1 = mean(s1)
+        μ2 = mean(s2)
+        sv1 = std(s1)
+        sv2 = std(s2)
+        @test μ1 - sv1 <= 0.3 * 5 <= μ1 + sv1 # since the stdev of μ1 is around sv1/sqrt(N), this should rarely fail
+        @test μ2 - sv2 <= 0.7 * 3 <= μ2 + sv2
```
simonschoelly, line 385: Interesting approach, so we are using a confidence interval to check if the mean of the distribution is within that interval?

I am not sure what kind of confidence level we get here and if it is really enough, but as the rng is probably seeded and the tests pass, this seems to be ok.

simonschoelly (approved):

test/simplegraphs/generators/randgraphs.jl
natema: I was also initially puzzled, that's why I suggested to add a comment. 
Now that I'm thinking about it again, multiplying all sides by $N$ we get that the bound is testing if the sum of all Bernoulli r.v.s in all experiments is within $\Theta(N)$, while a Chernouff bound should imply that they are within $O(\sqrt{N\log N})$ with high probability, so the probability is exponentially good. 
The way it is written is a bit obscure but I don't have a clearer alternative from the top of my head.

natema: @simonschoelly regarding this point, I thought that Aurora had copied the approach from elsewhere in the library. 
Instead, she told me that she introduced it herself.
I did find similar tests in the past, so we will try to reformulate it more clearly.

PR #197, opened 2022-11-23
Add regular tree generator

stecrotti:
Generator for a [regular tree](https://en.wikipedia.org/wiki/Bethe_lattice), also known as Bethe Lattice or Cayley Tree, which generalizes a binary tree to arbitrary degree. 

I tried to keep the code as close as possible to the one for `binary_tree` [here](https://github.com/JuliaGraphs/Graphs.jl/blob/ea6bcfeb71de9d1229bbd411ef9b10cac44fa795/src/SimpleGraphs/generators/staticgraphs.jl#L476-L512).


src/SimpleGraphs/generators/staticgraphs.jl
```diff
@@ -536,6 +536,68 @@ function double_binary_tree(k::Integer)
...
 
+"""
+    regular_tree(k::Integer, z::integer)
+
+Create a [k-regular tree](https://en.wikipedia.org/wiki/Bethe_lattice)
+of depth `k` and degree `z`.
+
+# Examples
+```jldoctest
+julia> binary_tree(4)
```
abraunst: Old doctest?

src/SimpleGraphs/generators/staticgraphs.jl
```diff
@@ -536,6 +536,54 @@ function double_binary_tree(k::Integer)
...
+
+julia> regular_tree(5, 2) == binary_tree(5)
+true
+```
+"""
+function regular_tree(k::T, z::T) where {T<:Integer}
+    k <= 0 && return SimpleGraph(0)
+    k == 1 && return SimpleGraph(1)
+    if Graphs.isbounded(k) && BigInt(z)^k - 1 > typemax(k)
+        throw(DomainError(k, "z^k - 1 not representable by type $T"))
```
aurorarossi, line 559: Here I would change the error from DomainError to InexactError.
I think the error is similar to:
```jl
julia> a::Int8=1923458675849328374548392
ERROR: InexactError: trunc(Int8, 1923458675849328374548392)
```
In our case `a` is `z^k`. What do you think?

src/SimpleGraphs/generators/staticgraphs.jl
```diff
@@ -536,6 +536,54 @@ function double_binary_tree(k::Integer)
...
+julia> regular_tree(4, 3)
+{40, 39} undirected simple Int64 graph
+
+julia> regular_tree(5, 2) == binary_tree(5)
+true
+```
+"""
+function regular_tree(k::T, z::T) where {T<:Integer}
+    k <= 0 && return SimpleGraph(0)
+    k == 1 && return SimpleGraph(1)
```
simonschoelly, line 557: In these cases we return a graph of of eltype `Int`, but in all other cases we return a graph of type `T`.

simonschoelly: To be completely honest, I am not sure if it is the best idea to determine the eltype of the graphs from `k` and `z` - we do that indeed for other graphs here, but I never really liked that.

simonschoelly: Maybe we can have a function signature
```julia
regular_tree(T::Type{<:Integer}, k::Integer, z::Integer)
```
and then define something like
```julia
regular_tree(k::Integer, z::Integer) = regular_tree(Int64, k, z)
```
that for be similar to the function `sprand` from `SparseArrays` for example.

src/SimpleGraphs/generators/staticgraphs.jl
```diff
@@ -536,6 +536,54 @@ function double_binary_tree(k::Integer)
     return g
 end
 
+"""
+    regular_tree(k::Integer, z::integer)
+
+Create a [regular tree](https://en.wikipedia.org/wiki/Bethe_lattice),
```
simonschoelly, line 542: That wikipedia article might be a bit confusing, because it talks about infinite trees in the introduction.

Maybe we can decribe that a bit more clearly here?

If I understood correctly, what you want to create is a perfect k-ary tree such as described here? https://en.wikipedia.org/wiki/M-ary_tree#Types_of_m-ary_trees

stecrotti:
By the way I called it 'regular tree' because of the analogy with regular graphs but that might not be too accurate because:
1. A regular tree is not a regular graph (because of the leaves)
2. z-regular graphs have nodes of degree z while z-ary trees have nodes of degree z+1, potentially confusing

That said, `mary_tree(k, m)`, `zary_tree(k,  z)` don't look great to me.

stecrotti:
Changes:
- New method with type `regular_tree(T::Type{<:Integer}, k::Integer, z::Integer)`
- Improved docs with correct wiki page
- `InexactError` instead of `DomainError` when number of vertices is not representable
- Assert z > 0, fallback to a path graph in case z = 1
- Appropriate conversions to the provided type `T`, checked by test

src/SimpleGraphs/generators/staticgraphs.jl
```diff
@@ -536,6 +536,60 @@ function double_binary_tree(k::Integer)
...
+    end
+    l = k
+    w = (z^(l - 1) - 1) ÷ (z - 1)
+    x = w + z^(l - 1)
+    @inbounds @simd for j in (w + 1):x
+        fadjlist[j] = T[ceil((j - x) / z) + w]
+    end
+    return SimpleGraph(ne, fadjlist)
+end
+
```
aurorarossi, line 590: I would add the docstring also to this function. 
````suggestion

"""
    regular_tree(k, z)

Create a regular tree or [perfect z-ary tree](https://en.wikipedia.org/wiki/M-ary_tree#Types_of_m-ary_trees): 
a `k`-level tree where all nodes except the leaves have exactly `z` children. 
For `z = 2` one recovers a binary tree.
# Examples
```jldoctest
julia> regular_tree(4, 3)
{40, 39} undirected simple Int64 graph
julia> regular_tree(5, 2) == binary_tree(5)
true
```
"""
````

**Review by aurorarossi — COMMENTED** [review]
Some comments on the docstrings. If you agree with the style then I will open a PR to fix `double_binary_tree` and `binary_tree`

File: src/SimpleGraphs/generators/staticgraphs.jl (hunk 5 above)
Comment (simonschoelly) [reply src/SimpleGraphs/generators/staticgraphs.jl:590]: It's quite often the case, that the docstrings for multiple methods of a function are only added for a single method - I think that would also be better here, than adding docstrings for both methods of this function.

**Comment by dapias** [comment]
I just realized that a $k$-ary tree is not the same as a Cayley tree of connectivity $k$ because in a $k$-ary tree, the root has connectivity $k$, and all the nodes excepting those in the leaves have connectivity $k+1$. Look, for instance, at the attached plot, nodes 2,3 and 4 have connectivity four and not three as it would be the case for a Cayley tree

Bildschirm­foto 2022-12-14 um 09 31 30

**Comment by stecrotti** [comment]
Agree! Indeed it was corrected and should be ok in the current version of this PR

**Comment by dapias** [comment]
Is there a function to create an actual Cayley tree?

File: src/SimpleGraphs/generators/staticgraphs.jl [hunk 6]
```diff
@@ -536,6 +536,60 @@ function double_binary_tree(k::Integer)
...
+julia> regular_tree(Int8, 3, 2)
+{7, 6} undirected simple Int8 graph
+
+julia> regular_tree(5, 2) == binary_tree(5)
+true
+```
+"""
+function regular_tree(T::Type{<:Integer}, k::Integer, z::Integer)
+    z <= 0 && throw(DomainError(z, "number of children must be positive"))
```
simonschoelly, line 559: Shouldn't we also check for `k >= 0` here?

src/SimpleGraphs/generators/staticgraphs.jl
```diff
@@ -536,6 +536,60 @@ function double_binary_tree(k::Integer)
...
+    z == 1 && return path_graph(T(k))
+    k <= 0 && return SimpleGraph(zero(T))
+    k == 1 && return SimpleGraph(one(T))
+    if Graphs.isbounded(k) && (BigInt(z)^k - 1) ÷ (z - 1) > typemax(T)
+        throw(InexactError(:convert, T, (BigInt(z)^k - 1) ÷ (z - 1)))
+    end
+
+    n = T((z^k - 1) / (z - 1))
```
simonschoelly, line 567: You already calculate the number of vertices with `(BigInt(z)^k - 1) ÷ (z - 1)`, so we we could just reuse this for `n`.

With the expression `n = T((z^k - 1) / (z - 1))` there are two issues:
- `z^k` might overflow before converting it to `T`
- By using the `/` operator instead of `÷` your intermediate result will be a double

simonschoelly: A lot of these overflow issues probably won't happen if we restrict  `k` and `z` to `Int` in the function definition.

stecrotti:
Changed the signature to match the rest of the file as suggested by @aurorarossi, but kept only one docstring.

Fixed potential overflow as suggested by @simonschoelly.

simonschoelly:
There is this one comment from me, about `k >= 0`  - not sure if you have seen it, but feel free to ignore that it you think we should not change anything there.

Otherwise looks good for me - I did not check the math, but choose to believe the tests.

stecrotti:
Sure, i replied under your review comment: 
> I stuck to what was used in the rest of the file which is that for k<=0 it returns a SimpleGraph(zero(T)). I guess there was a good reason for that? If not it makes total sense to error if k<0.

Krastanov:
@stecrotti , it seems this PR lost steam a few years ago. Could you run the formatter on it -- after that it should be good to go for merging.

Krastanov:
doctests are actually failing, so for now I will mark this as a draft to manage my review queue a bit more easily -- do not hesitate to convert it back to not draft if you decide to look into it

stecrotti:
I fixed the doctests.
Tests are failing on pre-releases which i understand is a known issue.
Coverage fails are pointing to lines of code consisting of only `end`, so i guess that's ok too.

Krastanov:
Thank you, Stefano! Your quick turnaround is extremely highly appreciated!

You might have noticed that we are struggling to have enough volunteers on the team. Let me know if you are interested in being added as a maintainer -- help with reviews can really help revive this ecosystem.

PR #198, opened 2022-11-23
Add newman_watts_strogatz graph generator

blwh:
In a later paper by Newman and Watts they generate networks that are slightly different from the original small-world networks by Watts and Strogatz. Instead of rewiring edges, they keep every original edge and add new long-range connections along the network.

This was already a pull request. However, I messed up and it got lost in the void (since I am still quite incompetent with git). Here is a new one, with tests and all. The tests use a loop to get an average degree (since it's a bit random). If this is a bad way to do this, please let me know and we'll sort something out.


simonschoelly:
Thanks!

PR #200, opened 2022-12-12
Add Bernoulli random graph

aurorarossi:
With this PR I add the Bernoulli random graph  and the $\rho$-correlated Bernoulli random graphs generators.

### DIFF
--- a/src/Graphs.jl
+++ b/src/Graphs.jl
@@ -295,6 +295,8 @@ export
     kronecker,
     dorogovtsev_mendes,
     random_orientation_dag,
+    bernoulli_graph,
+    rho_correlated_bernoulli_graphs,
 
     # community
     modularity,
--- a/src/SimpleGraphs/generators/randgraphs.jl
+++ b/src/SimpleGraphs/generators/randgraphs.jl
@@ -1426,7 +1426,7 @@ end
 """
     bernoulli_graph(Λ)
 
-Given the symmetric matrix ``\\Lambda \\in [0,1]^{n \\times n}``, return a Bernoulli graph with ``n`` vertices. Each edge ``(i,j)`` exists with  probability ``\\operatorname{Bernoulli}(\\Lambda[i,j])``.
+Given the symmetric matrix ``\\Lambda \\in [0,1]^{n \\times n}``, return a Bernoulli graph with ``n`` vertices. Each edge ``(i,j)`` exists with probability ``\\operatorname{Bernoulli}(\\Lambda[i,j])``.
 """
 function bernoulli_graph(
     Λ::Matrix{Float64};
--- a/src/SimpleGraphs/generators/randgraphs.jl
+++ b/src/SimpleGraphs/generators/randgraphs.jl
@@ -1435,10 +1435,11 @@ function bernoulli_graph(
 )
     @assert size(Λ)[1] == size(Λ)[2] "The probability matrix must be a square matrix!"
     n = size(Λ)[1]
+    rng = rng_from_rng_or_seed(rng, seed)
     A = SimpleGraph(n)
     for j in 1:n
         for i in (j + 1):n
-            if Bool(randbn(1, Λ[i, j]; rng=rng, seed=seed))
+            if Bool(randbn(1, Λ[i, j]; rng=rng))
                 add_edge!(A, i, j)
             end
         end
@@ -1452,21 +1453,22 @@ rho_correlated_bernoulli_graphs(Λ,ρ)
 Given the symmetric matrix ``\\Lambda \\in [0,1]^{n \\times n}`` and a real number ``\\rho \\in [0,1]`` return two ``\\rho``-correlated Bernoulli graphs with ``n`` vertices.
 """
 function rho_correlated_bernoulli_graphs(
-Λ::Matrix{Float64},
-ρ::Float64;
-rng::Union{Nothing,AbstractRNG}=nothing,
-seed::Union{Nothing,Integer}=nothing,
+    Λ::Matrix{Float64},
+    ρ::Float64;
+    rng::Union{Nothing,AbstractRNG}=nothing,
+    seed::Union{Nothing,Integer}=nothing,
 )
-n = size(Λ)[1]
-B = SimpleGraph(n)
-A = bernoulli_graph(Λ; rng=rng, seed=seed)
-A_adj = Int.(adjacency_matrix(A))
-for j in 1:n
-    for i in (j + 1):n
-        if Bool(randbn(1, (1 - ρ) * Λ[i, j] + ρ * A_adj[i, j]; rng=rng, seed=seed))
-            add_edge!(B, i, j)
+    n = size(Λ)[1]
+    rng = rng_from_rng_or_seed(rng, seed)
+    B = SimpleGraph(n)
+    A = bernoulli_graph(Λ; rng=rng, seed=seed)
+    A_adj = Int.(adjacency_matrix(A))
+    for j in 1:n
+        for i in (j + 1):n
+            if Bool(randbn(1, (1 - ρ) * Λ[i, j] + ρ * A_adj[i, j]; rng=rng))
+                add_edge!(B, i, j)
+            end
         end
     end
-end
-return A, B
+    return A, B
 end
--- a/src/SimpleGraphs/generators/randgraphs.jl
+++ b/src/SimpleGraphs/generators/randgraphs.jl
@@ -1422,3 +1422,25 @@ function random_orientation_dag(
     end
     return g2
 end
+
+"""
+    bernoulli_graph(Λ)
+
+Given the parametric symmetric matrix ``\\Lambda \\in [0,1]^{n \\times n}``, return a Bernoulli graph with ``n`` vertices. Each edge ``(i,j)`` exits with  probability ``\\operatorname{Bernoulli}(\\Lambda[i,j])``.
+"""
+function bernoulli_graph(
+    Λ::Matrix{Float64};
+    rng::Union{Nothing,AbstractRNG}=nothing,
+    seed::Union{Nothing,Integer}=nothing,
+)
+    n = size(Λ)[1]
+    A = SimpleGraph(n)
+    for j in 1:n
+        for i in (j + 1):n
+            if Bool(randbn(1, Λ[i, j]; rng=rng, seed=seed))
+                add_edge!(A, i, j)
+            end
+        end
+    end
+    return A
+end
--- a/src/SimpleGraphs/generators/randgraphs.jl
+++ b/src/SimpleGraphs/generators/randgraphs.jl
@@ -1444,3 +1444,28 @@ function bernoulli_graph(
     end
     return A
 end
+
+"""
+rho_correlated_bernoulli_graphs(Λ,ρ)
+
+Given the parametric symmetric matrix ``\\Lambda \\in [0,1]^{n \\times n}`` and a real number ``\\rho \\in [0,1]`` return two ``\\rho``-correlated Bernoulli graphs with ``n`` vertices.
+"""
+function rho_correlated_bernoulli_graphs(
+Λ::Matrix{Float64},
+ρ::Float64;
+rng::Union{Nothing,AbstractRNG}=nothing,
+seed::Union{Nothing,Integer}=nothing,
+)
+n = size(Λ)[1]
+B = SimpleGraph(n)
+A = bernoulli_graph(Λ; rng=rng, seed=seed)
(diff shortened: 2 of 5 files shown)


src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -1422,3 +1422,53 @@ function random_orientation_dag(
...
     return g2
 end
+
+"""
+    bernoulli_graph(Λ)
+
+Given the symmetric matrix ``\\Lambda \\in [0,1]^{n \\times n}``, return a Bernoulli graph with ``n`` vertices. Each edge ``(i,j)`` exists with probability ``\\operatorname{Bernoulli}(\\Lambda[i,j])``.
+"""
+function bernoulli_graph(
+    Λ::Matrix{Float64};
```
simonschoelly, line 1432:
```suggestion
    Λ::AbstractMatrix{<:AbstractFloat};
```
I would imagine that one wants to pass in a `SparseMatrixCSC` here.

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -1422,3 +1422,53 @@ function random_orientation_dag(
...
+    rng::Union{Nothing,AbstractRNG}=nothing,
+    seed::Union{Nothing,Integer}=nothing,
```
simonschoelly, line 1434: We have both `rng` and `seed` on older random functions so that we can switch from `seed` to `rng` without breaking anything. For newer functions I would only add `rng`.

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -1422,3 +1422,53 @@ function random_orientation_dag(
...
+)
+    @assert size(Λ)[1] == size(Λ)[2] "The probability matrix must be a square matrix!"
```
simonschoelly, line 1436: Using `@assert` in code is usually not a good idea, as the compiler might remove it. I would only use it in cases where it is absolutely clear, that the condition is always true, as a way of documentation.

From the Julia docs:
>  An assert might be disabled at various optimization levels. Assert should therefore only be used as a debugging tool and not used for authentication verification (e.g., verifying passwords), nor should side effects needed for the function to work correctly be used inside of asserts.

So we probably want to throw a proper exception here.

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -1422,3 +1422,53 @@ function random_orientation_dag(
...
+    n = size(Λ)[1]
+    rng = rng_from_rng_or_seed(rng, seed)
+    A = SimpleGraph(n)
+    for j in 1:n
+        for i in (j + 1):n
```
simonschoelly, line 1441: So no self-loops allowed? In that case we should document this, or add a flag that specifies if we allow self-loops.

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -1422,3 +1422,53 @@ function random_orientation_dag(
...
+            if Bool(randbn(1, Λ[i, j]; rng=rng))
```
simonschoelly, line 1442:
```suggestion
            if rand(rng) <= Λ[i, j]
```
much simpler and probably also more performant to use the build-in `rand` function.

simonschoelly: Also, we might skip the costly call to `rand` if `Λ[i, j]` is zero - that might be especially useful with sparse matrices.

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -1422,3 +1422,53 @@ function random_orientation_dag(
...
+    Λ::Matrix{Float64},
+    ρ::Float64;
+    rng::Union{Nothing,AbstractRNG}=nothing,
+    seed::Union{Nothing,Integer}=nothing,
+)
+    n = size(Λ)[1]
+    rng = rng_from_rng_or_seed(rng, seed)
+    B = SimpleGraph(n)
+    A = bernoulli_graph(Λ; rng=rng, seed=seed)
+    A_adj = Int.(adjacency_matrix(A))
```
simonschoelly, line 1465: This is a very costly operating of `A` is sparse. Can we somehow avoid it?

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -1422,3 +1422,53 @@ function random_orientation_dag(
...
+        end
+    end
+    return A
+end
+
+"""
+rho_correlated_bernoulli_graphs(Λ,ρ)
+
+Given the symmetric matrix ``\\Lambda \\in [0,1]^{n \\times n}`` and a real number ``\\rho \\in [0,1]`` return two ``\\rho``-correlated Bernoulli graphs with ``n`` vertices.
+"""
```
simonschoelly, line 1454: Can we specify a bit more, what it means to be rho-correlated for two graphs?

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -1422,3 +1422,53 @@ function random_orientation_dag(
...
+    seed::Union{Nothing,Integer}=nothing,
+)
+    n = size(Λ)[1]
+    rng = rng_from_rng_or_seed(rng, seed)
+    B = SimpleGraph(n)
+    A = bernoulli_graph(Λ; rng=rng, seed=seed)
+    A_adj = Int.(adjacency_matrix(A))
+    for j in 1:n
+        for i in (j + 1):n
+            if Bool(randbn(1, (1 - ρ) * Λ[i, j] + ρ * A_adj[i, j]; rng=rng))
```
simonschoelly, line 1468: `ρ` can also be negative, right? Does it still work in that case? Also, should we add some bound checks to ensure that `ρ` is in the range `[-1.0, 1.0]`?

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -1422,3 +1422,53 @@ function random_orientation_dag(
     end
     return g2
 end
+
+"""
+    bernoulli_graph(Λ)
+
+Given the symmetric matrix ``\\Lambda \\in [0,1]^{n \\times n}``, return a Bernoulli graph with ``n`` vertices. Each edge ``(i,j)`` exists with probability ``\\operatorname{Bernoulli}(\\Lambda[i,j])``.
```
simonschoelly, line 1429:
```suggestion
Given the symmetric matrix ``\\Lambda \\in [0,1]^{n \\times n}``, return a Bernoulli graph with ``n`` vertices. Each edge ``(i,j)`` exists with probability ``\\Lambda[i,j]``.
```
should be enough, right?

simonschoelly: Also, while I am not completely against the use of latex symbols in the docstrings, one should also keep in mind that these docstrings are often read in the REPL in a terminal, so one should use them sparingly.

src/SimpleGraphs/generators/randgraphs.jl
simonschoelly: Also, we ensure here, that the matrix is square, but not that it is symmetric. Maybe we should add a check for that, or we could also add to the documentation that we only check either the upper, or lower triangle of the matrix.

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -1422,3 +1422,53 @@ function random_orientation_dag(
...
+    A = bernoulli_graph(Λ; rng=rng, seed=seed)
+    A_adj = Int.(adjacency_matrix(A))
+    for j in 1:n
+        for i in (j + 1):n
+            if Bool(randbn(1, (1 - ρ) * Λ[i, j] + ρ * A_adj[i, j]; rng=rng))
+                add_edge!(B, i, j)
+            end
+        end
+    end
+    return A, B
```
simonschoelly, line 1473: We should probably document somewhere that what is returned is a tuple of two graphs

test/simplegraphs/generators/randgraphs.jl
```diff
@@ -384,4 +384,27 @@
         @test μ1 - sv1 <= 0.3 * 5 <= μ1 + sv1 # since the stdev of μ1 is around sv1/sqrt(N), this should rarely fail
         @test μ2 - sv2 <= 0.7 * 3 <= μ2 + sv2
     end
+
+    @testset "bernoulli graphs" begin
+        n = 50
+        Λ = rand(n, n)
+        ρ = 1.0 # isomorphism case
```
simonschoelly, line 391: If I understand correctly, it is a bit more than ismorphism here, as the graphs are actually equal without shuffling the vertices.

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -1422,3 +1422,53 @@ function random_orientation_dag(
     end
     return g2
 end
+
+"""
+    bernoulli_graph(Λ)
```
simonschoelly, line 1427: We need to document the `rng` kwarg somewhere.

simonschoelly:
Great! have you also thought about some functions for the directed case?

src/SimpleGraphs/generators/randgraphs.jl
aurorarossi: yes, it is unnecessary

src/SimpleGraphs/generators/randgraphs.jl
aurorarossi: In the definition I found $\rho$ is in $[0,1]$. I will add the check for it.

aurorarossi:
> Great! have you also thought about some functions for the directed case?

I have just implemented the generator described in [this paper](https://arxiv.org/pdf/1405.3133.pdf) , I have no knowledge of a directed case.

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -1422,3 +1422,54 @@ function random_orientation_dag(
     end
     return g2
 end
+
+"""
+    bernoulli_graph(Λ; rng = nothing, nodes_type = Int64)
+
+Given the symmetric matrix ``\\Lambda \\in [0,1]^{n \\times n}``, return a Bernoulli graph with ``n`` vertices. Each edge ``(i,j)`` exists with probability ``\\Lambda[i,j]``.
```
gdalle, line 1429: I would recommend using unicode for the docstrings, so that they are pleasant to read in the REPL too

src/Graphs.jl
```diff
@@ -295,6 +295,8 @@ export
     kronecker,
     dorogovtsev_mendes,
     random_orientation_dag,
+    bernoulli_graph,
+    rho_correlated_bernoulli_graphs,
```
gdalle, line 299: maybe the mention of `rho` is not necessary

test/runtests.jl
```diff
@@ -21,7 +21,7 @@ const testdir = dirname(@__FILE__)
 end
 
 @testset verbose = true "Code formatting (JuliaFormatter.jl)" begin
-    @test format(Graphs; verbose=false, overwrite=false, ignore=["vf2.jl"])  # TODO: remove ignore kwarg once the file is formatted correctly
+    @test format(Graphs; verbose=false, overwrite=false, ignore="vf2.jl")  # TODO: remove ignore kwarg once the file is formatted correctly
```
gdalle, line 24: this will be solved by merging `master`

test/simplegraphs/generators/randgraphs.jl
```diff
@@ -384,4 +384,33 @@
...
+        @test_throws ArgumentError("ρ must be in [0,1]") rho_correlated_bernoulli_graphs(
+            Λ, -1.0; rng
+        )
+        ρ = 1.0 # isomorphism case
+        g1, g2 = rho_correlated_bernoulli_graphs(Λ, ρ; rng)
+        g1_adj = adjacency_matrix(g1)
+        g2_adj = adjacency_matrix(g2)
+        @test g1_adj == g2_adj
+        @test diag(g1_adj) == diag(g2_adj) == zeros(n)
+        ρ = 0.5 # non isomorphism case
```
gdalle, line 405: can we do a probabilistic test on the correlation with ample error margins?

gdalle (changes requested):

gdalle:
@aurorarossi is this good to go?

gdalle:
Side note, this might be linked to #212

gdalle (changes requested):
This looks good but it still needs tests that the probability of an edge is approximately `p`, and that the correlation between (nondiagonal?) adjacency matrices is approximately `\rho`

PR #205, opened 2022-12-30
fix benchmarks

miguelraz:
a previous code cleanup massive PR seems to have introduced this macro change that broke benchmarking.
this should work now.


gdalle:
Thanks a bunch @miguelraz! I guess you can't make a clean up omelet without breaking some benchmark eggs :shrug:

gdalle (approved):

miguelraz:
Oh neat! No problem.

I needed to benchmark some stuff for the eventual lil' blogpost I wanted to write about Boolean graphs I wanted to write and I got yak-shaved into fixing the benchmarks.

Keep poking me until I write that package...

PR #206, opened 2022-12-31
Prufer coding for trees

SimonCoste:
Hi, 
I added elementary functionalities related to trees. There are 3 exported functions, 
- `is_tree` checks if a simple graph is a tree
- `prufer_encode` creates the Prüfer sequence associated to a tree
- `prufer_decode` creates the tree associated with a Prüfer sequence

This is my first PR; I did my best at following the style guide and provide a doc, please do not hesitate to explain how things should be done. 

I also modified the `docs` but I did not succeed in having it correctly deployed locally; I get a new `trees` pages but none of the docstrings I wrote seem to be parsed by the Documenter. I also get the following possibly related warning:
```
Warning: Documenter could not auto-detect the building environment Skipping deployment.
└ @ Documenter ~/.julia/packages/Documenter/bFHi4/src/deployconfig.jl:75
```

Related: `https://github.com/JuliaGraphs/Graphs.jl/pull/194`. I'll add a random Cayley tree generator very soon.


src/trees/prufer.jl
```diff
@@ -0,0 +1,91 @@
...
+
+Returns true if g is a tree: that is, a simple, connected undirected graph, with nv-1 edges (nv = number of vertices). Trees are the minimal connected graphs; equivalently they have no cycles.  
+
+This function does not apply to directed graphs. Directed trees are sometimes called [polytrees](https://en.wikipedia.org/wiki/Polytree)). 
+
+"""
+
+function is_tree(g::AbstractGraph)
+    is_directed(g) && throw(ArgumentError("g must be undirected. "))
+    return is_connected(g) && ne(g) == nv(g) - 1
```
simonschoelly, line 12:
```suggestion
    return nv(g) + 1 == ne(g)  && is_connected(g)
```
`is_connected` is much more expensive than calculating `nv(g) + 1 == ne(g)`  so I would do the second calculation first, if that one is false, then we don't need to check for connectivity.

src/trees/prufer.jl
```diff
@@ -0,0 +1,91 @@
...
+    is_tree(g)
+
+Returns true if g is a tree: that is, a simple, connected undirected graph, with nv-1 edges (nv = number of vertices). Trees are the minimal connected graphs; equivalently they have no cycles.  
+
+This function does not apply to directed graphs. Directed trees are sometimes called [polytrees](https://en.wikipedia.org/wiki/Polytree)). 
+
+"""
+
+function is_tree(g::AbstractGraph)
+    is_directed(g) && throw(ArgumentError("g must be undirected. "))
```
simonschoelly, line 11: We currently use the `@traitfn` macro to restrict functions to directed/undirected graphs. An example can be found here: https://github.com/JuliaGraphs/Graphs.jl/blob/ea6bcfeb71de9d1229bbd411ef9b10cac44fa795/src/traversals/dfs.jl#L14

src/trees/prufer.jl
```diff
@@ -0,0 +1,91 @@
...
+
+"""
+    prufer_decode(code)
+
+Returns the unique tree associated with the given (Prüfer) code. 
+Each tree of size n is associated with a Prüfer sequence (a[1], ..., a[n-2]) with 1 ⩽ a[i] ⩽ n. The sequence is constructed recursively by the leaf removal algoritm. At step k, the leaf with smallest index is removed and its unique neighbor is added to the Prüfer sequence, giving a[k]. The decoding algorithm goes backward.  
+Ref: [Prüfer sequence on Wikipedia](https://en.wikipedia.org/wiki/Pr%C3%BCfer_sequence)
+"""
+
+function prufer_decode(code::Array{T,1})::Graph{T} where {T<:Integer}
```
simonschoelly, line 36:
```suggestion
function prufer_decode(code::AbstractVector{T})::SimpleGraph{T} where {T<:Integer}
```
Probably more clear what is happening, if we use `SimpleGraph` here, `Graph` is just an alias for that. It is also not that usual to annotate return types in Julia - if I remember correctly it will call `convert(SimpleGraph{T}, g)` on the result `g` of that function - this will probably just be NOP instruction here, so I don't mind.

simonschoelly: `Vector{T}` is an alias for `Array{T, 1}` and `AbstractVector{T}` is an alias for `AbstractArray{T, 1}` - so in order to allow other types of vectors, we can maybe use `AbstractVector{T}` here.

simonschoelly: I would also debate, if the return type should be inferred from the type of the array, or if we should allow one to give an optional argument specifying the resulting graph eltype. See here for some for something similar: https://github.com/JuliaGraphs/Graphs.jl/pull/197

simonschoelly: Lastly, maybe we should move functions that only work with `SimpleGraph` and `SimpleDiGraph` to the `SimpleGraphs` submodule?

simonschoelly:
Hi, I already added a few comments so far - I will add more later, but I need to figure out how these Prüfer sequences work first.

src/trees/prufer.jl
```diff
@@ -0,0 +1,91 @@
...
+
+    return g
+end
+
+"""
+    prufer_encode(g::SimpleGraph)
+
+Given a tree (a connected minimal undirected graph) of size n⩾3, returns the unique Prüfer sequence associated with this tree. 
+
+Each tree of size n ⩾ 3 is associated with a Prüfer sequence (a[1], ..., a[n-2]) with 1 ⩽ a[i] ⩽ n. The sequence is constructed recursively by the leaf removal algoritm. At step k, the leaf with smallest index is removed and its unique neighbor is added to the Prüfer sequence, giving a[k].
```
simonschoelly, line 65: Shouldn't this also work for trees with two vertices? In such a case we would just get the empty sequence.

src/trees/prufer.jl
```diff
@@ -0,0 +1,91 @@
...
+        throw(ArgumentError("The graph must be a tree with n ⩾ 3 vertices. "))
+    g = copy(G)
+    code = zeros(T, n - 2)
+    d = degree(g)
+    L = BinaryMinHeap(findall(==(1), d))
+
+    for i in 1:(n - 2)
+        l = pop!(L)
+        v = neighbors(g, l)[1]
+        rem_edge!(g, l, v)
```
simonschoelly, line 83: `rem_edge!` might not work for all types of graphs. So either we should restrict this function to `SimpleGraph`, convert the graph to a `SimpleGraph` when we create a copy of it or implement this algorithm in a way that does need to modify a graph.

src/trees/prufer.jl
SimonCoste: It would, I restricted it to n >2 because I think empty sequences feel a little bit off here, but I can add it.

src/trees/prufer.jl
SimonCoste: Regarding return types: since there is an `encode` and `decode` function, I wanted to impose the return type directly from the input, in order to guarantee `decode(encode(g))==g`. If the return type was an argument this would not be granted directly.

Maybe you suggest to have two methods `encode(T::Type,g::SimpleGraph)` and `encode(g::SimpleGraph)`, and in the second method the type is directly inferred from `g` as in my function ? I'm not sure this would be really useful...

src/trees/prufer.jl
SimonCoste: What are the types of graphs for which it does not work ?

SimonCoste: Done.

src/trees/prufer.jl
```diff
@@ -0,0 +1,96 @@
...
+
+Ref: [Prüfer sequence on Wikipedia](https://en.wikipedia.org/wiki/Pr%C3%BCfer_sequence)
+
+"""
+
+function prufer_encode(G::SimpleGraph{T})::AbstractVector{T} where {T<:Integer}
+    !is_tree(G) && throw(ArgumentError("The graph must be a tree with n ⩾ 2 vertices. "))
+
+    n = nv(G)
+    n == 0 && return Vector{T}() # empty Prüfer sequence
```
simonschoelly, line 79:
```suggestion
    (nv(G) < 2 || !is_tree(G)) && throw(ArgumentError("The graph must be a tree with n ⩾ 2 vertices. "))

    n = nv(G)
    n == 2 && return Vector{T}() # empty Prüfer sequence 
```

src/trees/prufer.jl
```diff
@@ -0,0 +1,96 @@
...
+
+Given a tree (a connected minimal undirected graph) of size n⩾3, returns the unique Prüfer sequence associated with this tree. 
+
+Each tree of size n ⩾ 2 is associated with a Prüfer sequence (a[1], ..., a[n-2]) with 1 ⩽ a[i] ⩽ n. The sequence is constructed recursively by the leaf removal algoritm. At step k, the leaf with smallest index is removed and its unique neighbor is added to the Prüfer sequence, giving a[k]. The Prüfer sequence of the tree with only one edge is the empty sequence. 
+
+Ref: [Prüfer sequence on Wikipedia](https://en.wikipedia.org/wiki/Pr%C3%BCfer_sequence)
+
+"""
+
+function prufer_encode(G::SimpleGraph{T})::AbstractVector{T} where {T<:Integer}
```
simonschoelly, line 75:
```suggestion
function prufer_encode(G::SimpleGraph{T})::Vector{T} where {T<:Integer}
```
There is no advantage in weaking the return type postcondition.

simonschoelly: Personally I would also avoid upper case variable names for graphs, as these are more used for type and type parameters, but we currently do not state that or enforce that anywhere so that is your decision.

src/trees/prufer.jl
```diff
@@ -0,0 +1,96 @@
...
+    g = copy(G)
+    code = zeros(T, n - 2)
+    d = degree(g)
+    L = BinaryMinHeap(findall(==(1), d))
+
+    for i in 1:(n - 2)
+        l = pop!(L)
+        v = neighbors(g, l)[1]
+        rem_edge!(g, l, v)
+        d[l] -= 1
```
simonschoelly, line 89:
```suggestion
        u = pop!(L)
        v = neighbors(g, u)[1]
        rem_edge!(g, u, v)
        d[u] -= 1
```
Not a big thing, but it is kind of a convention to use the letters `u`, `v`, `w` for vertices.

src/trees/prufer.jl
```diff
@@ -0,0 +1,96 @@
+"""
+    is_tree(g)
+
+Returns true if g is a tree: that is, a simple, connected undirected graph, with nv-1 edges (nv = number of vertices). Trees are the minimal connected graphs; equivalently they have no cycles.  
+
+This function does not apply to directed graphs. Directed trees are sometimes called [polytrees](https://en.wikipedia.org/wiki/Polytree)). 
+
+"""
+
+function is_tree end
```
simonschoelly, line 10:
```suggestion
"""
function is_tree end
```
as far as I know, we can't have an empty line here - correct me if I am not right.

src/trees/prufer.jl
```diff
@@ -0,0 +1,96 @@
...
+           (isempty(c) || maximum(c) <= length(c) + 2)
+end
+
+function _degree_from_prufer(c::Vector{T})::Vector{T} where {T<:Integer}
+    """
+    Degree sequence from Prüfer code.
+    Returns d such that d[i] = 1 + number of occurences of i in c
+    """
+    n = length(c) + 2
+    return [T(count(==(i), c) + 1) for i in 1:n]
```
simonschoelly, line 28: I think this might have quadratic complexity, as we over the vector `c`  for each `i` - I think we can do that in linear time.

src/trees/prufer.jl
```diff
@@ -0,0 +1,96 @@
...
+function prufer_decode(code::AbstractVector{T})::SimpleGraph{T} where {T<:Integer}
+    !_is_prufer(code) && throw(
+        ArgumentError("The code must have one dimension and must be a Prüfer sequence. "),
+    )
+    isempty(code) && return path_graph(T(2)) # the empty Prüfer sequence codes for the one-edge tree
+
+    n = length(code) + 2
+    d = _degree_from_prufer(code)
+    L = BinaryMinHeap{T}(findall(==(1), d))
+    g = Graph{T}(n, 0)
```
simonschoelly, line 49:
```suggestion
    g = Graph{T}(n)
```
`Graph{T}(n, 0)` will create a random graph with `n` vertices and `0` edges - of course that is the same, but there might be some costly computation such as initializing a random number generator.

src/trees/prufer.jl
SimonCoste: I really don't know!

src/trees/prufer.jl
```diff
@@ -0,0 +1,96 @@
...
+
+@traitfn function is_tree(g::::(!IsDirected))
+    return ne(g) == nv(g) - 1 && is_connected(g)
+end
+
+function _is_prufer(c)
+    return ndims(c) == 1 &&
+           eltype(c) <: Integer &&
+           (isempty(c) || maximum(c) <= length(c) + 2)
+end
```
simonschoelly, line 20: I think we don't need the check for the numbers of dimensions and eltype anymore, as these are enforced by the type, right? On the we should probably check if the minimum is  >= 1.

src/trees/prufer.jl
SimonCoste: You're right, I'll simply loop over the Prüfer sequence.

src/trees/prufer.jl
SimonCoste: Totally. Will do.

src/trees/prufer.jl
```diff
@@ -0,0 +1,96 @@
...
+
+function prufer_decode(code::AbstractVector{T})::SimpleGraph{T} where {T<:Integer}
+    !_is_prufer(code) && throw(
+        ArgumentError("The code must have one dimension and must be a Prüfer sequence. "),
+    )
+    isempty(code) && return path_graph(T(2)) # the empty Prüfer sequence codes for the one-edge tree
+
+    n = length(code) + 2
+    d = _degree_from_prufer(code)
+    L = BinaryMinHeap{T}(findall(==(1), d))
```
simonschoelly, line 48:
```suggestion
    L = BinaryMaxHeap{T}(findall(==(1), d))
```
This should be a max-heap, right?

Otherwise I think this will not work for the path graph `2 -- 4 -- 3 -- 1` as this graph has the Prüfer sequence `[4, 3]` and that will then result in the path graph `1 -- 4 -- 3 -- 2`.

test/trees/prufer.jl
```diff
@@ -0,0 +1,61 @@
...
+        for t in testgraphs(t1, t2, t3)
+            @test is_tree(t)
+        end
+        for g in testgraphs(g1, g2, g3)
+            @test !is_tree(g)
+        end
+        for g in testgraphs(d1, d2, d3)
+            @test_throws MethodError is_tree(g)
+        end
+    end
```
simonschoelly, line 36: Can we also have the following special test cases here:
- An empty graph
- A graph with a single vertex
- Two single vertices
- A graph with a self-loop
- And maybe a graph that is a forest of two trees

I made the experience, that our algorithms often error on such test cases.

simonschoelly:
A few more minor things (and one major) - appart from that I haven't worked with the documentation for a long time, so I have not found out what is wrong here, maybe @gdalle knows more about that, otherwise I will try too look at it in the next fews days.

src/trees/prufer.jl
simonschoelly: Apparently it says here, that blanc lines are not allowed: https://docs.julialang.org/en/v1/manual/documentation/#Writing-Documentation

And yes, this might influence the documentation that is created by Documenter.jl (and also the documentation in the REPL when you type `?is_tree`).

src/trees/prufer.jl
SimonCoste: Prüfer encoding relies on the choice of a leaf selection rule at each step. Some papers use the minimum leaf rule, others use the maximum leaf rule. Different selection rules with result in different Prüfer encodings.

In the docstrings I chose the minimum leaf rule (I think it's the conventional choice), so at each step I remove the leaf with smallest index, hence my use of `BinaryMinHeap` (the `pop!` method removes the root, ie the smallest element).

SimonCoste: That being said, maybe we could add the selection rule as a parameter.

simonschoelly: A right, my fault, I calculated the Prüfer sequence incorrect, when trying to understand the algorithm. Then we probably won't need the import of `BinaryMaxHeap`?

SimonCoste: Haha, at the beginning I wasn't sure if I was going to use the min-rule or max-rule, so I imported both... what do you think about having an argument `selection_rule` that could be `"min"` (default) or `"max"` ?

SimonCoste:
Hi, I adapted to all your remarks. 

Importantly, I also added the `uniform_tree(n)` function. It generates a uniform tree over the $n^{n-2}$ such trees by decoding a random uniform Prüfer sequence.

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -967,6 +967,33 @@ function random_configuration_model(
...
+function uniform_tree(
+    n::Integer;
+    rng::Union{Nothing,AbstractRNG}=nothing,
+    seed::Union{Nothing,Integer}=nothing,
+)
+    n == 1 && return Graph(1, 0)
+    n == 2 && return path_graph(2)
+
+    rng = rng_from_rng_or_seed(rng, seed)
+    random_code = rand(rng, 1:n, n - 2)
```
simonschoelly, line 994:
```suggestion
    n <= 1 && return Graph(n)
    n == 2 && return path_graph(n)

    rng = rng_from_rng_or_seed(rng, seed)
    random_code = rand(rng, Base.OneTo(n), n - 2)
```
This way, we can determine the eltype of the graph form the type of `n` - alternatively we could also add an optional argument that determines the eltype - but we rarely do this at least for now.

simonschoelly: Not sure, if we should also accept 0 vertex graphs here, or if we should rather fail in that case - probably the later, as `is_tree` would not be true in that case.

simonschoelly: about `1:n` the problem with that is, that
```julia
julia> typeof(1:UInt32(4))
UnitRange{Int64}

julia> typeof(1:UInt64(4))
UnitRange{UInt64}
```
so this is probably not what we want - so its better to either cast `1` to the correct type, or use `Base.OneTo`.

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -967,6 +967,33 @@ function random_configuration_model(
...
+# Examples
+```jldoctest
+julia> uniform_tree(10)
+{10, 9} undirected simple Int64 graph
+```
+"""
+function uniform_tree(
+    n::Integer;
+    rng::Union{Nothing,AbstractRNG}=nothing,
+    seed::Union{Nothing,Integer}=nothing,
```
simonschoelly, line 988: I would remove the `seed` kwarg here - the other random generators just have this for legacy reasons, as we wanted to switch from  `seed` to `rng` but did not want to break existing code. Using `rng` instead of `seed` is consistent with other functions from the Julia standard library.

src/SimpleGraphs/generators/randgraphs.jl
simonschoelly: Without the change proposed here, this function would also not be type stable if `n` is a `UInt64` for example. One can check this by calling using the macro `@code_warntype`:
```julia
@code_warntype uniform_tree(UInt64(2))
```
should some type instabilities.

For testing there is also the `@inferred` macro that helps for checking type stability: https://docs.julialang.org/en/v1/stdlib/Test/#Test.@inferred

src/SimpleGraphs/generators/randgraphs.jl
simonschoelly: That is ususally a good strategy, just that  some of the code here is rather old, but we can't always fix it.

src/trees/prufer.jl
simonschoelly: I am not sure how much such a functionality is used - but what you could also do is to have an optional  `ordering` parameter that takes a `Base.Order.Ordering` object as described [here](https://docs.julialang.org/en/v1/stdlib/Test/#Test.@inferred). 

As far as I know, the `BinaryHeap` class from `DataStructures.jl` should work with that.

src/SimpleGraphs/generators/randgraphs.jl
SimonCoste: I removed `seed` from the arg and docstring. Is `rng = rng = rng_from_rng_or_seed(rng)` still necessary ?

src/trees/prufer.jl
SimonCoste: Ok. I really think that's unnecessary for the moment and I'll leave this for later.

src/SimpleGraphs/generators/randgraphs.jl
simonschoelly: I would still use it, because that function then returns a default random generator, if that argument is omitted.

SimonCoste:
I fixed a small bug (must specify `seed=nothing` in `rng_from_rng_or_seed` arguments). Everything should be fine now !

simonschoelly:
The documentation is also working now, right? Then it does indeed seem ready to be merged.

SimonCoste:
Yes, the doc is properly working!

simonschoelly (approved):

simonschoelly:
Great, thank you - for your next PR, the tests should also run automatically, its just for someones first PR, that we have to to trigger the tests automatically

SimonCoste:
In the near future (~2 months) I'll add more random tree generators, possibly: 
- random binary trees
- random recursive trees 
- random uniform trees with a fixed degree sequence
- random uniform trees with a fixed number of edges. 

Thanks Simon for the support !

simonschoelly:
> In the near future (~2 months) I'll add more random tree generators, possibly: 
> 
>     * random binary trees
> 
>     * random recursive trees
> 
>     * random uniform trees with a fixed degree sequence
> 
>     * random uniform trees with a fixed number of edges.
> 
> 
> Thanks Simon for the support !

That are definitely welcome additions. When taking inspiration from existing code for random graphs here, you just have to be a bit careful, that code sometimes quite old and buggy, and might not even calculate the expected distribution (If I remember correctly, for `random_regular_ graph`).

SimonCoste:
What is this potential bias in 'random_regular_graph' ?

simonschoelly:
I don't know where exactly the error is, but lets look at an example, the 2-regular graphs on 6-vertices, assuming that the vertices are labeled:

There 60 connected graphs (a single cycle) given by `5!/2 = 60`.
There are 10 disconnected graphs (two cycles of 3 vertices each) given by `{5 choose 2} = 10`.

Therefore the probability of getting a connected graph should be `60 / (10 + 60) = 6/7 ≈ 0.857`.

We can also estimate this probability by sampling repeatedly with `random_regular_graph`:

```julia
julia> using Graphs

julia> p_connected(n) = count(is_connected(random_regular_graph(6, 2)) for _ in 1:n) / n
p_connected (generic function with 1 method)

julia> p_connected(10^6)
0.6911
```
so there is something off with that algorithm And it can also not be for the (probably much more difficult) problem of random regular graphs on unlabeled vertices, because then the probability would be `1/2`.

PR #150, opened 2022-06-24
faster implementation of erdos_renyi

etiennedeg:
Master : 
``` julia
julia> @benchmark erdos_renyi(500, 0.5)
BenchmarkTools.Trial: 115 samples with 1 evaluation.
 Range (min … max):  31.674 ms … 102.297 ms  ┊ GC (min … max): 0.00% … 14.32%
 Time  (median):     39.373 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   43.619 ms ±  13.058 ms  ┊ GC (mean ± σ):  1.75% ±  5.16%

     ▇▂█▂▆▁▂ ▄ 
  ▄█████████▃█▆▆█▁▆▄▃▆▃▁▄▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▃▁▁▃▁▁▁▁▁▁▄▁▁▁▄▁▃▁▁▁▁▃ ▃
  31.7 ms         Histogram: frequency by time         90.9 ms <

 Memory estimate: 3.69 MiB, allocs estimate: 2502.

julia> @benchmark erdos_renyi(5000, 0.5)
BenchmarkTools.Trial: 1 sample with 1 evaluation.
 Single result which took 17.436 s (1.51% GC) to evaluate,
 with a memory estimate of 263.33 MiB, over 35003 allocations.

julia> @benchmark erdos_renyi(500, is_directed = true, 0.5)
BenchmarkTools.Trial: 82 samples with 1 evaluation.
 Range (min … max):  58.535 ms … 67.722 ms  ┊ GC (min … max): 0.00% … 5.64%
 Time  (median):     61.109 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   61.466 ms ±  1.972 ms  ┊ GC (mean ± σ):  0.98% ± 2.18%

         █▂▂  ▅ ▅  █▂▂    ▂      ▂
  ████▁▁███████▅█▅▁████▁██████▅▅▅█▅▁▅▁█▁▁▅█▅▅▅▁▅▅▁▁▁▁▁▅▁▁▁▁▁▅ ▁
  58.5 ms         Histogram: frequency by time        66.9 ms <

 Memory estimate: 7.38 MiB, allocs estimate: 5003.

julia> @benchmark erdos_renyi(5000, is_directed = true, 0.5)
BenchmarkTools.Trial: 1 sample with 1 evaluation.
 Single result which took 37.086 s (0.62% GC) to evaluate,
[...]

PR #212, opened 2023-01-07
faster erdos_renyi

abraunst:
See https://github.com/JuliaGraphs/Graphs.jl/pull/150#issuecomment-1374486801


abraunst:
It's a miracle that checks have passed because some of them depend on the specific random instance (that surely changed)...

abraunst:
Is this what you proposed in https://github.com/JuliaGraphs/Graphs.jl/pull/150#issuecomment-1165968212 for n,p graphs, @simonschoelly ?

abraunst:
Can I do something to make this more easily reviewable? The code essentially 
* (undirected case) samples a  $p$-sparse binary vector of size $\binom{N}{2}$ and then maps it into the upper triangle of a $N\times N$ matrix
* (directed case) samples a $p$-sparse binary vector of size $N*(N-1)$ and then maps it into the non-diagonal elements of a $N\times N$ matrix 

and then constructs the corresponding graph. The sampling of the binary vector is done by `randsubseq` which is by the way the function used for the same purpose by the `SparseArrays` standard library.

gdalle:
Sorry for the late review. @etiennedeg should we keep this one or #150?

etiennedeg:
More likely this one. I can try to review it this week (probably not before Wednesday, but I should be able to find some time)

abraunst:
> Sorry for the latency, and thanks for the contribution. @etiennedeg should we keep this one or #150?
> 
> @abraunst what kind of tests can guarantee correctness despite the randomness?

Maybe something like checking for average degree with a reasonable tolerance?

etiennedeg:
When I benchmark against the implementation proposed by @simonschoelly in #150, I get this (only for undirected graph because simon's code seems broken for directed graphs):
``` julia
julia> for N in (500, 5000), is_directed in (false, ), p in (10/N, 0.5, 1-10/N)
           @btime erdos_renyi_ss($N, $p; is_directed=$is_directed)
       end
  164.086 μs (504 allocations: 77.45 KiB)
  2.410 ms (504 allocations: 1.02 MiB)
  2.538 ms (504 allocations: 1.95 MiB)
  1.927 ms (5007 allocations: 792.03 KiB)
  351.821 ms (10007 allocations: 95.79 MiB)
  435.251 ms (10007 allocations: 190.78 MiB)

julia> for N in (500, 5000), is_directed in (false, ), p in (10/N, 0.5, 1-10/N)
           @btime erdos_renyi_pr($N, $p; is_directed=$is_directed)
       end
  450.744 μs (1296 allocations: 191.47 KiB)
  10.534 ms (3017 allocations: 4.76 MiB)
  17.142 ms (3017 allocations: 5.69 MiB)
  5.269 ms (13225 allocations: 2.03 MiB)
  1.107 s (42520 allocations: 361.60 MiB)
  2.319 s (47520 allocations: 788.97 MiB)
```
I appreciate the conciseness of your code but simon's seems 4 times faster with 4 times less allocations.

> Maybe something like checking for average degree with a reasonable tolerance?

That equivalent to check the number of edges, so it seems rather weak. Maybe something like degree distribution ?

abraunst:
I've rebased

abraunst:
> When I benchmark against the implementation proposed by @simonschoelly in #150, I get this (only for undirected graph because simon's code seems broken for directed graphs):
> 
> ```julia
> julia> for N in (500, 5000), is_directed in (false, ), p in (10/N, 0.5, 1-10/N)
>            @btime erdos_renyi_ss($N, $p; is_directed=$is_directed)
>        end
>   164.086 μs (504 allocations: 77.45 KiB)
>   2.410 ms (504 allocations: 1.02 MiB)
>   2.538 ms (504 allocations: 1.95 MiB)
>   1.927 ms (5007 allocations: 792.03 KiB)
>   351.821 ms (10007 allocations: 95.79 MiB)
>   435.251 ms (10007 allocations: 190.78 MiB)
> 
> julia> for N in (500, 5000), is_directed in (false, ), p in (10/N, 0.5, 1-10/N)
>            @btime erdos_renyi_pr($N, $p; is_directed=$is_directed)
>        end
>   450.744 μs (1296 allocations: 191.47 KiB)
>   10.534 ms (3017 allocations: 4.76 MiB)
>   17.142 ms (3017 allocations: 5.69 MiB)
>   5.269 ms (13225 allocations: 2.03 MiB)
>   1.107 s (42520 allocations: 361.60 MiB)
>   2.319 s (47520 allocations: 788.97 MiB)
> ```
> 
> I appreciate the conciseness of your code but simon's seems 4 times faster with 4 times less allocations.
> 
> > Maybe something like checking for average degree with a reasonable tolerance?
> 
> That equivalent to check the number of edges, so it seems rather weak. Maybe something like degree distribution ?

That looks rather unexpected to me. I'll have a better look at that.

gdalle:
> EDIT: the problem most likely is that SimpleGraphFromIterator is slow because it cannot preallocate memory for the adjacency lists and just push!es, requiring three or four allocations per vertex instead of one.

Could we `sizehint!` it if the iterator has a length?

abraunst:
> > EDIT: the problem most likely is that SimpleGraphFromIterator is slow because it cannot preallocate memory for the adjacency lists and just push!es, requiring three or four allocations per vertex instead of one.
> 
> Could we `sizehint!` it if the iterator has a length?

seems difficult because these are per-vertex lists, whereas the iterator in question is on the full list of edges...

etiennedeg:
When removing the `SimpleGraphFromIterator` call (and replacing it by an empty graph), I get 
```
  1.550 ms (4314 allocations: 174.41 KiB)
  25.122 ms (64000 allocations: 2.94 MiB)
  46.153 ms (124133 allocations: 5.70 MiB)
  16.399 ms (73717 allocations: 2.20 MiB)
  3.212 s (38215347 allocations: 774.12 MiB)
```

abraunst:
> ```julia
> for N in (500, 5000), is_directed in (false, ), p in (10/N, 0.5, 1-10/N)
>            @btime erdos_renyi_pr($N, $p; is_directed=$is_directed)
>        end
> ```

Are you sure? This seems too strange, The number of allocations should be about N. Cannot reproduce locally.

etiennedeg:
Weird, done it again, I got:
```
  107.043 μs (505 allocations: 57.03 KiB)
  2.767 ms (505 allocations: 532.59 KiB)
  2.753 ms (505 allocations: 1004.34 KiB)
  1.091 ms (5006 allocations: 553.33 KiB)
  300.378 ms (5006 allocations: 48.11 MiB)
  321.689 ms (5006 allocations: 95.64 MiB)
```
I don't know how I got the previous results

etiennedeg:
The biggest advantage of Simon's is that the neighbors are already sorted, so it is much less costly to build the adjacency lists. No matter what, with this approach, we are going to pay a cost to retrieve the adjacency lists.

abraunst:
> The biggest advantage of Simon's is that the neighbors are already sorted, so it is much less costly to build the adjacency lists. No matter what, with this approach, we are going to pay a cost to retrieve the adjacency lists.

The order is the same (Simon's code actually copies `randsubseq` which gives a sorted result). A modified `SimpleGraphFromIterator` already improves things significanly:

```
63.205 μs (505 allocations: 95.14 KiB)
  1.880 ms (505 allocations: 1.50 MiB)
  2.563 ms (505 allocations: 2.89 MiB)
  668.423 μs (5007 allocations: 955.06 KiB)
  236.858 ms (10007 allocations: 143.50 MiB)
  307.496 ms (10007 allocations: 286.03 MiB)
```

against Simon's

```
39.659 μs (504 allocations: 77.61 KiB)
  849.891 μs (504 allocations: 1.02 MiB)
  721.155 μs (504 allocations: 1.95 MiB)
  432.917 μs (5007 allocations: 791.56 KiB)
  120.317 ms (10007 allocations: 95.76 MiB)
  113.731 ms (10007 allocations: 190.78 MiB)
```

````julia
function SimpleGraphFromIterator(edges, nv)
    deg = zeros(Int, nv)
    ne = 0
    @inbounds for e in edges
        u, v = e.src, e.dst
        @assert (u,v) ⊂ eachindex(deg)
        deg[u] += 1
        deg[v] += 1
        ne += 1
    end

    fadjlist = zeros.(Int, deg)
    deg .= 0

    @inbounds for e in edges
        u, v = e.src, e.dst
        fadjlist[u][deg[u]+=1] = v
        fadjlist[v][deg[v]+=1] = u
    end
    return SimpleGraph(ne, fadjlist)
end
```

etiennedeg:
Simon's, but replacing the call to `_sample_with_replacement!` by `randsubseq!`
```
  217.053 μs (504 allocations: 77.73 KiB)
  3.721 ms (504 allocations: 1.02 MiB)
  4.865 ms (535 allocations: 2.00 MiB)
  2.517 ms (5023 allocations: 833.01 KiB)
  493.546 ms (10052 allocations: 95.83 MiB)
  642.595 ms (10055 allocations: 190.89 MiB)
```

abraunst:
> Simon's, but replacing the call to `_sample_with_replacement!` by `randsubseq!`, so does not improve.
> 
> ```
>   217.053 μs (504 allocations: 77.73 KiB)
>   3.721 ms (504 allocations: 1.02 MiB)
>   4.865 ms (535 allocations: 2.00 MiB)
>   2.517 ms (5023 allocations: 833.01 KiB)
>   493.546 ms (10052 allocations: 95.83 MiB)
>   642.595 ms (10055 allocations: 190.89 MiB)
> ```

amazing... these are almost the same code!

etiennedeg:
There is a gotcha in your implementation, the adjacency lists are not sorted... (This is a requirement for `SimpleGraph`). So you indeed have to pay more to sort the adjacency lists.

abraunst:
> There is a gotcha in your implementation, the adjacency lists are not sorted... (This is a requirement for `SimpleGraph`). So you indeed have to pay more to sort the adjacency lists.
> 
> EDIT: If we change trianglemap to not use modulo offset by 1, then the edges are returned in an order similar to simon's.

oh I see. Indeed, this can be easily fixed. Thanks for all the feedback! I will try to come up with something cleaner in the next few days, but otherwise we can use some versione of Simon's code. :smile:

etiennedeg:
Well, in fact it was not only the mod1 that caused vertices to not be sorted. Using your old definition of trianglemap (which follows simon's ordering), I was able to cook something, but it is not quite as fast.

PR #215, opened 2023-01-14
Fix could not import Graphs.prufer_decode warning

simonschoelly:
There was an incorrect import statement for `prufer_code` in the `SimpleGraphs` submodule. This PR removes it.

### DIFF
--- a/src/SimpleGraphs/SimpleGraphs.jl
+++ b/src/SimpleGraphs/SimpleGraphs.jl
@@ -36,8 +36,7 @@ import Graphs:
     num_self_loops,
     insorted,
     squash,
-    rng_from_rng_or_seed,
-    prufer_decode
+    rng_from_rng_or_seed
 
 export AbstractSimpleGraph,
     AbstractSimpleEdge,


simonschoelly:
@SimonCoste would you mind doing a quick review?

SimonCoste:
How did you get this warning ?

simonschoelly:
Usually one does not notice, but when Graphs.jl gets precompiled there is a warning:
```julia
julia> using Graphs
[ Info: Precompiling Graphs [86223c79-3864-5bf0-83f7-82e725a168b6]
WARNING: could not import Graphs.prufer_decode into SimpleGraphs
```

I did't realize it, when I made this PR, but I think the reason is, that in `src/Graphs.jl` we first include `src/SimpleGraphs/SimpleGraphs.jl` and only afterwards `src/trees/prufer.jl` - so then the `prufer_decode` symbol simply does not exist.
But including `src/trees/prufer.jl`  before would also not help, because that file relies on the definition of `SimpleGraph`.

I can't think of any pretty solution to that - but maybe we should simply move all the code into the `SimpleGraphs` submodule? Another solution would probably be, to declare an function without methods with
```julia
    function prufer_decode end
```
right before including `src/SimpleGraphs/SimpleGraphs.jl` - this is also not very nice though, because then we would keep the cyclic dependency.

SimonCoste:
I feel like including `prufer_decode` and other tree-related functions inside `SimpleGraphs` is really odd, since there are almost no other functions/algorithms implemented in `SimpleGraphs`.

What about modifying `trees/prufer.jl` so that it does not rely on `SimpleGraphs` ?  It only uses the `SimpleGraph{T}` type in two places : 
- as a return type annotation for `prufer_decode ` in https://github.com/JuliaGraphs/Graphs.jl/blob/c4360cfaca3936c3d3f784063ad312205cb4fdfe/src/trees/prufer.jl#L40
- as a type annotation in `prufer_encode` in https://github.com/JuliaGraphs/Graphs.jl/blob/c4360cfaca3936c3d3f784063ad312205cb4fdfe/src/trees/prufer.jl#L74

Maybe the simplest thing is to get rid of those in a consistent manner, by replacing them with `Graph `.

gdalle (approved):

PR #218, opened 2023-01-28
Added missing backtick to `laplacian_matrix` docstring

Nagefire:
Adds a backtick that breaks the formatting in the docstring for `laplacian_matrix`.


gdalle (approved):
Thanks for the fix!

PR #229, opened 2023-02-17
Fix #143

### LINKED ISSUE #143: Code implicitly assumes that the heuristic output is of the same type as the graph weights.
When using A* I was attempting to use a heuristic function that return a float, but the default weights of the graph were ints.

Here is the example

```julia
using Graphs, LinearAlgebra, Random
Random.seed!(0)

xyz = rand(4000, 3)
g = random_regular_graph(4000, 4)
src = rand(1:4000)
dst = rand(1:4000)
h(v) = Int(round(norm(xyz[v, :] - xyz[dst, :])))
edges = a_star(g, src, dst, weights(g), h)

# The above works since I convert to int, below does not, I get InexactError: Int64(1.7553136893610342)

h(v) = norm(xyz[v, :] - xyz[dst, :])
edges = a_star(g, src, dst, weights(g), h)
```

gdalle:
Add a recommendation to the docs saying that the heuristic for A* should return values that have the same type as the edge weights


etiennedeg (approved):

etiennedeg:
Keep the issue open though, I think it should be possible to find a solution to this

PR #232, opened 2023-03-07
Eulerian cycles/trails for undirected graphs

thchr:
The implementation is mostly adapted from SimpleGraphs.jl.

Tagging @etiennedeg who pointed out the lack of tooling for Eulerian tours/cycles as well as the implementation in SimpleGraphs.jl.

I didn't add similar functionality for directed graphs: that could be a follow up for anyone needing it.


thchr:
Gentle bump: this should be good to go (new commit today: I had forgotten to include the new tests in `test/runtests.jl`, hence the poor coverage above; fixed now).

gdalle:
Running the tests, if they pass I'll review, if not I will haunt your nightmares forever

gdalle:
WELL WELL

gdalle:
I think it's a matter of dealing with exceptions in a 1.6-compatible manner

thchr:
Yeah, I will push a fix later today 👍

thchr:
Okay, should be fixed now 👍 !

gdalle:
> Okay, should be fixed now 👍 !

Oh reallyN

thchr:
> > Okay, should be fixed now 👍 !
> 
> Oh really?

😭
Should be now though, for real this time, again...

gdalle:
I do love a good old running gag

gdalle:
Do you need a hand?

thchr:
My god; somehow I didn't commit or push the fix yesterday... The pain :(

gdalle:
The suspense is at its highest

thchr:
Remaining failures on v1 and nightly appear to just be spurious `Missing`-issues from JET, i.e., unimportant cf. #271.

gdalle:
#271 has been merged, so you can probably merge master into this branch and the errors will be fixed

thchr:
> #271 has been merged, so you can probably merge master into this branch and the errors will be fixed

Oki, done 👍

gdalle:
Oh my, look at that! The sweet sweet perfume of passing tests

src/traversals/eulerian.jl
```diff
@@ -0,0 +1,93 @@
...
+    eulerian(g::AbstractSimpleGraph{T}[, u::T]) --> T[]
+
+Returns a [Eulerian trail or cycle](https://en.wikipedia.org/wiki/Eulerian_path) through an
+undirected graph `g`, starting at `u`, returning a vector listing the vertices of `g` in the
+order that they are traversed. If no such trail or cycle exists, throws an error.
+
+## Optional arguments
+- If `u` is omitted, a Eulerian trail or cycle is computed with `u = first(vertices(g))`.
+"""
+function eulerian(g::AbstractSimpleGraph{T}, u::T=first(vertices(g))) where {T}
```
gdalle, line 15: Why not an `AbstractGraph`?

src/traversals/eulerian.jl
```diff
@@ -0,0 +1,93 @@
+# Adapted from SimpleGraphs.jl [Copyright (c) 2014, Ed Scheinerman]:
+#    https://github.com/scheinerman/SimpleGraphs.jl/blob/master/src/simple_euler.jl
+# Reproduced under the MIT Expat License.
+
+"""
+    eulerian(g::AbstractSimpleGraph{T}[, u::T]) --> T[]
+
+Returns a [Eulerian trail or cycle](https://en.wikipedia.org/wiki/Eulerian_path) through an
```
gdalle, line 8: Is it a trail or a cycle?

src/traversals/eulerian.jl
```diff
@@ -0,0 +1,93 @@
...
+
+## Optional arguments
+- If `u` is omitted, a Eulerian trail or cycle is computed with `u = first(vertices(g))`.
+"""
+function eulerian(g::AbstractSimpleGraph{T}, u::T=first(vertices(g))) where {T}
+    is_directed(g) && error("`eulerian` is not yet implemented for directed graphs")
+    
+    _check_eulerian_input(g, u) # perform basic sanity checks
+
+    g′ = SimpleGraph{T}(nv(g)) # copy `g`
```
gdalle, line 20: Why do you need a copy?

src/traversals/eulerian.jl
```diff
@@ -0,0 +1,93 @@
...
+    for e in edges(g)
+        add_edge!(g′, e)
+    end
+
+    return _eulerian!(g′, u)
+end
+
+function _eulerian!(g::AbstractSimpleGraph{T}, u::T) where {T}
+    # TODO: This uses Fleury's algorithm which is O(|E|²) in the number of edges |E|.
```
gdalle, line 29: It's okay, correctness before speed

src/traversals/eulerian.jl
```diff
@@ -0,0 +1,93 @@
...
+                    # not a bridge/cut-edge; add to trail
+                    rem_edge!(g, u, w)
+                    push!(trail, u)
+                    u = w
+                    break
+                end
+            end
+        end
+    end
+    error("unreachable reached")
```
gdalle, line 65: What is this?

src/traversals/eulerian.jl
```diff
@@ -0,0 +1,93 @@
...
+            w = first(Nu)
+            rem_edge!(g, u, w)
+            nverts -= 1
+            push!(trail, u)
+            u = w
+        else
+            # otherwise, pick whichever neighbor is not a bridge/cut-edge
+            bs = bridges(g)
+            for w in Nu
+                if all(e -> e ≠ Edge(u, w) && e ≠ Edge(w, u), bs)
```
gdalle, line 55: Won't work if the graph has generic edges

src/traversals/eulerian.jl
```diff
@@ -0,0 +1,93 @@
...
+        if length(Nu) == 1
+            # if only one neighbor, delete and move on
+            w = first(Nu)
+            rem_edge!(g, u, w)
+            nverts -= 1
+            push!(trail, u)
+            u = w
+        else
+            # otherwise, pick whichever neighbor is not a bridge/cut-edge
+            bs = bridges(g)
```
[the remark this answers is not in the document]
gdalle: Is there a way to compute bridges relative to a vertex?

gdalle (changes requested):
Nice contribution! A few remarks but overall good job and solid test suite

src/traversals/eulerian.jl
thchr: Afaict, [the nomenclature](https://en.wikipedia.org/wiki/Eulerian_path#Definition) is that it is a cycle if the start and end vertices are identical - and otherwise, it's a trail.

src/traversals/eulerian.jl
thchr: The algorithm below is relying on removing edges from the graph as it is doing the traversal; i.e., it destructively modifies the graph.

src/traversals/eulerian.jl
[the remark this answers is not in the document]
thchr: Yeah, the repeated calls to `bridges` is what gives the Fleury algorithm the O(E^2) scaling. This could be avoided with Hierholzer's algorithm, in principle. From what I can tell, there isn't currently a way to compute bridges relative to a vertex in Graphs.jl: would be cool to have though.

src/traversals/eulerian.jl
thchr: What would be a good way to check this then?

src/traversals/eulerian.jl
gdalle: What I mean is, how do I know which one will be returned?

src/traversals/eulerian.jl
thchr: This is to help inference: otherwise, it may think that `nothing` is a valid return value.

src/traversals/eulerian.jl
thchr: Okay, added some explanations to the docstring of what is meant.

src/traversals/eulerian.jl
thchr: Ah, this was because I was not aware of the `IsDirected` trait when I wrote this. Changed to `AbstractGraph` now.

test/traversals/eulerian.jl
```diff
@@ -0,0 +1,36 @@
+@testset "Eulerian tours/cycles" begin
```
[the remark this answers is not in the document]
thchr: Done; added some convenience constructors for `GenericGraph` to the `Test` module also.

src/traversals/eulerian.jl
thchr: Done.

thchr:
Thanks for review @gdalle - latest version should address all issues, I believe.
For once, I also tested locally and things pass there - so the CI suspense should be slightly lower than previously :)

src/traversals/eulerian.jl
```diff
@@ -0,0 +1,103 @@
...
+                    u = w
+                    break
+                end
+            end
+        end
+    end
+    error("unreachable reached")
+end
+
+@inline function _excludes_edge(u, w, e::AbstractEdge)
```
gdalle, line 73: I think this belongs in the core utilities somewhere

src/traversals/eulerian.jl
```diff
@@ -0,0 +1,103 @@
+# Adapted from SimpleGraphs.jl [Copyright (c) 2014, Ed Scheinerman]:
+#    https://github.com/scheinerman/SimpleGraphs.jl/blob/master/src/simple_euler.jl
+# Reproduced under the MIT Expat License.
+
+"""
+    eulerian(g::AbstractSimpleGraph{T}[, u::T]) --> T[]
+
+Returns a [Eulerian trail or cycle](https://en.wikipedia.org/wiki/Eulerian_path) through an
+undirected graph `g`, starting at vertex `u`, returning a vector listing the vertices of `g`
+in the order that they are traversed. If no such trail or cycle exists, throws an error.
```
gdalle, line 10: How do I know whether the result will be a trail or a cycle? What does it depend on?

src/traversals/eulerian.jl
```diff
@@ -0,0 +1,103 @@
...
+            w = first(Nu)
+            rem_edge!(g, u, w)
+            nverts -= 1
+            push!(trail, u)
+            u = w
+        elseif length(Nu) == 0
+            error("graph is not connected: a eulerian cycle/trail does not exist")
+        else
+            # otherwise, pick whichever neighbor is not a bridge/cut-edge
+            bs = bridges(g)
```
gdalle, line 58: Add a TODO here for complexity

src/traversals/eulerian.jl
gdalle: Only on the degree of the starting vertex?

gdalle (changes requested):

gdalle:
Actually, you know what? This is plenty good enough to merge.

src/Test/Test.jl
```diff
@@ -36,6 +36,10 @@ struct GenericGraph{T} <: Graphs.AbstractGraph{T}
     g::SimpleGraph{T}
 end
 
+function GenericGraph(elist::Vector{Graphs.SimpleGraphEdge{T}}) where {T<:Integer}
+    GenericGraph{T}(SimpleGraph(elist))
+end
```
simonschoelly, line 41: @thchr I spotted this too late - I think we should not implement any method (this includes constructors) that specialize on a `GenericGraph` or `GenericDiGraph`. The reason is, that we want to use `GenericGraph` as a placeholder for an arbitrary `AbstractGraph` to verify functions that take an `AbstractGraph` work correctly.

simonschoelly: I mean the issue I see here is, if someone writes a function like
```julia
function f(g::AbbstractGraph)
    T = typeof(g)
    return T([Edge(1, 2)])
end
```
then this function might fail for an arbitrary `AstractGraph` but not for a `GenericGraph`. I don't think it will break anything soon though, so there is no rush, but I can make a PR that works around that.

I would propose, if we need a workaround for constructors so that it will be easier to write tests, we just implement the functions `generic_graph` and `generic_digraph` -  maybe I already did such a thing in one of my PRs but I don't remember exactly.

gdalle: Sounds good to me :)

Issue #98, opened 2022-01-29
Create functions for constructing `SimpleGraph` and `SimpleDiGraph` from an `AbstractGraph`.


simonschoelly:
As mentioned in #88 , we should have some constructors that take an `AbstractGraph` and create a `SimpleGraph` or `SimpleDiGraph` from that. I.e. the functions that we need are:
```julia
SimpleGraph(g::AbstractGraph)
SimpleGraph{T}(g::AbstractGraph) where {T <: Integer}
SimpleDiGraph(g::AbstractGraph)
SimpleDiGraph{T}(g::AbstractGraph) where {T <: Integer}
```
It is not entirely clear what should happen if `g` is a directed graph and we try to construct an undirected graph from that, but to be consistent with what we already have in similar functions, we should probably add an edge between two vertices `u` and `v` if the input graph has an edge `u -> v` **or** an edge `v -> u`. Alternatively we could add a keyword argument
```julia
SimpleGraph(g::AbstractGraph; weakly_connected=true)
SimpleGraph{T}(g::AbstractGraph; weakly_connected=true) where {T <: Integer}
```
that would allow one to specify the expected behavior. Maybe there is also a better name than `weakly_connected` for that.

PR #237, opened 2023-03-21
create functions for constructing SimpleGraph and SimpleDiGraph from …

pgrepds:
I have tried to implement [#98](https://github.com/JuliaGraphs/Graphs.jl/issues/98).  In particular, I have decided to create an undirected edge if their appears to be an directed edge if one tries to construct a `SimpleGraph` from an `AbstractGraph` that happens to be directed in order to be consistent for the already existing constructor for constructing a `SimpleGraph` from a `SimpleDiGraph`.

Moreover, if one tries to construct a `SimpleDiGraph` from an `AbstractGraph` that happens to be undirected, then two directed edges will be added.

This is my first time contributing to a Julia package, so I appreciate any comment or suggestion regarding code style.

I wasn't sure if using the `add_edge` function is sufficient (or efficient), so I might need to implement this differently if you have any suggestions.

I also added a few tests , but there may still be cases here that I have not considered.

### DIFF
--- a/src/SimpleGraphs/simpledigraph.jl
+++ b/src/SimpleGraphs/simpledigraph.jl
@@ -26,6 +26,31 @@ end
 
 eltype(x::SimpleDiGraph{T}) where {T} = T
 
+# SimpleDiGraph(g)
+"""
+    SimpleDiGraph{T}(g::AbstractGraph)
+
+Construct a `SimpleDiGraph{T}` from an `AbstractGraph`. 
+If `g` is undirected, then every undirected edge x - y in `g` is added 
+as two directed edges x -> y and y -> x.
+"""
+SimpleDiGraph(g::AbstractGraph) = SimpleDiGraph{Int}(g)
+
+function SimpleDiGraph{T}(g::AbstractGraph) where {T<:Integer}
+    simpleDiGraph = SimpleDiGraph(nv(g))
+    if !is_directed(g)
+        @inbounds for e in edges(g)
+            add_edge!(simpleDiGraph, src(e), dst(e))
+            add_edge!(simpleDiGraph, dst(e), src(e))
+        end
+    else
+        @inbounds for e in edges(g)
+            add_edge!(simpleDiGraph, src(e), dst(e))
+        end
+    end
+    return simpleDiGraph
+end
+
 # DiGraph{UInt8}(6), DiGraph{Int16}(7), DiGraph{Int8}()
 """
     SimpleDiGraph{T}(n=0)
--- a/src/SimpleGraphs/simplegraph.jl
+++ b/src/SimpleGraphs/simplegraph.jl
@@ -21,6 +21,26 @@ end
 
 eltype(x::SimpleGraph{T}) where {T} = T
 
+# SimpleGraph(g)
+"""
+    SimpleGraph{T}(g::AbstractGraph)
+
+Construct a `SimpleGraph{T}` from an `AbstractGraph`. 
+If `g` is a `SimpleDiGraph`, then every directed edge in `g` is 
+added as an undirected edge.
+"""
+SimpleGraph(g::AbstractGraph) = SimpleGraph{Int}(g)
+
+# SimpleGraph{UInt8}(g)
+function SimpleGraph{T}(g::AbstractGraph) where {T<:Integer}
+    simpleGraph = SimpleGraph(nv(g))
+    @inbounds for e in edges(g)
+        # it seems that the `add_edge` function already handles multiple edges efficiently (e.g. in the case of directed edges 1 -> 2 and 2 -> 1).
+        add_edge!(simpleGraph, src(e), dst(e))
+    end
+    return simpleGraph
+end
+
 # Graph{UInt8}(6), Graph{Int16}(7), Graph{UInt8}()
 """
     SimpleGraph{T}(n=0)
--- a/test/simplegraphs/simplegraphs.jl
+++ b/test/simplegraphs/simplegraphs.jl
@@ -498,4 +498,145 @@ using Random: Random
     end
     # codecov for has_edge(::AbstractSimpleGraph, x, y)
     @test @inferred has_edge(DummySimpleGraph(), 1, 2)
+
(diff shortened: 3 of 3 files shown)


src/SimpleGraphs/simpledigraph.jl
```diff
@@ -26,6 +26,31 @@ end
...
+    SimpleDiGraph{T}(g::AbstractGraph)
+
+Construct a `SimpleDiGraph{T}` from an `AbstractGraph`. 
+If `g` is undirected, then every undirected edge x - y in `g` is added 
+as two directed edges x -> y and y -> x.
+"""
+SimpleDiGraph(g::AbstractGraph) = SimpleDiGraph{Int}(g)
+
+function SimpleDiGraph{T}(g::AbstractGraph) where {T<:Integer}
+    simpleDiGraph = SimpleDiGraph(nv(g))
```
simonschoelly, line 40:
```suggestion
    simpleDiGraph = SimpleDiGraph{T}(nv(g))
```

src/SimpleGraphs/simplegraph.jl
```diff
@@ -21,6 +21,26 @@ end
...
+Construct a `SimpleGraph{T}` from an `AbstractGraph`. 
+If `g` is a `SimpleDiGraph`, then every directed edge in `g` is 
+added as an undirected edge.
+"""
+SimpleGraph(g::AbstractGraph) = SimpleGraph{Int}(g)
+
+# SimpleGraph{UInt8}(g)
+function SimpleGraph{T}(g::AbstractGraph) where {T<:Integer}
+    simpleGraph = SimpleGraph(nv(g))
+    @inbounds for e in edges(g)
```
simonschoelly, line 37: I don't think that `@inbounds` has any effect here - the only place were some bound checks can happen in this loop are inside the `add_edge!` function, but that function does not allow the propagation of `@inbounds` anyways.

src/SimpleGraphs/simplegraph.jl
```diff
@@ -21,6 +21,26 @@ end
...
 
+# SimpleGraph(g)
+"""
+    SimpleGraph{T}(g::AbstractGraph)
+
+Construct a `SimpleGraph{T}` from an `AbstractGraph`. 
+If `g` is a `SimpleDiGraph`, then every directed edge in `g` is 
+added as an undirected edge.
+"""
+SimpleGraph(g::AbstractGraph) = SimpleGraph{Int}(g)
```
simonschoelly, line 32:
```suggestion
SimpleGraph(g::AbstractGraph) = SimpleGraph{eltype(g)}(g)
```
or 
```suggestion
SimpleGraph(g::AbstractGraph{T}) where {T} = SimpleGraph{T}(g)
```
so in that case, the constructed graph takes the eltype from the passed in graph. This is more consistent with existing constructors such as
```julia
SimpleGraph(g::SimpleGraph{T}) where {T} = SimpleGraph{T}(g)
```

src/SimpleGraphs/simpledigraph.jl
```diff
@@ -26,6 +26,31 @@ end
...
 
+# SimpleDiGraph(g)
+"""
+    SimpleDiGraph{T}(g::AbstractGraph)
+
+Construct a `SimpleDiGraph{T}` from an `AbstractGraph`. 
+If `g` is undirected, then every undirected edge x - y in `g` is added 
+as two directed edges x -> y and y -> x.
+"""
+SimpleDiGraph(g::AbstractGraph) = SimpleDiGraph{Int}(g)
```
simonschoelly, line 37:
```suggestion
SimpleDiGraph(g::AbstractGraph) = SimpleDiGraph{eltype(g)}(g)
```
or
```suggestion
SimpleDiGraph(g::AbstractGraph{T}) where {T} = SimpleDiGraph{T}(g)
```
see below for the reason

src/SimpleGraphs/simpledigraph.jl
```diff
@@ -26,6 +26,31 @@ end
 
 eltype(x::SimpleDiGraph{T}) where {T} = T
 
+# SimpleDiGraph(g)
+"""
+    SimpleDiGraph{T}(g::AbstractGraph)
```
simonschoelly, line 31:
```suggestion
    SimpleDiGraph(g::AbstractGraph)
    SimpleDiGraph{T}(g::AbstractGraph)
```
Lets either add both signatures to the docstring, or mention that `T` is optional and what happens if it is omitted.

src/SimpleGraphs/simplegraph.jl
```diff
@@ -21,6 +21,26 @@ end
 
 eltype(x::SimpleGraph{T}) where {T} = T
 
+# SimpleGraph(g)
+"""
+    SimpleGraph{T}(g::AbstractGraph)
+
+Construct a `SimpleGraph{T}` from an `AbstractGraph`. 
+If `g` is a `SimpleDiGraph`, then every directed edge in `g` is 
+added as an undirected edge.
```
simonschoelly, line 30:
```suggestion
Construct a `SimpleGraph{T}` from an `AbstractGraph`. 
If `g` is a directed, then every directed edge in `g` is 
added as an undirected edge.
```

simonschoelly: Also, how about we let the user specify what happens if `g` is directed? Then the signature could look something like
```julia
SimpleGraph(g::AbstractGraph; only_strong_edges::Bool = false)
```
or maybe there is a better name for `only_strong_edges` - I took the inspiration for the name from the concept of weak and strong components in directed graphs.

simonschoelly:
Thanks for this PR ... as you can see we are a bit understaffed at the moment, that's why it took so long for someone to take a look. I added some suggestion, what we could change - I had not yet time to look thoroughly at your tests, but I like that you started using `GenericGraph` that is something we should do in more tests in the future.

Using `add_edge!` might not be the most efficient solution in terms of performance because we have to resize the adjacency lists all the time,  but as the result is still correct. we can use that for now, and then if someone got some time in the future, we can implement a more efficient solution later.

Issue #192, opened 2022-11-20
Doctests are failing


gdalle:
Tried to add doctests to `runtests.jl` in #190 and they exploded. My guess is the documentation outputs are slightly outdated

gdalle:
PR merged but the problematic doctests remain. Copying them 

- modularity.jl has a reference on SimpleWeightedGraph which is not included by Graphs.jl.
- transitivity.jl in the function transitiveClosure. The issue here is that Documenter.jl truncates the output. Thus, it is not fully visible. We could compare the truncated output instead. However, it might then differ from the output which is seen by the user.
- shortestpaths.jl in the function has_negative_weight_cycle in src/Experimental. The SPFA package is not found. The same happens in spfa.jl.
- spfa.jl in spfa_shortest_paths. The output is compared to an error message. However, the full stack trace is obviously not included.
- euclidiagraphs.jl uses random values. We might want to add a local doc filter.
- randgraphs.jl uses random values. We might want to add a local doc filter here as well.

PR #239, opened 2023-03-28
fix doctests

pgrepds:
I have fixed all `jldoctest` as proposed by issue [#192](https://github.com/JuliaGraphs/Graphs.jl/issues/192). 

There are some cases for which I have deactivated the `jldoctest` and which might need further attention.

I have deactivated the following `jldoctest`s

- `modularity.jl` has a reference on `SimpleWeightedGraph` which is not included by Graphs.jl.
- `transitivity.jl` in the function `transitiveClosure`. The issue here is that Documenter.jl truncates the output. Thus, it is not fully visible. We could compare the truncated output instead. However, it might then differ from the output which is seen by the user.
- `shortestpaths.jl` in the function `has_negative_weight_cycle` in `src/Experimental`. The `SPFA` package is not found. The same happens in `spfa.jl`.
- `spfa.jl` in `spfa_shortest_paths`. The output is compared to an error message. However, the full stack trace is obviously not included.
- `euclidiagraphs.jl` uses random values. We might want to add a local doc filter.
- `randgraphs.jl` uses random values. We might want to add a local doc filter here as well.

### DIFF
--- a/docs/Manifest.toml
+++ b/docs/Manifest.toml
@@ -1,8 +1,12 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.8.2"
+julia_version = "1.9.0-rc2"
 manifest_format = "2.0"
 project_hash = "4fb688147559360c201b0f727fe4bd07d6143f40"
+
+[[deps.ArgTools]]
+uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"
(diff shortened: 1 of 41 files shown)


gdalle (approved):
Thanks for the painstaking work @pgrepds, this will definitely increase code reliability!!!

PR #246, opened 2023-04-13
Fix package docstring

Kolaru:
It can be slightly confusing for user who don't know the package history to say that Graphs.jl is faster than Graphs.jl.

Alos the doc link seemed to be broken.


aurorarossi (approved):

aurorarossi:
Thanks for pointed that out and for the PR!

gdalle (approved):

PR #250, opened 2023-04-27
Improve performance of bfs functions

Tortar:
I added an optimization to these bfs functions which I benchmarked with

```julia
using Graphs, BenchmarkTools

gs = [path_graph(500),
      path_graph(5000),
      path_graph(50000),
      erdos_renyi(50, 0.5, seed=123),
      erdos_renyi(500, 0.5, seed=123),
      erdos_renyi(5000, 0.5, seed=123),
      complete_graph(50),
      complete_graph(500),
      complete_graph(5000)]

for g in gs
    n = nv(g)
    u = Int(n/5)
    dists = vec(fill(typemax(Int), (1, n)))
    t_1 = @belapsed gdistances!($g, $u, $dists)
    t_2 = @belapsed bfs_parents($g, $u)
    println("$t_1 $t_2;")
end
```
and then I calculated the relative speed-up for the graphs in `gs` before and after this pr:

``` julia
 gdistances! bfs_parents
     0.99      0.99            --> path_graph(500)
     0.98      0.99            --> path_graph(5000)
     0.99      1.00            --> path_graph(50000)
     3.18      3.14            --> erdos_renyi(50, 0.5, seed=123)
    16.91      15.42           --> erdos_renyi(500, 0.5, seed=123)
   126.52      120.08          --> erdos_renyi(5000, 0.5, seed=123)
     5.40      5.19            --> complete_graph(50)
    46.14      40.11           --> complete_graph(500)
   512.26      475.42          --> complete_graph(5000)
```

The Path Graph is kind of the worst case scenario, and the time is approximately equal to the algorithm before this pr, but, as you can see, with a well connected Erdos Renyi the speed-up becomes very good, and finally I choose to benchmark a Complete Graph since it should be the best case scenario. 

I also ran the tests locally and everything should still work!

### DIFF
--- a/src/traversals/bfs.jl
+++ b/src/traversals/bfs.jl
@@ -43,15 +43,19 @@ end
 function _bfs_parents(g::AbstractGraph{T}, source, neighborfn::Function) where {T}
     n = nv(g)
     visited = falses(n)
-    parents = zeros(T, nv(g))
+    parents = zeros(T, n)
     cur_level = Vector{T}()
     sizehint!(cur_level, n)
     next_level = Vector{T}()
     sizehint!(next_level, n)
+    k = 0
     @inbounds for s in source
-        visited[s] = true
-        push!(cur_level, s)
-        parents[s] = s
+        if !visited[s]
+            visited[s] = true
+            push!(cur_level, s)
+            parents[s] = s
+            k += 1
+        end
     end
     while !isempty(cur_level)
         @inbounds for v in cur_level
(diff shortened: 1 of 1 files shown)


Tortar:
@gdalle If I can ask, what do you think about the following PR since to me it seems a pretty good speed-up :-)

etiennedeg (approved):
Looks good to me

simonschoelly:
This PR adds a heuristic that detects early that every vertex has been finished - it makes sense that it increases the speed in case of graphs with low diameter.

On the other hand, it adds some unnecessary if statements and a variable named `k` that could easily have been called something like `num_visited`. I think we should address such things in a code review instead of just merging them.

etiennedeg:
The if statements are necessary, if some vertices are duplicated in `source` then `k` will be incremented too much.

simonschoelly:
> The if statements are necessary, if some vertices are duplicated in `source` then `k` will be incremented too much.

Ok thats a fair point.

PR #252, opened 2023-05-02
New SimpleGraph constructor from tuples

GiggleLiu:
From the slack discussion. We already have the
`SimpleGraphFromIterator(vector_of_edges)` https://juliagraphs.org/Graphs.jl/dev/first_steps/construction/#Modifying-graphs (@gdalle )

However, I think this new interface might still be very useful
1. It does not require converting tuples to the `Edge` type.
2. Allows setting the number of vertices, which I think is very important when constructing graphs programmingly.

In mathematics, we also use tuple of length 2 to represent a pair of undirected edges $(i, j) \in E$. To me, it is natural notation. Comments are welcomes.

## Example
**Before**
```julia
g = SimpleGraph(Int32(4))

for (i, j) in  [(1, 2), (2, 3)]
    add_edge!(g, i, j)
end
```

**After**
```julia
g = SimpleGraph(Int32(4), [(1, 2), (2, 3)])
```

### DIFF
--- a/src/SimpleGraphs/simplegraph.jl
+++ b/src/SimpleGraphs/simplegraph.jl
@@ -15,8 +15,31 @@ mutable struct SimpleGraph{T<:Integer} <: AbstractSimpleGraph{T}
     end
 end
 
-function SimpleGraph(ne, fadjlist::Vector{Vector{T}}) where {T}
+function SimpleGraph(ne::Integer, fadjlist::Vector{Vector{T}}) where {T}
     return SimpleGraph{T}(ne, fadjlist)
+end
+
+# initialize a graph by a vector of tuples
+"""
+    SimpleGraph{T}(n, tuples_as_edges)
+
+Construct a `SimpleGraph{T}` with `n` vertices, with the edges specified as a vector of tuples.
+The element type `T` is the type of `n`.
+
+## Examples
+```jldoctest
+julia> using Graphs
+
+julia> SimpleGraph(UInt8(10), [(1,2), (2, 3)])
+{10, 2} undirected simple UInt8 graph
+```
+"""
+function SimpleGraph(n::Integer, tuples_as_edges::AbstractVector{Tuple{T,T}}) where {T<:Integer}
+    g = SimpleGraph{typeof(n)}(n)
+    for (i, j) in tuples_as_edges
+        add_edge!(g, i, j)
+    end
+    return g
 end
 
 eltype(x::SimpleGraph{T}) where {T} = T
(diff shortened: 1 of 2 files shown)


simonschoelly:
Thanks for this contribution. In the past we had tried to not extend the number of constructors any further, in order not to grow the already large list of constructors any further, but maybe it it would be useful if instead of having to write
```julia
SimpleGraph(Edge.([
  (1,2),
  (3,4),
]))
```
we can simply write 
```julia
SimpleGraph([
  (1,2),
  (3,4),
])
```
Maybe we can simplify the documentation by having both the same docstring for both constructors, so that it shows up as only one when checking the documentation.

That being said, I think we should stay consistent, and implement `SimpleGraph(list_of_tuples)` instead of `SimpeGraph(n, list_of_tuples)`, as we have done it similarly  for the vector of edges. In the rare case that one wants more isolated vertices, they can always call `add_vertices!`  afterwards, with not much more computational cost.

I guess we might also think about implementing `SimpleGraph{T}(list_of_tuples)` so that one can specify the eltype there and won't have to do that for each tuple - in that case we should also do that for the edge list constructors.  I am not sure why we did not already do that in the past.

Also, if we implement this new constructor, we should also do the same for `SimpleDiGraph`, `SimpleGraphFromIterator` and `SimpleDiGraphFromIterator`.

@gdalle as you were involved in the discussion on slack, you might also have an opinion on this.

GiggleLiu:
> Maybe we can simplify the documentation by having both the same docstring for both constructors, so that it shows up as only one when checking the documentation.

Agreed.

> implement `SimpleGraph(list_of_tuples)` instead of `SimpeGraph(n, list_of_tuples)`

I knew the first interface long ago, but I never use it for productivity. Maybe also due to my using cases are often related to graph gadget finding that dangling vertices are allowed. I think **allowing users to only remember only one API** is important. The first interface can cover a lot of using cases, but I still need to remember multiple ways to construct the graph.

> I guess we might also think about implementing SimpleGraph{T}(list_of_tuples) so that one can specify the eltype there and won't have to do that for each tuple - in that case we should also do that for the edge list constructors. I am not sure why we did not already do that in the past.

In the added constructor, the data type follows the first argument, the number of vertices `n`.

> Also, if we implement this new constructor, we should also do the same for `SimpleDiGraph`, `SimpleGraphFromIterator` and `SimpleDiGraphFromIterator`.

Sure, I can do that.

gdalle:
I'm not sure about the tuple syntax, but I agree with the need for the number of vertices to feature as an argument

gdalle:
On second thought the list of constructors is already quite large, and adding `Edge.(...)` is only a minor inconvenience. I would suggest we close this

gdalle:
No worries, lack of workforce is also the reason behind my reasoning that simpler interfaces are easier to maintain! But please don't feel discouraged to open other issues or PRs in the future

PR #254, opened 2023-06-06
Set initial path length in yen with zero(T)

SuperGrobi:
Says in the title. Should I add tests for this?


gdalle:
I hereby dispense you of tests

gdalle (approved):

PR #262, opened 2023-06-27
Revert the error thrown by badj to NotImplementedError

simonschoelly:
This PR reverts the change that was done to `badj` in #249 , as apparently it is not necessary to change the error in order to satisfy JET.jl.

There are still some good arguments for and also against throwing a custom error message when an interface method is not implemented, but I think that needs a separate discussion as we still have the pattern where we throw `NotImplementedError` in various places, so we should not simply change it for that single function.


gdalle (approved):
Fine by me

gdalle:
Opened an issue in the new API repo to track this

Issue #248, opened 2023-04-17
`issymmetric(::SimpleDiGraph)` not correct


jlapeyre:
This dispatches here
https://github.com/JuliaGraphs/Graphs.jl/blob/53cb58152a19ca62fa01439300948c30035dca6a/src/operators.jl#L498

This of course returns false for any digraph, and digraphs can indeed be symmetric.

I don't know how many meanings of `symmetric` are common in fields using graphs. But wikipedia says this means that for each edge $(v, w)$ there is an edge $(w, v)$.

simonschoelly:
I agree with you here. I think if we change it, we have to do it, so that it is still true for all undirected graphs. For directed graphs I am not sure if there is an easy way to that so have some default for all directed `AbstractGraph` or if we simply should return a not-implemented message.

etiennedeg:
Should it also take into account weights or metadata ?

gdalle:
But Graphs.jl itself doesn't have a method for getting the weight of an edge. Best we could do is to add an optional argument `weights` passing a matrix? But who would ever use this function anyway?

simonschoelly:
> Should it also take into account weights or metadata ?

I think the solution that we currently have in #248 kind of solves this, although certainly there are other ways to handle that. It basically uses the `edges(g)`, `reverse(e)`, `has_edge(g, e)`, and `e1 == e2` functions to verify if a graph is symmetric. If some graph type  has some weights/edge-metadata and correctly implements these functions, then symmetry is correctly checked.

Of course one could argue, that there should be a way to ignore weights/metadata but I am not sure how important that is. perhaps in the future, we can specify that with an additional parameter.

PR #265, opened 2023-06-28
Fix issymmetric

gdalle:
Solve #248

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -511,7 +511,7 @@ length(g::AbstractGraph) = widen(nv(g)) * widen(nv(g))
 ndims(g::AbstractGraph) = 2
 
 @traitfn function issymmetric(g::AG) where {AG <: AbstractGraph; !IsDirected{AG}}
-    return false
+    return true
 end
 
 @traitfn function issymmetric(g::AG) where {AG <: AbstractGraph; IsDirected{AG}}
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -510,11 +510,11 @@ sparse(g::AbstractGraph) = adjacency_matrix(g)
 length(g::AbstractGraph) = widen(nv(g)) * widen(nv(g))
 ndims(g::AbstractGraph) = 2
 
-function issymmetric(g::AG) where {AG <: AbstractGraph; !IsDirected{G}}
+@traitfn issymmetric(g::AG) where {AG <: AbstractGraph; !IsDirected{G}}
     return false
 end
 
-function issymmetric(g::AG) where {AG <: AbstractGraph; IsDirected{G}}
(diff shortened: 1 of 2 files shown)


src/operators.jl
```diff
@@ -509,7 +509,19 @@ sparse(g::AbstractGraph) = adjacency_matrix(g)
...
+
+@traitfn function issymmetric(g::AG) where {AG <: AbstractGraph; !IsDirected{AG}}
+    return true
+end
+
+@traitfn function issymmetric(g::AG) where {AG <: AbstractGraph; IsDirected{AG}}
+    for e in edges(g)
+        if !has_edge(g, dst(e), src(e))
+            return false
+        end
```
simonschoelly, line 521: Can we change this do something like
```julia
if !has_edge(g, reverse(e))
            return false
        end
```
?
Otherwise it will return incorrect values for graphs with different edge types. An other option would be to not provide a default implementation for directed graphs.

test/operators.jl
```diff
@@ -203,6 +203,15 @@
...
 
+    gx = SimpleDiGraph(4)
+    add_edge!(gx, 1, 2)
+    add_edge!(gx, 2, 1)
+    add_edge!(gx, 1, 3)
+    add_edge!(gx, 3, 1)
+    @testset "Matrix operations: $g" for g in testdigraphs(gx)
+        @test @inferred(issymmetric(g))
+    end
+
```
simonschoelly, line 214: Can we rewrite this test in terms of `GenericDiGraph`, now that we have it?

simonschoelly:
Mostly looks good to me, perhaps that one issue about the tests but otherwise I am all for merging.

test/operators.jl
gdalle: I didn't check but I don't think `Generic(Di)Graph` is used anywhere in the tests yet? The corresponding issue #133 is still open and doesn't have a PR associated with it?

gdalle: I agree that in time `GenericGraph` should be everywhere, but inserting it in only one spot seems a bit weird, kinda like removing the "not implemented" error the other day 😉

simonschoelly: It is indeed only uses somehwhere in `traversals/dfs.jl`. We don't have to insert it here, I just approve it for now, although in general I think we can suggest it in new PRs.

I haven't gotten around, but I guess then I will start creating some PRs that will switch to GenericGraph - I don't want to do the whole Graphs package in one PR though, that will be way too big and newer get reviewed.

simonschoelly (approved):

test/operators.jl
gdalle: Maybe GenericGraph could wait for GraphsBase before spreading through our tests?
https://github.com/JuliaGraphs/GraphsBase.jl/issues/6

PR #266, opened 2023-06-28
fix quadratic time in `is_cyclic` and `topological_sort` (fix #263)

### LINKED ISSUE #263: Possible issue with topological_sort_by_dfs
Looks like the function ```topological_sort_by_dfs``` can have a run-time $O(n\cdot n)$ even for sparse graphs. The simplest example to get this behaviour is

```
function gen_graph_topsort(k)
    g = SimpleDiGraph(k+1)
    for i = 1:k
        add_edge!(g, 1, i+1)
    end
    return g
end

g = gen_graph_topsort(10^5)
@time topological_sort_by_dfs(g)
```

For graphs with about 10^5 vertices this already takes a few seconds. For larger graphs, this quickly becomes infeasible (due to the quadratic run-time). Expected behaviour would be that the implementation has worst-case O(n+m) run-time and takes fractions of a second for graphs of this size.

The problem appears to be that this loop https://github.com/JuliaGraphs/Graphs.jl/blob/e5405d105e034ccb9396d30f86c26641ce94c504/src/traversals/dfs.jl#L97 goes through the neighbors of u over and over everytime starting from the beginning of the list.

I.e., it will find ```n``` with ```vcolor[n] == 0``` and break, then (with some further steps) ```vcolor[n]``` will be set to ```2```. Afterwards, again the search over the neighbors of u starts at the first neighbor and goes through them until it finds one with ```vcolor == 0``` and breaks etc.

Thus, for a vertex with $k$ neighbor it can take $O(k^2)$ steps, leading to potentially O(n^2), resp. O(m^2), run-time for sparse graphs.

etiennedeg:
Fixes #263

### DIFF
--- a/src/traversals/dfs.jl
+++ b/src/traversals/dfs.jl
@@ -32,27 +32,30 @@ function is_cyclic end
 end
 # see https://github.com/mauro3/SimpleTraits.jl/issues/47#issuecomment-327880153 for syntax
 @traitfn function is_cyclic(g::AG::IsDirected) where {T,AG<:AbstractGraph{T}}
+    # 0 if not visited, 1 if visited, 2 if in the current dfs path, 3 if fully explored 
     vcolor = zeros(UInt8, nv(g))
-    S = Vector{T}()
+    vertex_queue = Vector{T}()
     for v in vertices(g)
         vcolor[v] != 0 && continue
-        push!(S, v)
+        push!(vertex_queue, v)
         vcolor[v] = 1
-        while !isempty(S)
-            u = S[end]
+        while !isempty(vertex_queue)
+            u = vertex_queue[end]
             if vcolor[u] == 1
                 vcolor[u] = 2
                 for n in outneighbors(g, u)
+                    # we hit a loop when reaching back a vertex of the main path
                     if vcolor[n] == 2
                         return true
                     elseif vcolor[n] == 0
+                        # we store neighbors, but these are not yet on the path
                         vcolor[n] = 1
-                        push!(S, n)
+                        push!(vertex_queue, n)
                     end
                 end
             else
                 vcolor[u] = 3
-                pop!(S)
+                pop!(vertex_queue)
             end
         end
     end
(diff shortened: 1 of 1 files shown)


src/traversals/dfs.jl
```diff
@@ -33,26 +33,25 @@ end
 # see https://github.com/mauro3/SimpleTraits.jl/issues/47#issuecomment-327880153 for syntax
 @traitfn function is_cyclic(g::AG::IsDirected) where {T,AG<:AbstractGraph{T}}
     vcolor = zeros(UInt8, nv(g))
+    S = Vector{T}()
```
gdalle, line 36: Comment what `S` stands for?

src/traversals/dfs.jl
```diff
@@ -87,32 +86,31 @@ function topological_sort_by_dfs end
 @traitfn function topological_sort_by_dfs(g::AG::IsDirected) where {T,AG<:AbstractGraph{T}}
     vcolor = zeros(UInt8, nv(g))
     verts = Vector{T}()
+    S = Vector{T}()
```
gdalle, line 89: Comment what `S` stands for?

src/traversals/dfs.jl
```diff
@@ -33,26 +33,25 @@ end
 # see https://github.com/mauro3/SimpleTraits.jl/issues/47#issuecomment-327880153 for syntax
 @traitfn function is_cyclic(g::AG::IsDirected) where {T,AG<:AbstractGraph{T}}
     vcolor = zeros(UInt8, nv(g))
```
gdalle, line 35: Comment what the color values stand for?

src/traversals/dfs.jl
```diff
@@ -87,32 +86,31 @@ function topological_sort_by_dfs end
 @traitfn function topological_sort_by_dfs(g::AG::IsDirected) where {T,AG<:AbstractGraph{T}}
     vcolor = zeros(UInt8, nv(g))
```
gdalle, line 87: Comment what the color values stand for?

src/traversals/dfs.jl
```diff
@@ -87,32 +86,31 @@ function topological_sort_by_dfs end
...
-            end
-            if w != 0
-                vcolor[w] = 1
-                push!(S, w)
             else
-                vcolor[u] = 2
-                push!(verts, u)
+                vcolor[u] = 3
                 pop!(S)
+                pushfirst!(verts, u)
```
gdalle, line 109: How are we sure that each vertex lands in `verts` exactly once?

gdalle (changes requested):

src/traversals/dfs.jl
etiennedeg: Each vertex is visited at least once (outer loop). Each vertex is visited at most once (added to vertex_heap only if color is 0).
Then vertex is visited once to explore its unvisited subtree, and color goes to 2. Next time it is visited, its color goes to 3, and vertex is removed from queue and added to verts, so not visited anymore.

src/traversals/dfs.jl
```diff
@@ -32,28 +32,30 @@ function is_cyclic end
 end
 # see https://github.com/mauro3/SimpleTraits.jl/issues/47#issuecomment-327880153 for syntax
 @traitfn function is_cyclic(g::AG::IsDirected) where {T,AG<:AbstractGraph{T}}
+    # 0 if not visited, 1 if visited, 2 if in the current dfs path, 3 if fully explored 
     vcolor = zeros(UInt8, nv(g))
+    vertex_queue = Vector{T}()
```
gdalle, line 37: Why a Vector instead of another data structure?

etiennedeg: I kept the last one, why do you want something else ? We only need to push and pop, there are no weight ordering

simonschoelly: There should be some queue data structures in [DataStructures.jl](https://github.com/JuliaCollections/DataStructures.jl) - but I am not sure if and how much more efficient they would be. But if I understand correctly, you always add or remove at the end, so it should be called `stack` not `queue`.

gdalle: I think we should merge this as is (with a renaming) and keep track of the potential for a better stack in a new issue?

gdalle (approved):
Seems fine to me

PR #268, opened 2023-06-29
two sided dijkstra

Dolgalad:
Implemented the Two sided Dijkstra algorithm, still requires documentation and tests


Dolgalad:
I added a very simple test, not sure if it is sufficient

gdalle:
Tests are failing because of #271 which has now been merged. If you merge the `JuliaGraphs:master` branch of into `Dolgalad:bidij` this should work

Project.toml
```diff
@@ -7,13 +7,15 @@ ArnoldiMethod = "ec485272-7323-5ecc-a04f-4719b315124d"
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
```
gdalle, line 10: Why did you add this dependency? It doesn't seem needed and would make the package much heavier to load
```suggestion
```

Project.toml
```diff
@@ -7,13 +7,15 @@ ArnoldiMethod = "ec485272-7323-5ecc-a04f-4719b315124d"
...
 Inflate = "d25df0c9-e2be-5dd7-82c8-3ad0b3e990b9"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SharedArrays = "1a1011a3-84de-559e-8e89-a11a2f7dc383"
 SimpleTraits = "699a6c99-e7fa-54fc-8d76-47d257e15c1d"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+StatsPlots = "f3b207a7-027a-5e70-b257-86293d7955fd"
```
gdalle, line 18: Same remark as for Distributions
```suggestion
```

test/shortestpaths/dijkstra.jl
```diff
@@ -111,4 +111,20 @@
             ds = @inferred(dijkstra_shortest_paths(g, 3, m;maxdist=3.0))
             @test ds.dists == [2, 1, 0, Inf, Inf, 3]
         end
+
+    # bidijkstra_shortest_path
```
gdalle, line 115: It's good to have a test but it would be even better to have more. Can you duplicate the tests from A*, since the algorithms give similar outputs?

src/shortestpaths/dijkstra.jl
```diff
@@ -169,3 +169,115 @@ function dijkstra_shortest_paths(
...
+    src::U,
+    dst::U,
+    distmx::AbstractMatrix{T}=weights(g)
+) where {T<:Real} where {U<:Integer}
+    if src == dst
+        return Int64[]
+    end
+    # keep weight of the best seen path and the midpoint vertex
+    mu, mid_v = typemax(T), -1
+
```
gdalle, line 230: All that white space may not be necessary

src/shortestpaths/dijkstra.jl
```diff
@@ -169,3 +169,115 @@ function dijkstra_shortest_paths(
...
+```
+"""
+function bidijkstra_shortest_path(
+    g::AbstractGraph,
+    src::U,
+    dst::U,
+    distmx::AbstractMatrix{T}=weights(g)
+) where {T<:Real} where {U<:Integer}
+    if src == dst
+        return Int64[]
```
gdalle, line 226: Why `Int64` instead of `Int`?

src/shortestpaths/dijkstra.jl
```diff
@@ -169,3 +169,115 @@ function dijkstra_shortest_paths(
...
+    end
+    # keep weight of the best seen path and the midpoint vertex
+    mu, mid_v = typemax(T), -1
```
gdalle, line 229: Be sure that you never add to `mu`

src/shortestpaths/dijkstra.jl
```diff
@@ -169,3 +169,115 @@ function dijkstra_shortest_paths(
...
+
+    dists_b[dst] = zero(T)
+    visited_b[dst] = true
+    Qb[dst] = zero(T)
+
+    while !isempty(Qf) && !isempty(Qb)
+        uf, ub = dequeue!(Qf), dequeue!(Qb)
+
+        for v in outneighbors(g, uf)
+            relax(uf, v, distmx, dists_f, parents_f, visited_f, Qf)
```
gdalle, line 254: This relax function clarifies things, would it be hard to add it to the original `dijkstra` too?

src/shortestpaths/dijkstra.jl
```diff
@@ -169,3 +169,115 @@ function dijkstra_shortest_paths(
...
+                mid_v = v
+            end
+        end
+        if dists_f[uf]+dists_b[ub] >= mu
+            break
+        end
+    end
+    if mid_v == -1
+        # no path exists between source and destination
+        return Int64[]
```
gdalle, line 276: Same question as before

src/shortestpaths/dijkstra.jl
```diff
@@ -169,3 +169,115 @@ function dijkstra_shortest_paths(
...
+    end
+    ds_f = DijkstraState{T,U}(parents_f, dists_f, preds_f, zeros(nvg), Vector{U}())
+    ds_b = DijkstraState{T,U}(parents_b, dists_b, preds_b, zeros(nvg), Vector{U}())
+    path = vcat(enumerate_paths(ds_f, mid_v), reverse(enumerate_paths(ds_b, mid_v)[1:end-1]))
```
gdalle, line 280: What if there are several paths?

gdalle (approved):
It's a good contribution, thanks! The tests will probably fail because of formatting alone, so be sure to install JuliaFormatter.jl in your global environment (`@v1.9`) and then to run `using JuliaFormatter; format(".")` from the root of your fork of Graphs.jl

gdalle (changes requested):
The additional dependencies are not acceptable, and it needs more tests, but it's a solid start!

Project.toml
Dolgalad: Sorry, I think some undesired dependencies have been added because of my dubious workflow.

test/shortestpaths/dijkstra.jl
Dolgalad: Will do

src/shortestpaths/dijkstra.jl
Dolgalad: `mu` should be strictly decreasing, it is only modified through calls to `min`

src/shortestpaths/dijkstra.jl
Dolgalad: I was hoping you would notice =p I'll make the changes

src/shortestpaths/dijkstra.jl
Dolgalad: I'm not sure how to deal with such situations, as far as I can tell multiple paths with the same cost can be detected if the condition `mu == ls + lt` is reached. I'll look into it

Project.toml
Dolgalad: Dependency has been removed

src/shortestpaths/dijkstra.jl
Dolgalad: Dealt with

Project.toml
gdalle: The [documentation of Pkg.jl](https://pkgdocs.julialang.org/v1/getting-started/) is a good place to start improving said workflow ;)

src/shortestpaths/dijkstra.jl
Dolgalad: I have a new version of the `relax` function  that should work for both the `dijkstra_shortest_paths` and `bidijkstra_shortest_path` but it is a bit clunky. I am wondering; why the need for the `parents` and `preds` structures ? wouldn't the list of predecessors be enough ? 
A cosmetic note: should we rename the function the `bidijkstra_shortest_paths` if it should be able to deal with multiple paths ?

src/shortestpaths/dijkstra.jl
Dolgalad: Given that `bidijkstra_shortest_path` only returns `src`-`dst`-paths it doesn't make much sense to keep a full list of predecessors, does it ? I am not sure what the most elegant way to deal with these situations, especially when trying to share code with `dijkstra_shortest_paths`.
I've also noticed that there doesn't seem to be a straightforward way to retrieve multiple paths for a single `(src,dst)` pair. As far a I can tell the `enumerate_paths` routine returns a single path per destination. Should this be addressed at some point ?

test/shortestpaths/dijkstra.jl
Dolgalad: Not sure how to deal with the following problem. Some tests fail when calling `relax` when `u` and `v` are not of the same type. This seams to occur when calling `dijkstra_shortest_paths(g, 1)` when `g`'s node type is not `Int64`. 
I've chosen not to specify the type of arguments `u,v` in `relax`. 

On another note: I've noticed that the `a_star` function returns paths as list of edges rather then nodes. Is this a deliberate choice ?

Dolgalad: Added tests in `test/shortestpaths/bidijkstra.jl`

Dolgalad:
Not sure why tests are failing

gdalle:
I think your modification to Dijkstra somehow broke a test in the centrality part of the package.
My two propositions:
- revert your modifications to only implement bidirectional Dijkstra without changing the original
- investigate why you broke centrality

The error comes from the following test:
```julia
# test #1405 / #1406
g = GenericGraph(grid([50, 50]))
z = betweenness_centrality(g; normalize=false)
@test maximum(z) < nv(g) * (nv(g) - 1)
```
and it looks like this:
```
Betweenness: Test Failed at /home/runner/work/Graphs.jl/Graphs.jl/test/centrality/betweenness.jl:95
  Expression: maximum(z) < nv(g) * (nv(g) - 1)
   Evaluated: 3.193572232997055e10 < 6247500
```
It is related to the following issue in the old LightGraphs.jl package
https://github.com/sbromberger/LightGraphs.jl/issues/1405

simonschoelly (changes requested):
There seems to be some different behavior for the `pathcount` when using `dijkstra_shortest_path`. Might be some kind of overflow issue, as I only saw it on large graphs so far.

before:
```julia
julia> dijkstra_shortest_paths(grid([50, 50]), 1).pathcounts[50*50]
2.5477612258980867e28
```

after:
```julia
julia> dijkstra_shortest_paths(grid([50, 50]), 1).pathcounts[50*50]
8.581105107791177e17
```

I think the right solution here should be `(2 * 49 \choose 49)`, i.e
```julia
julia> binomial(BigInt(2*49), 49)
25477612258980856902730428600
```
so apart from some rounding errors, the old solution looks much more correct to me.

src/shortestpaths/dijkstra.jl
```diff
@@ -83,7 +83,7 @@ function dijkstra_shortest_paths(
     parents = zeros(U, nvg)
     visited = zeros(Bool, nvg)
 
-    pathcounts = zeros(nvg)
```
simonschoelly, line 86: We probably should not change this, it might overflow.

simonschoelly:
It also look to me, as if those changes might make the algorithm perform slower than before, but I only checked this single case:
before:
```julia
julia> @btime dijkstra_shortest_paths($(grid([50, 50])), 1);
  548.740 μs (81 allocations: 165.11 KiB)
```
after
```julia
julia> @btime dijkstra_shortest_paths($(grid([50, 50])), 1);
  585.437 μs (83 allocations: 184.72 KiB)
```

gdalle:
@Dolgalad is this ready for another round of review? sorry for the delay

Krastanov:
this regrettably ended up bitrotting in the last few years. @Dolgalad , if you have the energy to revive it, do not hesitate to push for it (e.g. starting with the merge conflicts)! For the moment I will convert this to a draft in order to keep my review queue a bit more organized, but do not hesitate to mark it back as not a draft.

Issue #224, opened 2023-02-09
Move testing to `GenericGraph`


gdalle:
See #133

simonschoelly:
I don't think this should be tagge as a v2.0 issue, there is no reason not to use `GenericGraph` right now.

gdalle:
Fair enough

gdalle:
@simonschoelly is there any more we can do on this end? I would keep the issue open though

PR #270, opened 2023-07-02
Use GenericGraph for testing biconnectivity algorithms

simonschoelly:
This PR changes the biconnectivity algorithms so that they test with `GenericGraph` instead of `SimpleGraph`.

With this, I discovered an error, where we incorrectly assumed that the result of `outneighbors(g, v)` is indexable. To work around this without , I added a function `collect_if_not_vector` that converts the result of `outneighbors` to a vector unless it is already one. In the future we might think, if we either should adjust these algorithms, or instead require the neighbor functions to always return an indexable collection.

### DIFF
--- a/src/biconnectivity/articulation.jl
+++ b/src/biconnectivity/articulation.jl
@@ -35,17 +35,25 @@ function articulation end
         cnt::T = one(T)
         first_time = true
 
+        # TODO the algorithm currently relies on the assumption that
+        # outneighbors(g, v) is indexable. This assumption might not be true
+        # in general, so in case that outneighbors does not produce a vector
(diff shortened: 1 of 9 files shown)


gdalle:
@etiennedeg interesting insight here

gdalle (approved):
Good eye! This looks fine to me so I'm merging, but @simonschoelly maybe we should put a progress tracker in issue #224 so that we know where generic graphs are used in testing already?

simonschoelly:
> Good eye! This looks fine to me so I'm merging, but @simonschoelly maybe we should put a progress tracker in issue #224 so that we know where generic graphs are used in testing already?

Right, I will do that after my next PR

PR #271, opened 2023-07-02
Add ignore_missing_comparison=true for JET tests

simonschoelly:
For testing with JET `>= 0.8.4` we need to specify `ignore_missing_comparison=true` in order for the type inference test with the `==` operator to correctly work.

See: https://github.com/aviatesk/JET.jl/releases/tag/v0.8.4


gdalle (approved):
LGTM, good catch

PR #272, opened 2023-07-02
Test centrality with generic graph

simonschoelly:
This PR changes the centrality algorithms so that they test with GenericGraph instead of SimpleGraph.

Some tests did fail, because we passed the output of `vertices(g)` to `degree` - but previously that function would not only accept an `AbstractVector` and not a generic collection of vertices, so I changed `indegree`, `outdegree` and `degree`, so that they can take an arbitrary collection of vertices. Maybe in the future, we could also just demand, that `vertices(g)` returns an `AbstractVector`.

Furthermore there was an issue with `betweenness_centrality` where we tried to index into `vertices(g)` - this also would have been a non-issue, if `vertices(g)` was an `AbstractVector`.

Lastly, I had to change the input in the `sample` method, that could also not deal with the output of `vertices(g)`.

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -44,6 +44,10 @@ julia> add_vertices!(g, 2)
 """
 add_vertices!(g::AbstractGraph, n::Integer) = sum([add_vertex!(g) for i in 1:n])
 
+# TODO the behaviour of indegree (and as well outdegree and degree) is very
+# badly documented for the case indegree(g, vs) where vs is not a single vertex
+# but rather a collection of vertices
+
 """
     indegree(g[, v])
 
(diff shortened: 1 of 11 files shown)


gdalle:
Partial solution for #224

src/core.jl
```diff
@@ -44,6 +44,10 @@ julia> add_vertices!(g, 2)
 """
 add_vertices!(g::AbstractGraph, n::Integer) = sum([add_vertex!(g) for i in 1:n])
 
+# TODO the behaviour of indegree (and as well outdegree and degree) is very
```
gdalle, line 47: See #279

gdalle (approved):
Good catches, once again!

PR #273, opened 2023-07-02
Test community with generic graph

simonschoelly:
This PR makes the tests for the community algorithm use `GenericGraph`

There were only some small changes necessary to `src/community/clustering.jl` to make the tests work.


gdalle:
Partial solution to #224

gdalle (approved):

PR #274, opened 2023-07-02
Test cycles with generic graph

simonschoelly:
This PR makes the tests for the cycles algorithm use `GenericGraph`

As this PR is based on #270 , it should not be merged/reviewed until that other PR is.

Currently the algorithms in `src/cycles/incremental.jl` and `src/cycles/johnson.jl` do not work with generic graphs - this is, because they relay on modifying graph functions. At least for johnson this might be fixable but for incremental it seems to be necessary to have modifiable graphs.

In addition, this PR changes the function `cycle_basis` so that it accepts an arbitrary `AbstractGraph` instead of just an `AbstractSimpleGraph`.


gdalle:
Partial solution to #224

gdalle (approved):
LGTM, I opened issues to track the missing pieces

PR #275, opened 2023-07-02
Use GenericGraph for testing shortestpaths algorithms

simonschoelly:
This PR makes the tests for the shortestpath algorithm use `GenericGraph`

The function in `src/shortestpaths/yen.jl` does currently not work with GenericGraph yet. Also `a_star` only kind of works with `GenericGraph` as this function by default takes `edgetype(g)` as an argument and then tries to use that type as an edge constructor. This does work for `GenericEdge` though. So I doubled some tests - some are run with `SimpleGraph` and `SimpleDiGraph` and some are run with `GenericGraph` and `GenericDiGraph`, but the default edgetype is replaced by `SimpleEdge`.

In addition, this PR fixes a bug where the `has_negative_edge_cycle` might not only fail on a generic graph, but return an incorrect result, as the `MethodError` exception was silently caught and ignored.

### DIFF
--- a/src/shortestpaths/bellman-ford.jl
+++ b/src/shortestpaths/bellman-ford.jl
@@ -48,7 +48,8 @@ function bellman_ford_shortest_paths(
     for i in vertices(graph)
         no_changes = true
         new_active .= false
-        for u in vertices(graph)[active]
+        for u in vertices(graph)
+            active[u] || continue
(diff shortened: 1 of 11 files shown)


gdalle (approved):
As usual, nice work! Opened an issue to track Yen

PR #276, opened 2023-07-02
Use GenericGraph for testing spanningtrees algorithms

simonschoelly:
This PR makes the tests for the spanningtrees algorithms use `GenericGraph`


gdalle (approved):
OK by me, I think the edge comparisons will be a native feature in GraphsBase

PR #277, opened 2023-07-04
Added edge-betweenness.jl to centralities

jwassmer:
I have added a new file `edge-betweenness.jl` to `src/centrality/`. Here I include a function to compute the edge betweenness of a graph (directed and weighted). In theory I could also add a version for MultiGraphs, but these are not in the base version of Graphs atm. My code is based on the version from networkX. 

This is my first contribution, and I hope I have followed all the guidelines correctly.

Jonas

### DIFF
--- a/src/centrality/edge-betweenness.jl
+++ b/src/centrality/edge-betweenness.jl
@@ -1,0 +1,102 @@
+"""
+    edge_betweenness_centrality(g, k)
+
+Compute the [edge betweenness centrality](https://en.wikipedia.org/wiki/Centrality#Betweenness_centrality) of an edge `e`.
+It is defined as the sum of the fraction of all-pairs shortest paths that pass through `e`
+``
+bc(e) =  \\sum_{s, t \\in V}
+\\frac{\\sigma_{st}(e)}{\\sigma_{st}}
+``.
+
+where `V`, is the set of nodes, \\frac{\\sigma_{st}} is the number of shortest-paths, and \\frac{\\sigma_{st}(e)} is the number of those paths passing through edge.
+
+### Optional Arguments
+- `normalize=true`: If true, normalize the betweenness values by the
+total number of possible distinct paths between all pairs in the graphs.
+For an undirected graph, this number is ``\\frac{(|V|-1)(|V|-2)}{2}``
+and for a directed graph, ``{(|V|-1)(|V|-2)}``.
+
+
+### References
+- Brandes 2001 & Brandes 2008
+
+# Examples
+```jldoctest
+julia> using Graphs
+
+julia> edge_betweenness_centrality(star_graph(5))
+Dict{Graphs.SimpleGraphs.SimpleEdge{Int64}, Float64} with 4 entries:
+  Edge 1 => 2 => 0.4
+  Edge 1 => 3 => 0.4
+  Edge 1 => 4 => 0.4
+  Edge 1 => 5 => 0.4
+
+julia> edge_betweenness_centrality(path_digraph(6))
+Dict{Graphs.SimpleGraphs.SimpleEdge{Int64}, Float64} with 5 entries:
+  Edge 4 => 5 => 0.266667
+  Edge 1 => 2 => 0.166667
+  Edge 3 => 4 => 0.3
+  Edge 5 => 6 => 0.166667
+  Edge 2 => 3 => 0.266667
+```
+"""
+
+function edge_betweenness_centrality(
+    g::AbstractGraph, vs=vertices(g), distmx::AbstractMatrix=weights(g); normalize=true
+)
+    edge_betweenness = Dict(edges(g) .=> 0.0)
+    for o in vs
+        state = dijkstra_shortest_paths(g, o, distmx; allpaths=true, trackvertices=true)
+        _accumulate_edges!(edge_betweenness, state)
+    end
+    _rescale_e!(edge_betweenness, nv(g), normalize, is_directed(g))
+
+    return edge_betweenness
+end
+
+function _accumulate_edges!(edge_betweenness::AbstractDict, state::Graphs.AbstractPathState)
+    σ = state.pathcounts
+    pred = state.predecessors
+    seen = state.closest_vertices
+    δ = Dict(seen .=> 0.0)
+
+    while length(seen) > 0
+        w = pop!(seen)
+
+        coeff = (1.0 + δ[w]) / σ[w]
+        for v in pred[w]
+            c = σ[v] * coeff
+            if Edge(v, w) ∉ edge_betweenness.keys
+                edge_betweenness[Edge(w, v)] += c
+            else
+                edge_betweenness[Edge(v, w)] += c
+                δ[v] += c
+            end
+        end
+    end
+    return nothing
+end
+
+function _rescale_e!(
+    edge_betweenness::AbstractDict, n::Integer, normalize::Bool, directed::Bool=false
+)
+    if normalize
+        if n <= 1
+            scale = nothing  # no normalization b=0 for all nodes
+        else
+            scale = 1 / (n * (n - 1))
+        end
+    else  # rescale by 2 for undirected graphs
+        if !directed
+            scale = 0.5
+        else
+            scale = nothing
+        end
+    end
+    if scale !== nothing
+        for (k, v) in edge_betweenness
+            edge_betweenness[k] *= scale
+        end
+    end
+    return nothing
+end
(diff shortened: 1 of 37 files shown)


simonschoelly:
Hi Jonas, thanks for your contribution.

Code reviews here take quite a while at the moment, especially when the reviewer tries to understand the algorithm and we are also a bit understaffed, just to warn you :)

That being said, the first thing that you definitely should also implement are some tests - that is why the codecov bot is complaining so loudly. If you want to use them, we recently introduced some graph types called `GenericGraph` and `GenericDiGraph` - with these graph types we should be able to catch some errors that we might not see if we use `SimpleGraph` or `SimpleDiGraph` - this has not been changed in a lot of the existing tests yet, but here is an open PR for the other centrality algorithms if you want to look how it is done there: https://github.com/JuliaGraphs/Graphs.jl/pull/272

src/centrality/edge-betweenness.jl
```diff
@@ -0,0 +1,102 @@
...
+  Edge 4 => 5 => 0.266667
+  Edge 1 => 2 => 0.166667
+  Edge 3 => 4 => 0.3
+  Edge 5 => 6 => 0.166667
+  Edge 2 => 3 => 0.266667
+```
+"""
+
+function edge_betweenness_centrality(
+    g::AbstractGraph, vs=vertices(g), distmx::AbstractMatrix=weights(g); normalize=true
```
simonschoelly, line 45:
```suggestion
    g::AbstractGraph, vs=vertices(g), distmx::AbstractMatrix=weights(g); normalize::Bool=true
```

simonschoelly: Does it make sense to restrict the amount of source vertices here, or should we actually always consider all vertices of the graph? Especially as we do not seem to restrict the target vertices.

src/centrality/edge-betweenness.jl
```diff
@@ -0,0 +1,102 @@
...
+)
+    edge_betweenness = Dict(edges(g) .=> 0.0)
```
simonschoelly, line 47: I think in other places where we calculate some quantity for all edges, we usually return a `|V| x |V|` matrix (`SparseMatrixCSC`) instead of a dictionary. Maybe that would also make sense here? 

This will otherwise probably also fail for arbitrary `AbstractGraph`, as `AbstractEdge` might not be usable as a dictionary key (unfortunately Julia does not warn about that).

src/centrality/edge-betweenness.jl
```diff
@@ -0,0 +1,102 @@
...
+    for o in vs
+        state = dijkstra_shortest_paths(g, o, distmx; allpaths=true, trackvertices=true)
+        _accumulate_edges!(edge_betweenness, state)
+    end
```
simonschoelly, line 51: It might be better to use names like `u`, `v` , `w`  or `source`  as name for a vertex, as this is kind of a convention, so might make it easier for other people to understand the code.

src/centrality/edge-betweenness.jl
```diff
@@ -0,0 +1,102 @@
...
+    seen = state.closest_vertices
+    δ = Dict(seen .=> 0.0)
+
+    while length(seen) > 0
+        w = pop!(seen)
+
+        coeff = (1.0 + δ[w]) / σ[w]
+        for v in pred[w]
+            c = σ[v] * coeff
+            if Edge(v, w) ∉ edge_betweenness.keys
```
simonschoelly, line 69:
```suggestion
            if Edge(v, w) ∉ keys(edge_betweenness)
```
`edge_betweenness.keys` is some internal datastructure of that dictionary, we should not access it.
For example this is probably not what you would except:
```julia
julia> Dict("x" => 2).keys
16-element Vector{String}:
 #undef
 #undef
 #undef
 #undef
 #undef
 #undef
 #undef
 #undef
    "x"
 #undef
 #undef
 #undef
 #undef
 #undef
 #undef
 #undef
```

src/centrality/edge-betweenness.jl
```diff
@@ -0,0 +1,102 @@
...
+
+julia> edge_betweenness_centrality(path_digraph(6))
+Dict{Graphs.SimpleGraphs.SimpleEdge{Int64}, Float64} with 5 entries:
+  Edge 4 => 5 => 0.266667
+  Edge 1 => 2 => 0.166667
+  Edge 3 => 4 => 0.3
+  Edge 5 => 6 => 0.166667
+  Edge 2 => 3 => 0.266667
+```
+"""
```
simonschoelly, line 42: We probably should also have an example with `normalize=false`  and maybe (but that might be a bit too much text) one with a different `distmx`.

src/centrality/edge-betweenness.jl
jwassmer: Good point. Now it returns a SparseMatrix

src/centrality/edge-betweenness.jl
jwassmer: ✅

src/centrality/edge-betweenness.jl
```diff
@@ -0,0 +1,118 @@
...
+    return edge_betweenness
+end
+
+function edge_betweenness_centrality(
+    g::AbstractGraph,
+    k::Integer,
+    distmx::AbstractMatrix=weights(g);
+    normalize=true,
+    rng::Union{Nothing,AbstractRNG}=nothing,
+    seed::Union{Nothing,Integer}=nothing,
```
simonschoelly, line 69: We should not use the `seed` argument for randomized algorithms anymore, the only reason we still have it in some old algorithms is, because we initialy only had a seed, so we would not break the code of people using this library.

But for new algorithms we don't have this concern, so we can safely remote it.

src/centrality/edge-betweenness.jl
```diff
@@ -0,0 +1,118 @@
...
+            g, source, distmx; allpaths=true, trackvertices=true
+        )
+        _accumulate_edges!(edge_betweenness, state)
+    end
+    _rescale_e!(edge_betweenness, nv(g), normalize, is_directed(g), k)
+
+    return edge_betweenness
+end
+
+function edge_betweenness_centrality(
```
simonschoelly, line 63: We also need to document and write tests for this function.

simonschoelly: You could also write a combined documentation for both functions, in that case you would probably do something like
```julia
"""
    edge_betweenness_centrality(g, [,vertices, distmx]; [normalize, rng])
   
[... some more explanation here ...]
"""
functione dge_betweenness_centrality end
```
speaking of, maybe we should also make `vertices` a keyword argument? Although we also do not seem to to so for other algorithms.

src/centrality/edge-betweenness.jl
jwassmer: ✅

src/centrality/edge-betweenness.jl
jwassmer: I wrote tests and updated the documentary. Furthermore I made `vertices` a keyword argument, as that makes more sense. We should maybe change it for the node betweenness as well.

jwassmer: Also I am really unhappy with `vertices` being called `vs`. Maybe we should update the name as well

src/centrality/edge-betweenness.jl
```diff
@@ -0,0 +1,145 @@
...
+)
+    σ = state.pathcounts
+    pred = state.predecessors
+    seen = state.closest_vertices
+    δ = Dict(seen .=> 0.0)
+
+    while length(seen) > 0
+        w = pop!(seen)
+
+        coeff = (1.0 + δ[w]) / σ[w]
```
gdalle, line 117: can you explain the dynamics of `coeff` and `δ` through this loop?

src/centrality/edge-betweenness.jl
```diff
@@ -0,0 +1,145 @@
...
+is calculated as ``1 / (|V|(|V|-1))``.
+`vs=vertices(g)`: A subset of nodes in the graph g for which the edge betweenness centrality is to be estimated. By including more nodes in this subset, 
+you can achieve a better estimate of the edge betweenness centrality.
+`distmx=weights(g)`: The weights of the edges in the graph g represented as a matrix. This argument allows you to specify custom weights for the edges. 
+The weights can be used to influence the calculation of betweenness centrality, giving higher importance to certain edges over others.
+`rng`: A random number generator used for selecting k vertices. This argument allows you to provide a custom random number generator that will be used for the vertex selection process. 
+
+
+### References
+- Brandes 2001 & Brandes 2008
```
gdalle, line 29: can you give more details to help us check the algorithm?

test/centrality/edge-betweenness.jl
```diff
@@ -0,0 +1,81 @@
+
+@testset "Edge Betweenness" begin
```
gdalle, line 2: how did you pick your test cases?

gdalle (changes requested):
Thank you for the contribution! I'll try to review it, and have asked a few questions to understand the procedure a bit better

PR #278, opened 2023-07-04
Tests for various files with generic graph

simonschoelly:
This PR changes the tests to use generic graphs for various files that were not grouped together like others.

There where no larger changes necessary to make the tests work, but a small change was necessary for `bfs_parents` and some type annotations in `operators.jl` had to be changed so that they would accept abstract graphs.


gdalle:
Is this mergeable @simonschoelly ?

gdalle (approved):

PR #284, opened 2023-07-05
fix is_cyclic and topological_sort

etiennedeg:
Fixes a bug introduced by https://github.com/JuliaGraphs/Graphs.jl/pull/266

We should not mark a vertex visited when we store the neighbors on the stack.


gdalle:
Is the meaning of the numbers (the one I had you add in a comment) still the same?

etiennedeg:
No, I forgot to update the comment for `is_cyclic`. When we stack the neighbors of a node, we should not mark these already as visited, so we need one less color.

gdalle (approved):
LGTM

PR #289, opened 2023-07-27
Create feature_request version of issues

Tortar:
Noticed that there is no Feature request version of issues, what about adding one?


simonschoelly:
The reason why we have a template for bugs is, that in the past people reported bugs but forgot to add the necessary information to reproduce, I don't mind having a template for feature requests - as long as there are not a lot of unnecessary fields to will shy people away from creating such feature requests.

Tortar:
I think that just having the button is probably fine then :-)

gdalle (approved):

PR #293, opened 2023-08-25
Avoid the `zeros(nthreads())[threadid()]` buffering pattern

Drvi:
As recommended in this blogpost https://julialang.org/blog/2023/07/PSA-dont-use-threadid/

### DIFF
--- a/src/Parallel/centrality/betweenness.jl
+++ b/src/Parallel/centrality/betweenness.jl
@@ -98,25 +98,24 @@ function threaded_betweenness_centrality(
     k = length(vs)
     isdir = is_directed(g)
 
-    local_betweenness = [zeros(n_v) for i in 1:nthreads()]
     vs_active = findall((x) -> degree(g, x) > 0, vs) # 0 might be 1?
+    d, r = divrem(length(vs_active), Threads.nthreads())
+    ntasks = d == 0 ? r : Threads.nthreads()
+    local_betweenness = [zeros(n_v) for i in 1:ntasks]
 
-    Base.Threads.@threads for s in vs_active
-        state = Graphs.dijkstra_shortest_paths(
-            g, s, distmx; allpaths=true, trackvertices=true
-        )
-        if endpoints
-            Graphs._accumulate_endpoints!(
-                local_betweenness[Base.Threads.threadid()], state, g, s
+    @sync for t in 1:ntasks
+        Threads.@spawn for s in @view(vs_active[(d*(t-1)+1):(d*t + (t <= r))])
+            state = Graphs.dijkstra_shortest_paths(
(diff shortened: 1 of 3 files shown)


etiennedeg:
This is known issue: https://github.com/JuliaGraphs/Graphs.jl/issues/10
We already moved the algorithms causing failing tests in Experimental, but maybe all these algorithms should be in Experimental, as we can't ensure the correctness of these algorithms.

gdalle:
Where do you think there might be other such algorithms?

Drvi:
Hi folks and sorry for the delay in my reply. To clarify, this aims to fix an issue we might get into on 1.9.3 where multithreading behavior changed slightly in that `Threads.nthreads()` gives you the number of threads in the default threadpool, but the default thread ids are enumerated sequentially _after_ the interactive threads, so with `julia193 -t7,2`, `nthreads() == 7` and thread ids in the default thread pool are `3, 4, ..., 8, 9`, e.g:

```julia
julia> buf = zeros(Threads.nthreads())
7-element Vector{Float64}:
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0

julia> Threads.@threads for i in 1:Threads.nthreads()
           buf[Threads.threadid()] = Threads.threadid()
       end
ERROR: TaskFailedException

    nested task error: BoundsError: attempt to access 7-element Vector{Float64} at index [9]
    Stacktrace:
     [1] setindex!(A::Vector{Float64}, x::Int64, i1::Int64)
       @ Base ./array.jl:969
     [2] macro expansion
       @ ./REPL[2]:2 [inlined]
     [3] (::var"#26#threadsfor_fun#4"{var"#26#threadsfor_fun#3#5"{UnitRange{Int64}}})(tid::Int64; onethread::Bool)
       @ Main ./threadingconstructs.jl:200
     [4] #26#threadsfor_fun
       @ ./threadingconstructs.jl:167 [inlined]
     [5] (::Base.Threads.var"#1#2"{var"#26#threadsfor_fun#4"{var"#26#threadsfor_fun#3#5"{UnitRange{Int64}}}, Int64})()
       @ Base.Threads ./threadingconstructs.jl:139

...and 1 more exception.
```
To avoid that, instead of `Threads.threadid()` we are using _task_ ids, simple integer values we increment for each task we `@spawn`.

I don't know if this PR fixes any other, older known bugs or issues though.

gdalle:
Thanks for the PR! Do you know why tests are failing?

Drvi:
@gdalle I think I fixed the issue :+1:

gdalle:
will review soon, thanks!

gdalle (approved):
Looks good to me

PR #308, opened 2023-09-21
Remove non-default GenericGraph constructor

simonschoelly:
This PR removes the `GenericGraph(elist)` constructor.

The reason why we do not want to have such a constructor, is that we use `GenericGraph` to verify that an interface that uses an `AbstractGraph` only uses the methods as defined in the informal interface.

In the future, we might add convenience functions such as `generic_graph(elist)`  that could help us de-clutter our tests.


gdalle (approved):

gdalle:
wanna merge it?

PR #309, opened 2023-09-21
Split nightly testing from rest of CI

gdalle:
- Allow the CI jobs running on nightly to error without making the whole workflow fail. For more details, see
  - https://discourse.julialang.org/t/ignoring-nightly-failure-for-ci-badge/98028
  - https://github.com/actions/runner/issues/2347

- Remove an outdated `codecov.yml` at the root of the repo


simonschoelly (approved):
I guess we have to merge this to to actually see the changes to the github badge.

gdalle:
Yes, I'm a bit skeptical because there is a red X next to the commit, but if you look in the "actions" tab of the repo there is a green check. Let's hope for the best!

Issue #310, opened 2023-09-28
Make benchmarks operational and clean them up


gdalle:
- [ ] See what they're doing
- [ ] Make them compatible with PkgBenchmark.jl
- [ ] Run them on CI?

Links:
- https://github.com/JuliaCI/PkgBenchmark.jl
- https://github.com/tkf/BenchmarkCI.jl

filchristou:
That's me. xD

SuperGrobi:
Just for inspiration:
- Makie is running [this](https://github.com/MakieOrg/Makie.jl/blob/master/.github/workflows/compilation-benchmark.yaml) CI which runs [this](https://github.com/MakieOrg/Makie.jl/blob/master/metrics/ttfp/run-benchmark.jl) on every pr.
- CUDA.jl uses buildkite and [codespeed](https://github.com/tobami/codespeed/) to create these very nice [dashboards](https://speed.juliagpu.org/timeline/#/?exe=9,11&env=1&base=none&ben=grid&revs=50) of performance development.

gdalle:
AAAAAH I JUST REALIZED THAT HENRIK WOLF IS SUPERGROBI

PR #315, opened 2023-11-16
First try on benchmark CI

filchristou:
https://github.com/JuliaGraphs/Graphs.jl/issues/310

# Options
As mentioned in the discourse thread https://discourse.julialang.org/t/recommendation-for-ci-benchmark-to-catch-regression/104698 the available options to have CI benchmarks are
- [PkgBenchmark.jl](https://github.com/JuliaCI/PkgBenchmark.jl) : a thin layer on top of BenchmarkTools.jl
- [BenchmarkCI.jl](https://github.com/JuliaCI/PkgBenchmark.jl) : an opinionated CI workflow using PkgBenchmark.jl
- [benchmark-action/github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark) : a CI machinery on top of BenchmarkTools.jl 
- [AirspeedVelocity.jl](https://github.com/MilesCranmer/AirspeedVelocity.jl) : inspired by [python's asv](https://asv.readthedocs.io/en/stable/)

# Implementation
In this approach PkgBenchmark.jl and BenchmarkCI.jl are used.

## Benchmarks
Regarding the benchmarks in `./benchmark/*` directory I did nothing but translating them to a compatible formulation.
This means that the `SUITE:BenchmarkGroup` must be defined in `./benchmark/benchmark.jl`
All options above need the same form (with minor changes for `benchmark-action/github-action-benchmark`).
This means that we can continue authoring more benchmarks with the security that the code is reusable even if in the CI approach changes in the future.

## CI 
This is described in the single file `/.github/workflows/benchmark.yaml`
There are 2 paths: if the event is a `pull_request` or if it is a `push`. 

If the event is a `pull_request` a comparison is made (using `BenchmarkCI.judge`) between the base and the target refs.
This will also post a summary in the PR as a comment.
The base and target benchmarks are uploaded as artifacts in the workflow.

If the event is a `push` (or a merge) the benchmarks from the commit are run (using `PkgBenchmark.benchmarkpkg`).
A `median` estimation of the benchmarks is uploaded as an artifact in the workflow.

# Further discussion

## Strategy
Benchmarks are just running on `ubuntu-latest` and `julia-v1`.
I don't see any reason to populate the workflow with more data.

## Artifacts uploads
Artifacts uploads remain available [only for 90 days](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#artifact-and-log-retention-policy).
After that, the only thing that will stay are the comment-posts in the PRs.
If we want to store all benchmarks we can discuss on some alternatives.
- uploading them in another github repo
- uploading them as part of the current repo (the files need to be small (that's the thinking behind the median estimations)
- uploading in another cloud service

I would like that to be part of the repo, since it's way easier to manage, but it's also completely useless for the users to get.

## Permissions and approval
Right now the PR benchmark is run everytime a new update of the branch is pushed.
That might not be necessary. 
I think it would be better to have some kind of a manual trigger to start the machinery when other more obvious issues are already resolved. 
Also, it is insecure to allow everyone to run or modify the workflow, but I think that's already solved in the organization level.

## Single and multithread 
Right now both single and multithread benchmarking code is given 2 threads. 
We can seperate that.

# Example
See how a PR can look
https://github.com/filchristou/Graphs.jl/pull/3
which triggers  the workflow 
https://github.com/filchristou/Graphs.jl/actions/runs/6889672830
After a PR is merged, a benchmark will run again
https://github.com/filchristou/Graphs.jl/actions/runs/6889854479

Overall I cannot promise it will work perfectly directly, since I was testing stuff in the fork and not on the main or protected branch. Also, I am not very familiar with the Graphs.jl way of things, so I guess we can tune several preferences in the future.


filchristou:
For experimentation reasons we can make a `benchx` branch here actually.

benchmark/benchmarks.jl
```diff
@@ -1,18 +1,37 @@
...
+const GRAPHS = Dict{String,Graph}(
     "complete100" => complete_graph(100),
     "tutte" => smallgraph(:tutte),
     "path500" => path_graph(500),
 )
 
-suite = BenchmarkGroup()
-include("core.jl")
+serialbenchmarks = [
+    "serial/core.jl",
```
gdalle, line 15: use `joinpath`?

benchmark/tune.json
```diff
@@ -0,0 +1 @@
...,"seconds":5.0,"overhead":0.0,"memory_tolerance":0.01}]},"tags":[]}]},"tags":[]}],"serial":["BenchmarkGroup",{"data":{"egonet":["BenchmarkGroup",{"data":{"vertexfunction":["Parameters",{"gctrial":true,"time_tolerance":0.05,"samples":10000,"evals":1,"gcsample":false,"seconds":5.0,"overhead":0.0,"memory_tolerance":0.01}],"twohop":["Parameters",{"gctrial":true,"time_tolerance":0.05,"samples":10000,"evals":1,"gcsample":false,"seconds":5.0,"overhead":0.0,"memory_tolerance":0.01}]},"tags":[]}]},"tags":[]}],"edges":["BenchmarkGroup",{"data":{"fillp":["Parameters",{"gctrial":true,"time_tolerance":0.05,"samples":10000,"evals":1,"gcsample":false,"seconds":5.0,"overhead":0.0,"memory_tolerance":0.01}],"fille":["Parameters",{"gctrial":true,"time_tolerance":0.05,"samples":10000,"evals":1,"gcsample":false,"seconds":5.0,"overhead":0.0,"memory_tolerance":0.01}],"tsume":["Parameters",{"gctrial":true,"time_tolerance":0.05,"samples":10000,"evals":1,"gcsample":false,"seconds":5.0,"overhead":0.0,"memory_tolerance":0.01}],"tsump":["Parameters",{"gctrial":true,"time_tolerance":0.05,"samples":10000,"evals":1,"gcsample":false,"seconds":5.0,"overhead":0.0,"memory_tolerance":0.01}]},"tags":[]}]},"tags":[]}]]]
```
gdalle, line 1: do we need this?

benchmark/benchmarks.jl
```diff
@@ -1,18 +1,37 @@
 using BenchmarkTools
 using Graphs
 
-DIGRAPHS = Dict{String,DiGraph}(
+const DIGRAPHS = Dict{String,DiGraph}(
     "complete100" => complete_digraph(100), "path500" => path_digraph(500)
 )
 
-GRAPHS = Dict{String,Graph}(
+const GRAPHS = Dict{String,Graph}(
```
gdalle, line 8: IIUC benchmarking is always done on all of these, and the total time is reported?

gdalle (commented):
Thank you so much for taking this on! The general setup seems okay to me, I didn't review the benchmark contents in detail because that's not the point of this PR: let's first get it working

gdalle:
> Benchmarks are just running on ubuntu-latest and julia-v1.

Agreed

> Artifacts uploads remain available [only for 90 days](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#artifact-and-log-retention-policy).

It would be useful to track the performance of the main branch over longer periods of time. For pull requests however we don't really care. Where are benchmarks for Julia itself stored for example?

> Right now the PR benchmark is run everytime a new update of the branch is pushed.

Can we base this on a GitHub label, like `benchmark`, which repo admins would add to a PR when it's nearly ready?

> For experimentation reasons we can make a benchx branch here actually.

Knock yourself out!

gdalle:
I'd be curious to have @simonschoelly's opinion on the general approach

benchmark/benchmarks.jl
filchristou: ah yeah. with `benchdir = dirname(@__FILE__)` and the `joinpath(benchdir, ...)`. yeah that's the way it should be.

benchmark/tune.json
filchristou: Yes. Why is described on the [Caching Parameters docs of BenchmarkTools.jl](https://juliaci.github.io/BenchmarkTools.jl/stable/manual/#Caching-Parameters). Basically we win in twofold: 
- tuning is slow and it doesn't have to happen everytime when benchmark.
- better consistency as the same evaluations/samples are being done in every benchmark

benchmark/benchmarks.jl
filchristou: Yes. The benchmark results are accumulated across all `Graphs` and `DiGraphs`. I find it too granular to get benchmarks for each one of the Graphs. For me, It's better to add them up.
Now the `const` is not really needed I think, since the `Graph` is always interpolated in the benchmark, but it also doesn't harm.

filchristou:
> Can we base this on a GitHub label, like benchmark, which repo admins would add to a PR when it's nearly ready?

I believe that is technically possible. Maybe the label should be more like `to-benchmark` because with `benchmark` there might be confusion that the PR itself was made for performance/benchmarking reasons ?

filchristou:
The PR is failing because the benchmark baseline (i.e. the master branch) doesn't follow the requirements. I believe this is okey, and I wouldn't want to introduce complicated if statements just for this first PR. From the second on, once the master branch is properly structured, there should be no problem. 
But as I said before we can test it further in a seperate `benchx` branch.

benchmark/tune.json
gdalle: Can we find a way to automate the re-generation of the json every time the suite object is updated? Otherwise people will forget

filchristou:
> Where are benchmarks for Julia itself stored for example?

In a separate github repo
https://github.com/JuliaCI/NanosoldierReports

gdalle:
See https://github.com/JuliaGraphs/JuliaGraphs-meta/issues/12 for the roadmapo

filchristou:
I merged on the branch with the following TODOs pending before pushing to master:
- make sure consecutive PRs benchmarks work as intended
- find a way to trigger PR benchmarks (e.g. labels)
- long store benchmark results upon push to master (github post or other repo)

Issue #316, opened 2023-11-20
Question on neighbours functions returning a reference


jward0:
Question - the Graphs.inneighbors, Graphs.outneighbors and consequently Graphs.neighbors specify that they return a reference to the current graph's internal structures, rather than a copy. Having fallen afoul of this, I was wondering why this is the case - as far as I can tell this would be trivial to have return a copy instead, so I'm assuming it's intentional, but the only use case I can see would be modifying a graph in a way that would presumably be better handled by add_edge!() and rem_edge!()?

Thanks for your time!

gdalle:
The short answer is that copies are expensive because they allocate memory. So you should not modify the output of `neighbors` and co unless you know what you are doing.

jward0:
Oh, of course - that should've occurred to me! Thanks for humouring my question.

simonschoelly:
To add to this, this is indeed, because the adjancency structure used by `SimpleGraph` and `SimpleDiGraph` are indeed already the allocated vectors. In graph algorithms one often does something like
```
for u in vertices(g)
   for v in outneighbors(g, u)
      [...]
   end
end
```
which would be rather slow.

we could create some wrapper structure `FrozenVector` (similar to `frozenset` in Python) that wraps a vector but does not allow to modify the vector. One just has to be a bit careful so that propagating `@inbounds` works, as otherwise we will lose a lot of performance.
There is at least one package that already creates something like that, but we might as well implement it ourselves in order to keep dependencies low: <https://github.com/bkamins/ReadOnlyArrays.jl>

gdalle:
I have no clue, does so much of Graphs.jl perf hinge on `@inbounds``

simonschoelly:
> Unrelated but does so much of Graphs.jl perf hinge on `@inbounds`? I genuinely have no clue

I think in thight loops it can have quite an impact, on one hand we have to check the array bounds for each iteration, and on the other hand, this also prevents the compiler from doing more optimizations such as using simd instructions.

simonschoelly:
I made a PR with a suggestion for such a `FrozenVector`: #317

gdalle:
While I don't disagree with the principle of making neighbors immutable, I wonder whether this is necessary. Users may be a bit surprised by the type returned, so I would rather keep the existing behavior if possible. @jward0 in what kind of situation did this trick you?

jward0:
[...]

PR #317, opened 2023-11-22
Add FrozenVector to wrap neighbors

simonschoelly:
This PR makes the output of `neighbors`, `inneighbors` and `outneighbors` immutable by wrapping the returned `Vector` in an immutable wrapper called `FrozenVector`.

I noticed that the generated machine code for `Base.getindex` is slightly different (one additional instruction) with `FrozenVector` compared to the previously used `Vector` - this does probably not have a significant performance impact, but I should run some benchmarks.

See also this issue: #316

### DIFF
--- a/src/Graphs.jl
+++ b/src/Graphs.jl
@@ -452,6 +452,7 @@ and tutorials are available at the
 Graphs
 include("interface.jl")
 include("utils.jl")
+include("frozenvector.jl")
 include("deprecations.jl")
 include("core.jl")
 
--- a/src/SimpleGraphs/SimpleGraphs.jl
+++ b/src/SimpleGraphs/SimpleGraphs.jl
@@ -15,6 +15,7 @@ import Graphs:
     AbstractGraph,
     AbstractEdge,
     AbstractEdgeIter,
+    FrozenVector,
     src,
     dst,
     edgetype,
@@ -152,8 +153,8 @@ add_edge!(g::AbstractSimpleGraph, x) = add_edge!(g, edgetype(g)(x))
 has_edge(g::AbstractSimpleGraph, x, y) = has_edge(g, edgetype(g)(x, y))
 add_edge!(g::AbstractSimpleGraph, x, y) = add_edge!(g, edgetype(g)(x, y))
 
-inneighbors(g::AbstractSimpleGraph, v::Integer) = badj(g, v)
-outneighbors(g::AbstractSimpleGraph, v::Integer) = fadj(g, v)
+inneighbors(g::AbstractSimpleGraph, v::Integer) = FrozenVector(badj(g, v))
+outneighbors(g::AbstractSimpleGraph, v::Integer) = FrozenVector(fadj(g, v))
 
 function issubset(g::T, h::T) where {T<:AbstractSimpleGraph}
     nv(g) <= nv(h) || return false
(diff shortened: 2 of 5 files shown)


gdalle:
We could wait until #315 is merged to have a proper benchmarking pipeline?

simonschoelly:
> We could wait until #315 is merged to have a proper benchmarking pipeline?

That one has been merged now.

gdalle:
It has been merged to a branch of the main repo (`benchx`) so that @filchristou could experiment with CI setups and actions outside of a fork. It's not yet on `master`, and more importantly even once the pipeline is set up we still won't have a satisfying set of tasks to benchmark.

Besides performance issues, I am personally opposed to this change, because the additional wrapper might be confusing for users (what is this `FrozenVector` thing and what can I do with it?) and incompletely implemented (what if we forget a useful vector operation?). Also I think there is a case for considering such a change breaking

simonschoelly:
> It has been merged to a branch of the main repo (`benchx`) so that @filchristou could experiment with CI setups and actions outside of a fork. It's not yet on `master`, and more importantly even once the pipeline is set up we still won't have a satisfying set of tasks to benchmark.

Right, I had just looked at the `merged` status.

> Besides performance issues, I am personally opposed to this change, because the additional wrapper might be confusing for users (what is this `FrozenVector` thing and what can I do with it?) and incompletely implemented (what if we forget a useful vector operation?). Also I think there is a case for considering such a change breaking

Nowhere do we guarantee that we run an `Vector`. From the documentation:
```
  inneighbors(g, v)

  Return a list of all neighbors connected to vertex v by an incoming edge.

  Implementation Notes
  ––––––––––––––––––––

  Returns a reference to the current graph's internal structures, not a copy. Do not modify result. If the
  graph is modified, the behavior is undefined: the array behind this reference may be modified too, but
  this is not guaranteed.
  ```

And `FrozenVector` is even an `AbstractVector` although we also do not guarantee that by the documentation. So why should one use any other feature of a `Vector` and why should we support that?
On the other hand, we are able to prevent users from making a lot of hard to detect errors when they - for some -reason - try to modify that result without making a copy.

And I don't think performance should be affected seriously.

simonschoelly:
Futhermore, if someone is actually confused what a `FrozenVector` is, they can simply consult the documentation.

gdalle:
Okay, I'm convinced that the benefits outweigh the costs on the correctness side.

Can you maybe run some benchmarks locally, say Dijkstra on a large enough graph, to see if there is a big performance hit from creating the wrappers every time? Maybe even `degree(g)` could suffer a bit.

filchristou:
I will probably merge to master before the community call. After you can use that for benchmarks and talk further.

gdalle:
Awesome, thank you so much! Don't hesitate to ask for a review beforehand though, to follow ColPrac

Krastanov:
this needs a reformatting, but otherwise seems good to go

Krastanov:
Given Guillaume's comment, I will wait for the cleaned up benchmark runner before I merge this (see #490 )

Krastanov:
None of the benchmarks show slowdown (not that the benchmarks are super thorough). I will merge this after the tests pass, given that the review from a few years ago seems to have been positive, even if somewhat muted. On skimming it quickly today, I also do not see issues with this approach, and it seems like a good protection against footguns.

Krastanov:
two doctests fail in an expected fashion, will submit a patch

PR #322, opened 2023-12-15
Switch to Documenter v1

gdalle:
Documenter v1 brings lots of improvements and more safety checks by default.
The only change required for the build to succeed was adding a docs page for the `Graphs.Test` submodule by @simonschoelly, while specifying that it is not public API.


hyrodium (approved):
LGTM!

simonschoelly:
Why should `Graphs.Test` not be part of the public api?

gdalle:
That's a good point. I wanted this PR to focus on the Documenter change, without introducing API commitments that we may not want to honor. That's why I specified that it was internal for the time being, but if we deem Graphs.Test to be stable enough, we can definitely make it public

PR #323, opened 2023-12-21
Fix formatting for file vf2.jl and remove `ignore` in test + autoformatter pass

gdalle:
PR #190 introduced a BlueStyle formatting check, but back then the file `src/Experimental/vf2.jl` caused problems to JuliaFormatter. So it was ignored as follows: `format(Graphs; ignore=["vf2.jl"])`.

Some time later, PR #210 changed this to `format(Graphs; ignore="vf2.jl")`, which seems to also have ignored a lot of other files. The reasons are a bit unclear to me.

I checked again, and now JuliaFormatter can handle `src/Experimental/vf2.jl` without erroring. I removed the `ignore` keyword in the test, and reformatted all the files that had been overlooked since #210.

The only meaningful changes are in `test/runtests.jl` and `src/Experimental/vf2.jl`, the rest is just JuliaFormatter doing its thing.

### DIFF
--- a/test/shortestpaths/astar.jl
+++ b/test/shortestpaths/astar.jl
@@ -10,7 +10,10 @@
             @inferred(a_star(g, 1, 4, d2))
         @test isempty(@inferred(a_star(dg, 4, 1)))
     end
-    @testset "GenericGraph and GenricDiGraph with SimpleEdge" for g in test_generic_graphs(g3), dg in test_generic_graphs(g4)
+    @testset "GenericGraph and GenricDiGraph with SimpleEdge" for g in
+                                                                  test_generic_graphs(g3),
+        dg in test_generic_graphs(g4)
+
         zero_heuristic = n -> 0
         Eg = SimpleEdge{eltype(g)}
         Edg = SimpleEdge{eltype(dg)}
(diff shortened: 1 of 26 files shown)


gdalle:
@mcognetta available for a quick review? This is mostly formal

test/shortestpaths/astar.jl
```diff
@@ -10,7 +10,10 @@
             @inferred(a_star(g, 1, 4, d2))
         @test isempty(@inferred(a_star(dg, 4, 1)))
     end
-    @testset "GenericGraph and GenricDiGraph with SimpleEdge" for g in test_generic_graphs(g3), dg in test_generic_graphs(g4)
+    @testset "GenericGraph and GenricDiGraph with SimpleEdge" for g in
```
mcognetta, line 13: this is all from the automatic formatter, right?

mcognetta (changes requested):
it seems this is entirely a autoformatter review. i would also change the name of the pr to reflect that other files were formatted as well. otherwise lgtm.

test/shortestpaths/astar.jl
mcognetta: i point this case out specifically, because it looks funky to me.

gdalle: yeah, it does look weird but it's JuliaFormatter's work
thanks for pointing it out!

gdalle (commented):
all checked

Issue #324, opened 2023-12-25
[BUG] mincut() gives wrong result


axsk:
Attempting to use the `mincut()` in todays advent of code problem I had to discover that it was not actually finding the min-cut.

The provided graph has a min-cut of 3 edges. `mincut()` on the other side returns another cut with 4 edges.

The following excerpt computes the true min-cut via `karger_min_cut`, and also computes the wrong one with `mincut()` of length 4.

```julia
function test_mincut()
    g, nodes = parseinput(readlines("25.in"))
    while true
        cut = karger_min_cut(g)
        length(karger_cut_edges(g, cut)) == 3 && break
    end

    cut2, _ = mincut(g)
    @show cut == cut2
    @show length.(karger_cut_edges.(Ref(g), [cut, cut2]))
end
```
The whole code+graph data (file `25.in`) can be found [here](https://github.com/axsk/advent-of-code/blob/main/2023/25.jl).

Julia 1.10rc3 with Graphs 1.9.0.

Moelf:
I have found that:
- https://juliagraphs.org/Graphs.jl/dev/algorithms/cut/#Graphs.karger_cut_edges-Union{Tuple{T}, Tuple{AbstractGraph{T}, Vector{<:Integer}}} where T<:Integer

can over estimate the number of graphs as well, for my input I see occasional `760, 790` while the curret answer should be `759, 791`  (for size of the nodes in `1` and `2`)

gdalle:
Thank you for filing the issue! It seems you wrote your [own version of `mincut`](https://github.com/axsk/advent-of-code/blob/main/2023/25%20sw.jl) instead? Could you explain what you changed to fix it?

gdalle:
Related: #105

axsk:
I ported the first C++ implementation from [Wikipedia](https://en.wikipedia.org/wiki/Stoer%E2%80%93Wagner_algorithm#Example_code) (seem pretty solid and nice to me).
They keep all state in a single copy of the weight matrix, doing the vertex-merging by weight updates.

I tried crosschecking it with the packages `mincut` but could not really align them.
[...]

PR #325, opened 2023-12-27
fix mincut()

axsk:
To my understanding `adj_cost` stores the cut weight between the last two vertices, s and t.
If that is lower then the current best cut it becomes the new best.

I do not really understand the `cutweight` computation inside the mincut phase. (It reminds me more of a flow computation) but also didn't think hard enough about the new intermediate pruning.

Anyways, with this change I get the correct cut for the AoC graph (c.f. #324).
I did not test this any further, so please someone knowledgable look into it before merge

### DIFF
--- a/src/traversals/maxadjvisit.jl
+++ b/src/traversals/maxadjvisit.jl
@@ -106,7 +106,7 @@ function _merge_vertex!(merged_vertices, fadjlist, is_merged, w, u, v)
         w[v2, root] = w[root, v2]
     end
     # update neighbors
-    fadjlist[root] = union(fadjlist[root], fadjlist[non_root])
+    union!(fadjlist[root], fadjlist[non_root])
     for v in fadjlist[non_root]
         if root ∉ fadjlist[v]
             push!(fadjlist[v], root)
--- a/src/traversals/maxadjvisit.jl
+++ b/src/traversals/maxadjvisit.jl
@@ -64,8 +64,7 @@ assumed to be 1.
         local cutweight
         while true
             last_vertex = u
-            u, cutweight = first(pq)
-            dequeue!(pq)
+            u, cutweight = dequeue_pair!(pq)
             isempty(pq) && break
             for v in fadjlist[u]
                 (is_processed[v] || is_merged[v] || u == v) && continue
--- a/src/traversals/maxadjvisit.jl
+++ b/src/traversals/maxadjvisit.jl
@@ -63,6 +63,7 @@ assumed to be 1.
             cutweight += w[u, v]
         end
         # Minimum cut phase
+        local adj_cost
         while true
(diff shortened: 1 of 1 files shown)


axsk:
I removed the weird computation of `cutweight` alltogether and set it to the weight of the last cut (what was `adj_weight`) before. 
To my understanding this is what plain Stoer-Wagner does.
I now believe the trick to remove stronger connected vertices early still works with this "simple" cutweight computation.

src/traversals/maxadjvisit.jl
```diff
@@ -106,7 +106,7 @@ function _merge_vertex!(merged_vertices, fadjlist, is_merged, w, u, v)
         w[v2, root] = w[root, v2]
     end
     # update neighbors
-    fadjlist[root] = union(fadjlist[root], fadjlist[non_root])
+    union!(fadjlist[root], fadjlist[non_root])
```
axsk, line 109: I believe linked lists would improve performance further (this needs to grow memory).
However, it seems they are broken atm (c.f. https://github.com/JuliaCollections/DataStructures.jl/issues/886)

gdalle:
Thanks for this contribution! I will need a bit of time to review it but I fully intend to

gdalle:
Feel free to ping me if I take too long

axsk:
Sure let me know if you have any questions.

src/traversals/maxadjvisit.jl
axsk: MutableLinkedLists are fixed by now, but they don't provide unique elements as Set does, so they are not suited here.

etiennedeg:
Oh well, you are right, the computation of cutweight inside the cut of the phase is unneeded.
What it does is to compute the running cut of the phase (which upper bound the mincut).
It was part of the previous incorrect implementation when the cutweght was compared to the mincut at each step of each phase.
I kept it in a first time because used in combination with my heuristic, it may save times, but according to by benchmarks, it was not worth it, so I removed it, but then, the whole cutweight calculation is unneeded.

Thus I think we can continue with this PR, I will close mine.

> To my understanding `adj_cost` stores the cut weight between the last two vertices, s and t.
> If that is lower then the current best cut it becomes the new best.

This is only true at the last step of the phase. during the phase, it is equal to $w(A_v, v)$ from [wikipedia](https://en.wikipedia.org/wiki/Stoer%E2%80%93Wagner_algorithm#Proof_of_correctness), and it gives a lower bound on the $uv$-mincut, where $u$ is the last vertex added before $v$ (which is the justification of the heuristic)

etiennedeg:
If nobody disagree, I will merge, there is nothing controversial in these changes

etiennedeg (approved):

PR #327, opened 2024-01-15
fix mincut (fixes #324)

### LINKED ISSUE #324: [BUG] mincut() gives wrong result

etiennedeg:
When a vertex is merged during the cut of the phase, if the current vertex is not kept, we need to switch to the kept vertex

fixes #324

### DIFF
--- a/src/traversals/maxadjvisit.jl
+++ b/src/traversals/maxadjvisit.jl
@@ -26,6 +26,9 @@ assumed to be 1.
     # in which case we'll return immediately.
     (nvg > one(U)) || return (Vector{Int8}([1]), zero(T))
 
+    # to avoid reallocating lists in fadjlist, we have some already merged vertices
+    # still appearing in fadjlist. When iterating neighbors, is_merged makes sure
+    # don't consider them
     is_merged = falses(nvg)
     merged_vertices = IntDisjointSets(U(nvg))
     graph_size = nvg
@@ -75,6 +78,7 @@ assumed to be 1.
             dequeue!(pq)
             isempty(pq) && break
             for v in fadjlist[u]
+                # ignore already merged neighbors
                 (is_merged[v] || u == v) && continue
                 # if the target of e is already marked then decrease cutweight
                 # otherwise, increase it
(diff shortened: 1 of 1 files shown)


gdalle:
@etiennedeg did you check out https://github.com/JuliaGraphs/Graphs.jl/pull/325? haven't had time to review yet

etiennedeg:
Yes, his implementation drops the heuristics but the heuristic was not at fault. I noticed that he struggled to understand my code, so I added more comments

axsk:
It's not true I did not drop the heuristic.

What is true that I did not understand is the algorithm itself, esp. the part with the `cutweight -= ew` which is why removed those parts to obtain the original `https://en.wikipedia.org/wiki/Stoer-Wagner_algorithm` but WITH the heuristic still valid.

(And some performance improvements)

axsk:
So I agree that updating u to the merged root should be done and have updated my PR again.
I would appreciate if you could give it another look @etiennedeg

axsk:
If you come to the conclusion you don't like my simplifications or they are not working with the heuristic let me know so I will close my PR and prepare a seperate one with the performance fixes.

etiennedeg:
Closing in favor of #325 (see https://github.com/JuliaGraphs/Graphs.jl/pull/325#issuecomment-1900129402)

Issue #326, opened 2024-01-14
Question regarding `non_backtracking_matrix` function's return type


KeishiS:
Thank you for your ongoing development efforts.
I have a question concerning the `non_backtracking_matrix` function. I noticed it returns a non-sparse matrix, whereas the `adjacency_matrix` function returns a sparse matrix. Can I ask you for some insight into the reasons behind this choice?

If there's no specific reason for using a non-sparse matrix, I'd love to submit a pull request to modify the function so that it returns a sparse matrix.

Thanks for your time!

gdalle:
I don't think there is a reason for it not to be sparse, feel free to submit a PR fixing it! Review may take a while but if the PR is simple enough we shouldn't have any issues

PR #328, opened 2024-01-17
Modify `non_backtracking_matrix` function's Return Type

KeishiS:
As mentioned in #326, I modified `non_backtracking_matrix` so that it returns a spase matrix.

### DIFF
--- a/src/linalg/nonbacktracking.jl
+++ b/src/linalg/nonbacktracking.jl
@@ -27,9 +27,9 @@ function non_backtracking_matrix(g::AbstractGraph)
     end
 
     nz = if is_directed(g)
-        sum(indegree(g) .* outdegree(g))
+        sum(indegree(g, v) * outdegree(g, v) for v in vertices(g))
     else
-        sum(degree(g) .* (degree(g) .- 1))
+        sum(degree(g, v) * (degree(g, v) - 1) for v in vertices(g))
     end
     rowidx = sizehint!(Vector{Int}(), nz)
     colidx = sizehint!(Vector{Int}(), nz)
--- a/src/linalg/nonbacktracking.jl
+++ b/src/linalg/nonbacktracking.jl
@@ -35,10 +35,13 @@ function non_backtracking_matrix(g::AbstractGraph)
     colidx = sizehint!(Vector{Int}(), nz)
     for (e, u) in edgeidmap
         i, j = src(e), dst(e)
+        for k in inneighbors(g, i)
+            k == j && continue
+            v = edgeidmap[Edge(k, i)]
 
-        rows = [edgeidmap[Edge(k, i)] for k in inneighbors(g, i) if k != j]
-        append!(rowidx, rows)
-        append!(colidx, fill(u, length(rows)))
+            push!(rowidx, v)
+            push!(colidx, u)
+        end
     end
 
     B = sparse(rowidx, colidx, ones(Int, length(rowidx)), m, m)
--- a/src/linalg/nonbacktracking.jl
+++ b/src/linalg/nonbacktracking.jl
@@ -26,17 +26,20 @@ function non_backtracking_matrix(g::AbstractGraph)
         end
     end
 
-    B = zeros(Float64, m, m)
-
+    rowidx = Int[]
+    colidx = Int[]
     for (e, u) in edgeidmap
         i, j = src(e), dst(e)
         for k in inneighbors(g, i)
             k == j && continue
             v = edgeidmap[Edge(k, i)]
-            B[v, u] = 1
+
+            push!(rowidx, v)
+            push!(colidx, u)
         end
     end
 
+    B = sparse(rowidx, colidx, ones(Int, length(rowidx)), m, m)
     return B, edgeidmap
 end
 
--- a/src/linalg/nonbacktracking.jl
+++ b/src/linalg/nonbacktracking.jl
@@ -26,17 +26,19 @@ function non_backtracking_matrix(g::AbstractGraph)
         end
     end
 
-    rowidx = Int[]
-    colidx = Int[]
+    nz = if is_directed(g)
+        sum(indegree(g) .* outdegree(g))
+    else
+        sum(degree(g) .* (degree(g) .- 1))
+    end
+    rowidx = sizehint!(Vector{Int}(), nz)
+    colidx = sizehint!(Vector{Int}(), nz)
     for (e, u) in edgeidmap
         i, j = src(e), dst(e)
-        for k in inneighbors(g, i)
-            k == j && continue
-            v = edgeidmap[Edge(k, i)]
 
-            push!(rowidx, v)
-            push!(colidx, u)
-        end
+        rows = [edgeidmap[Edge(k, i)] for k in inneighbors(g, i) if k != j]
+        append!(rowidx, rows)
+        append!(colidx, fill(u, length(rows)))
     end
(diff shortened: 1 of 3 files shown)


src/linalg/nonbacktracking.jl
```diff
@@ -26,17 +26,20 @@ function non_backtracking_matrix(g::AbstractGraph)
         end
     end
 
-    B = zeros(Float64, m, m)
-
+    rowidx = Int[]
```
gdalle, line 29: Since we know the size we can allocate the full vector ahead of time and avoid pushing

src/linalg/nonbacktracking.jl
```diff
@@ -26,17 +26,20 @@ function non_backtracking_matrix(g::AbstractGraph)
...
             k == j && continue
             v = edgeidmap[Edge(k, i)]
-            B[v, u] = 1
+
+            push!(rowidx, v)
+            push!(colidx, u)
         end
     end
 
+    B = sparse(rowidx, colidx, ones(Int, length(rowidx)), m, m)
```
gdalle, line 42: Can you add a test to make sure that the returned matrix is sparse?

gdalle (changes requested):
Thanks for the contribution! I made a few remarks but once these are addressed we can merge

src/linalg/nonbacktracking.jl
KeishiS: Thanks for your reviewing. I modified to allocate a capacity by `sizehint!`. The size `nz` is calculated as follows:

$$
\text{nz} \coloneqq \begin{cases}
\displaystyle
\sum_{u\in V} d_{u} (d_{u} - 1), & \text{graph is undirected}
\\
\displaystyle
\sum_{u\in V} d_{u}^{\text{in}} d_{u}^{\text{out}}, & \text{graph is directed}
\end{cases}
$$

src/linalg/nonbacktracking.jl
KeishiS: Thanks! I added some tests to check returned type and matrix values of simple graphs.

src/linalg/nonbacktracking.jl
```diff
@@ -26,17 +26,22 @@ function non_backtracking_matrix(g::AbstractGraph)
...
+    colidx = sizehint!(Vector{Int}(), nz)
     for (e, u) in edgeidmap
         i, j = src(e), dst(e)
-        for k in inneighbors(g, i)
-            k == j && continue
-            v = edgeidmap[Edge(k, i)]
-            B[v, u] = 1
-        end
+
+        rows = [edgeidmap[Edge(k, i)] for k in inneighbors(g, i) if k != j]
```
gdalle, line 39: iterating over neighbors like you did before would be just as efficient, and lead to less allocations

src/linalg/nonbacktracking.jl
```diff
@@ -26,17 +26,22 @@ function non_backtracking_matrix(g::AbstractGraph)
         end
     end
 
-    B = zeros(Float64, m, m)
-
+    nz = if is_directed(g)
+        sum(indegree(g) .* outdegree(g))
```
gdalle, line 30: this allocates needlessly

gdalle (changes requested):
minor efficiency considerations, otherwise I think this will be good to go

gdalle:
Hi @KeishiS just checking in to see if you want to take care of the last changes or if I should?

gdalle:
No worries, thanks!

src/linalg/nonbacktracking.jl
KeishiS: Is this what you have in mind for the code?

```julia
rows = sizehint!(Vector{Int}(), indegree(g, i))
for k in inneighbors(g, i)
    k == j && continue
    push!(rows, edgeidmap[Edge(k, i)])
end
append!(rowidx, rows)
```

src/linalg/nonbacktracking.jl
KeishiS: Do you mean that it is better not to assign the value to `nz`?

gdalle: I mean that `indegree(g)` and `outdegree(g)` are vectors, so you allocate two vectors plus a third one for their elementwise product, just to compute the sum

src/linalg/nonbacktracking.jl
gdalle: why do you need to store the rows before iterating on them? just do it within a loop over neighbors

src/linalg/nonbacktracking.jl
```diff
@@ -26,17 +26,22 @@ function non_backtracking_matrix(g::AbstractGraph)
...
+    else
+        sum(degree(g) .* (degree(g) .- 1))
```
gdalle, line 32: same issue here for the degree

src/linalg/nonbacktracking.jl
gdalle: you can compute it in a for loop with zero allocation:
```julia
nz = sum(indegree(g, v) * outdegree(g, v) for v in vertices(g))
```

gdalle (changes requested):

src/linalg/nonbacktracking.jl
KeishiS: I appreciate your suggestion. Thank you very much!

src/linalg/nonbacktracking.jl
KeishiS: I modified this part. If the code does not reflect your intentions, I would appreciate it if you could correct it.

gdalle: yeah that's good

test/linalg/nonbacktracking.jl
```diff
@@ -0,0 +1,61 @@
...
+        0 0 0 0 0 0 0 0
+        0 0 0 0 0 0 0 0
+        0 0 0 0 1 0 0 0
+        0 0 0 0 0 1 0 0
+        0 0 0 0 0 0 1 0
+    ]
+    egs = Edge.([(1, 2), (2, 3), (3, 4), (4, 5), (2, 1), (3, 2), (4, 3), (5, 4)])
+    indices = getindex.(Ref(edgemap), egs)
+    @test typeof(B) <: SparseMatrixCSC
+    @test all(B[indices, indices] .== B_)
```
gdalle, line 30:
```suggestion
    @test typeof(B) <: SparseMatrixCSC
    @test B == B_
```

test/linalg/nonbacktracking.jl
```diff
@@ -0,0 +1,61 @@
...
+        0 0 0 1 1 0
+        0 0 0 1 1 0
+        0 0 0 0 0 0
+        0 0 0 0 0 0
+        0 0 0 0 1 0
+    ]
+    egs = Edge.([(1, 2), (1, 3), (2, 3), (3, 4), (3, 5), (4, 3)])
+    indices = getindex.(Ref(edgemap), egs)
+    @test typeof(B) <: SparseMatrixCSC
+    @test all(B[indices, indices] .== B_)
```
gdalle, line 60:
```suggestion
    @test typeof(B) <: SparseMatrixCSC
    @test B == B_
```

gdalle (changes requested):
almost there, just a few things to tweak in the tests

gdalle (approved):

Issue #108, opened 2022-02-22
Move code from GraphFlows and GraphsMatching to Graphs


etiennedeg:
As discussed on slack, it is annoying to have `GraphFlows.jl` and `GraphsMatching.jl` as separate packages. We should see what we can bring back to `Graph.jl` and maybe create a package with the more complex problems that need JuMP as a dependency.

gdalle:
Thank you for opening the issue! I agree with this proposition

gdalle:
By the way, this could be an opportunity for standardizing package names: half of them start with `Graphs` and the other half with `Graph` (no s), with no apparent logic to me

gdalle:
I would like to start splitting the code that requires JuMP or other outside optimizers, what do you think about the name `GraphsOptim` for this satellite package?

gdalle:
I have started a first draft at https://github.com/gdalle/GraphsOptim.jl, when it is mature enough I'll transfer it to the JuliaGraphs organization (cc @matbesancon).

I feel like the network interdiction problems of [GraphsExtras.jl](https://github.com/JuliaGraphs/GraphsExtras.jl) are sufficiently specific and advanced enough to deserve their individual package, unlike flows and matchings which are (in my opinion) more mainstream. With that in mind, what kind of solver-requiring algorithms would we want to add in GraphsOptim.jl?
I was thinking of Wardrop equilibria and congestion problems, but that would require a dependency on Convex.jl in addition to JuMP.jl, so maybe not reasonable?

matbesancon:
why Convex.jl for Wardrop?

gdalle:
Following up on this, I am wondering if GraphsOptim.jl actually deserves to exist at all. The new [JuMP tutorials](https://jump.dev/JuMP.jl/stable/tutorials/linear/network_flows/) are very well done, and I'm not sure providing ready-made LP formulations makes much sense in this case.
Thus I propose to move all the combinatorial code for flows, matchings and so on into the main library, and forget the rest.
All those in favor, say aye!

matbesancon:
GraphsMatching will still exist because it depends on the blossom implementation.
Even if there are tutorials, having a ready-to-use package can be handy?

gdalle:
My  bad, I didn't know there were other nontrivial dependencies for GraphsMatching and GraphsFlows besides JuMP.
[...]

PR #329, opened 2024-01-17
Migrate flow algorithms from GraphsFlows

gdalle:
Take everything in [GraphsFlows.jl](https://github.com/JuliaGraphs/GraphsFlows.jl) and put it in Graphs.jl.

This is half of the solution to https://github.com/JuliaGraphs/Graphs.jl/issues/108

- [x] remove the min cost flow part, which relies on JuMP.jl (see GraphsOptim.jl)
- [x] copy-paste the code to `src/flows`
- [x] copy-paste the tests to `test/flows`
- [x] copy-paste the docs and put it in `docs/src/flows.md`
- [x] modify `src/Graphs.jl` and `test/runtests.jl` and `docs/make.jl` to include necessary files
- [x] format everything
- [x] modify docs to include private names in a separate section
- [x] modify nearly all tests to use generic graphs

@matbesancon since it's your code do you have anything to say about this?


gdalle:
Tests are failing because type inference in 1.6 is less good

matbesancon:
I'm good with it yup thanks!

gdalle:
This is the part that is not well inferred:

```julia
@traitfn residual(flow_graph::::IsDirected) = Graphs.DiGraph(Graphs.Graph(flow_graph))
```

simonschoelly:
Can I suggest that we move everything initially to the `Experimental` submodule. This way
* We won't have any name clashes as long as we don't export anything.
* We don't initially guarantee that the API for flows is stable so we still can't change things
* It is easier to track what we change from `GraphFlows.jl` when we do this in a separate PR instead of doing it in this one, where we don't have a history.

gdalle:
That's a good idea but as for tracking changes, beyond renaming I only did the bare minimum so that the code could run. A lot of initial changes are gonna be necessary even in the `Experimental` submodule

gdalle:
Also I'm a bit lazy at the prospect of redoing all of these necessary changes from scratch just so we can track the renaming, but I understand your point

gdalle:
Closing this to reopen a cleaner PR with visible changes

PR #330, opened 2024-01-29
Put formal tests (Aqua, JET, JuliaFormatter) inside the main test set

gdalle:
Upside: CI does not terminate if the code fails Aqua / JET / JuliaFormatter / Documenter tests, and can run the test suite until the end to detect other errors
Downside: the full test suite will always be run, including when JET can guarantee something will go wrong

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -154,11 +154,11 @@ tests = [
         end
     end
 
-    @testset verbose = true "Code quality (Aqua.jl)" begin
+    @testset "Code quality (Aqua.jl)" begin
         Aqua.test_all(Graphs; ambiguities=false)
     end
 
-    @testset verbose = true "Code formatting (JuliaFormatter.jl)" begin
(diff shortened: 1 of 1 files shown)


gdalle:
where do you see that? the tests are passing on 1.10, aren't they?

simonschoelly:
I meant, during the nightly run we see now Documenter failing (although I don't understand why) - It is very much possible that this would have already been the case before but it was hidden because the test would not run further than JET.

gdalle:
I think that is due to unstable RNGs between Julia versions, which is why doctests should probably use StableRNGs.jl.
One more argument for merging this, if you approve? I just opened https://github.com/JuliaGraphs/Graphs.jl/issues/341 to keep track of the doctest issue

simonschoelly (approved):
Ah yeah sure

gdalle:
thanks!

PR #336, opened 2024-01-30
Implement Kruskal traversal iterator

ameligrana:
part of #163, to be revisited after that pr is merged


src/iterators/kruskal.jl
```diff
@@ -0,0 +1,46 @@
...
+
+
+"""
+    mutable struct KruskalIteratorState
+`KruskalIteratorState` is a struct to hold the current state of iteration which is need for the Base.iterate() function.
+"""
+mutable struct KruskalIteratorState <: AbstractIteratorState
+    edge_id::Int
+    mst_len::Int
+end
```
simonschoelly, line 33: What is the advantage of having a mutable iterator  state here? Couldn't we simply return a new state?

src/iterators/kruskal.jl
```diff
@@ -0,0 +1,46 @@
...
+        edge_list = collect(edges(graph))
+        for e in edge_list
+            push!(weights, distmx[src(e), dst(e)])
+        end
+        e = edge_list[sortperm(weights; rev=!minimize)]
+        new(graph, IntDisjointSets(nv(graph)), distmx, e)
+    end
+end
+
+Base.length(t::KruskalIterator) = nv(t.graph)-1
```
simonschoelly, line 22: Is that true in case the graph is not connected?

simonschoelly:
To be honest, I am not sure if an iterator is of much use here. We seem to do most of the allocations when create the iterator and we also do not seem to free any memory after each iteration. But perhaps I am wrong here.

src/iterators/kruskal.jl
Tortar: I think you are completely correct, but I don't have time to work on this at the moment, instead I would be willing to finish up #163, if you happen to have some review time I think it could be better to redirect it there :-)

(by the way I think also the other PR suffer from the same problem)

Krastanov:
This has stalled since the last round of review. For now I will mark it as a draft, to keep my review queue a bit more organized, but do not hesitate to convert it to not a draft if you have the energy to continue this submission.

Krastanov:
forgot to note: this would also need to be tested for correctness against the related existing code

PR #339, opened 2024-02-07
Add CodeCov token

gdalle:
see https://discourse.julialang.org/t/psa-new-version-of-codecov-action-requires-additional-setup/109857

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
       - uses: codecov/codecov-action@v4
         with:
           file: lcov.info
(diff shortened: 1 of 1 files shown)


gdalle:
Do not merge: codecov is failing for server error reasons
https://github.com/JuliaGraphs/Graphs.jl/actions/runs/7826337073/job/21352010540?pr=339

gdalle:
ok we're good

simonschoelly (approved):
Looks good to me. I guess if we merge this, if the code coverage is indeed correctly reported on the bade.

gdalle:
> on the bade.

what does this mean?

PR #340, opened 2024-02-07
Added types and note to `erdos_renyi` function

SuperGrobi:
I added typed-definitions to the docstring of `erdos_renyi`, as we differentiate the two different definitions of the Erdös-Rényi ensemble by dispatching on the type of the second argument.

As I spent some time with the strange edge case of `erdos_renyi(n, 1) != erdos_renyi(n , 1.0)` today this feels like a good idea.

### DIFF
--- a/src/SimpleGraphs/generators/randgraphs.jl
+++ b/src/SimpleGraphs/generators/randgraphs.jl
@@ -127,7 +127,7 @@ function randbn(
 end
 
 """
-    erdos_renyi(n, p)
+    erdos_renyi(n, p::Real)
 
 Create an [Erdős–Rényi](http://en.wikipedia.org/wiki/Erdős–Rényi_model)
 random graph with `n` vertices. Edges are added between pairs of vertices with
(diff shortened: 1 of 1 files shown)


simonschoelly (approved):
I like these changes.

Eventually we should fix this mess and use something like `erdos_renyi(n; probability=0.1)` and `erdos_renyi(n; num_edges=23)`.

Approved, but will merge it later, so someone else will have time to comment on this if they have a different opinion.

And don't worry about the failing tests, nightly is currently broken.

gdalle (approved):

gdalle:
Some notes:
- nightly isn't broken, JET is just more sensitive, so we'll have to fix it ourselves eventually (but not here of course)
- DocStringExtensions.jl might be a good tool in the future to automatically generate typed signatures in the docstrings and avoid this uncertainty

simonschoelly:
>     * DocStringExtensions.jl might be a good tool in the future to automatically generate typed signatures in the docstrings and avoid this uncertainty

Afaik it is not always recommended to put the type signature in the doc string. This is anyways a special case, because `Integer` as subtype of a `Real` leads to a different behavior - something that I don't think is very desirable.

PR #347, opened 2024-02-26
Trivial change, test PR

filchristou:


gdalle:
It's weird cause the actions indeed have read/write permissions in Graphs.jl as well as the organization in general. It's not the most secure setting but it should work

gdalle:
OK, so the workflow completes by setting `permissions: write-all` in the YAML but
- it should be done by default for the GitHub token in the repo settings, not sure why that isn't the case (do you have access to the settings?)
- the comment doesn't seem to show up :thinking:

gdalle:
I think this is the answer: https://github.com/cli/cli/issues/8374#issuecomment-1829442375

It happens cause you're on a fork

gdalle:
`write-all` doesn't change anything, I've tried it

The issue is that you're doing it from a fork, and so the comment posting will never work. In this case the best we can do is use the option from BenchmarkCI to [print the benchmarking results inside the workflow log](https://github.com/tkf/BenchmarkCI.jl?tab=readme-ov-file#printing-benchmark-result-optional), in addition to the PR comment (which will only work from branches, i.e. for PRs made by maintainers).

gdalle:
it seems CI didn't run on those latest commits

filchristou:
yeah. sorry for the noise. I will be playing around for a while. I am not satisfied with the logging public PR benchmarks in the workflow so I am investigating other possibilities. Let me know if you have any ideas.

gdalle:
> When a workflow is triggered by the pull_request_target event, the GITHUB_TOKEN is granted read/write repository permission, even when it is triggered from a public fork. For more information, see "Events that trigger workflows."

Might be worth a shot?

From https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

filchristou:
probably it is worth a chance. Now I am considering a chain of workflows as described here: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

gdalle:
why so complicated? I think `pull_request_target` is much easier

filchristou:
I don't get what's the practical difference between `pull_request` and `pull_request_target`. An activity type of the `pull_request_target` is `opened` which I guess is the default created PR ? Also the [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) mention

> Avoid using this event if you need to build or run code from the pull request.

Which is exactly what benchmarking does.
However, if I can't get the chained workflows to work I will start considering it.

gdalle:
I have asked on slack: https://julialang.slack.com/archives/C681P8ABG/p1709137580258319

gdalle:
On second thought I agree with you this is a bad security practice, let's not use `pull_request_target`

https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

gdalle:
Let's discuss it tonight at the community call if you can attend?

filchristou:
okey. yes I will join

PR #350, opened 2024-03-04
Bringing the `.github/workflows/postbench.yml` to master

filchristou:
as discussed in the [community call](https://github.com/JuliaGraphs/JuliaGraphs-meta/issues/14#issuecomment-1971474422), the trigger workflow file needs to be in the default, i.e. master, branch.

I guess I shouldn't push directly to master, so I made this PR, but maybe for future iterations it makes more sense if I can push directly. That might create some commits, but as already discussed they are not user critical since changes are only made to the CI.


gdalle:
Thanks for the PR, I'll review soon. In accordance with the ColPrac guidelines, please don't push directly and instead open a PR + wait for review if you want to change the master branch.

filchristou:
ok, btw is it normal that the CI fails so hard and that the nighly is still in progress after 3 hours ?

gdalle:
Nightly is unpredictable, but the rest is not normal no

gdalle:
It's the formatting

benchmark/Project.toml
```diff
@@ -0,0 +1,7 @@
+[deps]
+BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
+Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+
+[compat]
+BenchmarkTools = "1"
```
gdalle, line 7: might wanna do 1.5 cause it's the most recent one

.gitignore
```diff
@@ -6,8 +6,11 @@ docs/build/
 docs/site/
 benchmark/.results/*
 benchmark/.tune.jld
+benchmark/Manifest.toml
+.benchmarkci
 *.cov
 /Manifest.toml
+notes
```
gdalle, line 13: this is specific to your setup, no need for it in Graphs.jl

benchmark/benchmarks.jl
```diff
@@ -1,18 +1,36 @@
...
+
+parallelbenchmarks = ["parallel/egonets.jl"]
+
+foreach(parallelbenchmarks) do bm
+    include(joinpath(BENCHDIR, bm))
+end
 
-tune!(suite);
-results = run(suite; verbose=true, seconds=10)
+nothing
```
gdalle, line 36: ?

benchmark/parallel/egonets.jl
```diff
@@ -1,62 +1,57 @@
...
 
-        function mapvertices(f, g::Graph)
-            n = nv(g)
-            a = zeros(Int, n)
-            Threads.@threads for i in 1:n
-                a[i] = f(g, i)
-            end
-            return a
-        end
+SUITE["parallel"] = BenchmarkGroup([], "egonet" => BenchmarkGroup([]))
```
gdalle, line 4: this is no longer necessary from BenchmarkTools v1.4 cause nested dict keys are created on the fly

benchmark/serial/centrality.jl
```diff
@@ -0,0 +1,37 @@
+SUITE["centrality"] = BenchmarkGroup(
+    [], "graphs" => BenchmarkGroup([]), "digraphs" => BenchmarkGroup([])
+)
+
+# graphs
+SUITE["centrality"]["graphs"]["degree_centrality"] = @benchmarkable [
+    Graphs.degree_centrality(g) for (_, g) in $GRAPHS
```
gdalle, line 7: it's weird to benchmark all graphs inside of the same loop, but let's not change the contents of the benchmark for now

test/runtests.jl
```diff
@@ -171,3 +171,5 @@ tests = [
         end
     end
 end;
+
+nothing;
```
gdalle, line 175: ?

.gitignore
```diff
@@ -6,8 +6,11 @@ docs/build/
 docs/site/
 benchmark/.results/*
 benchmark/.tune.jld
+benchmark/Manifest.toml
```
gdalle, line 9: I think this might be necessary for the benchmark CI to work properly? can you double check whether Graphs.jl is properly `dev`-ed into the environment without the manifest?

.github/workflows/runbenchandupload.yml
```diff
@@ -0,0 +1,62 @@
...
+          "
+      - name: Save PR number
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          echo $PR_NUMBER > .benchmarkci/pr_number
+      - uses: actions/upload-artifact@v3
+        with: 
+          name: BenchmarkResults
+          path: .benchmarkci/
```
gdalle, line 61: so this is populated by `BenchmarkCI.judge`?

.github/workflows/runbenchandupload.yml
```diff
@@ -0,0 +1,62 @@
...
+    branches:
+     - master
+     - benchx
+jobs:
+  benchmark:
+    if: ${{ github.event.label.name == 'to-benchmark' }}
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.allow_failure }}
+    permissions: write-all
```
gdalle, line 16: can we be more granular?

.github/workflows/runbenchandupload.yml
```diff
@@ -0,0 +1,62 @@
+name: RunUploadBenchmarks
+env:
+  JULIA_NUM_THREADS: 1
```
gdalle, line 3: do we really want that for multithreaded benchmarks?

.github/workflows/postbench.yml
```diff
@@ -0,0 +1,61 @@
+name: PostBenchmarks
+on:
+  workflow_run:
+    workflows: [RunUploadBenchmarks]
+    types: [completed]
+    branches: [master, benchx]
+jobs:
+  on-success:
+    runs-on: ubuntu-latest
+    # if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
gdalle, line 10: why comment it out?

.github/workflows/postbench.yml
```diff
@@ -0,0 +1,61 @@
...
+          script: |
+            let fs = require('fs');
+            let issue_number = Number(fs.readFileSync('./pr_number'));
+            await github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: issue_number,
+              body: 'Thank you for the PR!'
+            });
+      - name: 'Comment on PR test 3'
```
gdalle, line 53: What is the difference between test 2 and test 3?

.github/workflows/postbench.yml
```diff
@@ -0,0 +1,61 @@
+name: PostBenchmarks
+on:
+  workflow_run:
+    workflows: [RunUploadBenchmarks]
+    types: [completed]
+    branches: [master, benchx]
+jobs:
```
gdalle, line 7: can we give this job granular permissions?

gdalle (changes requested):
Thanks for the huge work! I have some comments, the most important ones are on the yml files, but it's looking good :)

benchmark/benchmarks.jl
filchristou: otherwise the `results` variable is printed when running tests (e.g. with `include("test/runtests.jl")`).

.github/workflows/postbench.yml
filchristou: I expect that to get `permissions: write-all` automatically, but I can add it explicitly to raise our chances we get it right in the first trial.

.github/workflows/postbench.yml
filchristou: They are different approaches of doing the same thing. Since I don't get to push easily to the master to test, I tried to test different approaches in one go.

.github/workflows/runbenchandupload.yml
filchristou: Multithreaded benchmarks are quite poor right now, but if in the future are more developed we can either have a seperate job for them or increase the threads here.
For consistency we should always benchmark serial code with a single thread, so I figured we might as well declare we need a single thread.
However, I was actually been thinking that we might as well use 2 threads just for the parallel precompilation/instantiation which I think Julia supports since a recent version. So, with this mindset, I would then change it back to `JULIA_NUM_THREADS: 2`.

.github/workflows/runbenchandupload.yml
filchristou: yes, the content of `.benchmarkci` is generated by `BenchmarkCI.judge`.

.gitignore
filchristou: right

.gitignore
filchristou: The `dev` happens dynamically in https://github.com/tkf/BenchmarkCI.jl/blob/bca191f70d7fe54bb1c6ecc58e4c9b9d0ee8e4c6/src/BenchmarkCI.jl#L249

However I do believe the `] dev ..` should be hardwired (e.g. for local testing). But luckily we will have a way to do that in the future just with Project.toml because of https://github.com/JuliaLang/Pkg.jl/pull/3783

benchmark/Project.toml
filchristou: okey

benchmark/serial/centrality.jl
filchristou: yes. I didn't want to tweak anything. Once the foundation is there improvements on the content will be the next thing.

benchmark/benchmarks.jl
filchristou: At this case for `include("benchmark/benchmarks.jl")`

test/runtests.jl
filchristou: otherwise the last result is printed in the REPL, and I find that disturbing. See also https://github.com/JuliaGraphs/Graphs.jl/pull/350/files#r1519487219

benchmark/parallel/egonets.jl
filchristou: yes, but let's do that kind of content modifications later :)

filchristou:
@gdalle thanks for the review. Please mark all discussions you are satisfied with as "resolved". When all are resolved I will make the changes.

benchmark/benchmarks.jl
gdalle: You can also run include with a semicolon at the end if you want, so I would remove the nothing

.gitignore
gdalle: Okay then let's keep it that way

.github/workflows/runbenchandupload.yml
gdalle: I don't think this workflow needs any write permission?

benchmark/benchmarks.jl
filchristou: okey.

.github/workflows/runbenchandupload.yml
filchristou: ah yes. I will remove the `permissions` as a whole then.

.github/workflows/postbench.yml
gdalle: by more granular I was actually thinking we restrict the write permissions to the ones we need (probably just PR)

.github/workflows/postbench.yml
gdalle: sounds good

.github/workflows/runbenchandupload.yml
gdalle: we might as well set the number of threads to `"auto"` (these days, GitHub runners have 4)

filchristou: ok

.github/workflows/postbench.yml
filchristou: ah okey, you mean something like that 
```
permissions:
  pull-requests: write
```
I am not sure if that's enough, however since there are many permissions [here ](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#defining-access-for-the-github_token-scopes). Can we push like that and later narrow down the permissions ? Similar to the [previous discussion](https://github.com/JuliaGraphs/Graphs.jl/pull/350#discussion_r1519502292) let's keep it simple in the beginning because debugging is hard in these circumstances.

gdalle: okay for me

.github/workflows/runbenchandupload.yml
```diff
@@ -0,0 +1,61 @@
...
+          "
+      - name: Save PR number
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          echo $PR_NUMBER > .benchmarkci/pr_number
+      - uses: actions/upload-artifact@v3
+        with: 
+          name: BenchmarkResults
+          path: .benchmarkci/
```
gdalle, line 60: the highlighting is funny from here on GitHub's PR review, maybe there's an indent missing or sth?

filchristou: indeed highlighting stop after the quote on line 51.
I don't think the yaml is wrong. Online services like [yalmcheker](https://yamlchecker.com/) do not report any problems with the file. So, I guess it's a github peculiarity.

gdalle (approved):

gdalle:
Following ColPrac @filchristou you can merge your own PR after review since you have the rights

PR #352, opened 2024-03-15
Test bench CI PR

filchristou:
Not to merge

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -1,7 +1,6 @@
 using BenchmarkTools, Graphs
 
 const BENCHDIR = dirname(@__FILE__)
-
 const DIGRAPHS = Dict{String,DiGraph}(
     "complete100" => complete_digraph(100), "path500" => path_digraph(500)
 )


filchristou:
Unfortunately, the PostBenchmarks action from https://github.com/JuliaGraphs/Graphs.jl/blob/master/.github/workflows/postbench.yml was not triggered.

gdalle:
Any idea why?

filchristou:
Not sure at the moment.. But it also doesn't work in my [fork](https://github.com/filchristou/Graphs.jl/pull/5) so I ill first fix it there and then make a PR here.

filchristou:
I got the workflow chain to work, but BenchmarkCI.jl is failing. 
The problem is that BenchmarkCI.jl is engineered such that it uses the current github workflow event to receive the information about where to post. 
Since the new workflow is not triggered with a Pull-Request event the code [fails here](https://github.com/tkf/BenchmarkCI.jl/blob/bca191f70d7fe54bb1c6ecc58e4c9b9d0ee8e4c6/src/BenchmarkCI.jl#L543). 
I know I've been supporting `BenchmarkCI.jl` because it "just works", but it seems that in our case it just doesn't. The options are either to fork it, or make a customized thing out of PkgBenchmark.jl or use something else.

gdalle:
Sorry that you went through all of this trouble to conclude that BenchmarkCI is not the thing for us. Let's put this on pause for now and think about what other options we have for benchmarking infrastructure.

@willow-ahrens is taking over maintenance of BenchmarkTools, and one of her goals is to figure out the CI infrastructure (see the [announcement on Discourse](https://discourse.julialang.org/t/benchmarktools-needs-a-maintainer/111178/10?u=gdalle)). I have explained to her the limits that @filchristou and I have discovered, and she said she'd ping when she has stuff to review in this area.

willow-ahrens:
Is there a simple fix to benchmarkCI that would make it work, or do we need to rewrite it?

Krastanov:
closing in favor of #490

PR #353, opened 2024-03-18
All simple paths (refresh #20)

thchr:
This refreshes #20 by @etiennedeg.
(I couldn't figure out how to do it there directly and also didn't want to bother with fixing merge conflicts either)

The original implementation is by @i-aki-y in https://github.com/sbromberger/LightGraphs.jl/pull/1540.

This "refresh" also simplifies the original implementation a bit (to the extent that I understand it), implements the comments/suggestions that were made in #20 (e.g., `cutoff` is now set to `nv(g)`), and improves the documentation (and removes superfluous documentation of internal methods), and removes redundant methods (e.g., the `length` and `collect` implementations; the latter comes from Base, and the former is a performance trap).

Cc. @gdalle cf. Slack conversation about this (and him pointing out the existence of #20).

I suspect there will be some issue from Formatter.jl: but there's no clue as to what's wrong when I run it locally, just that something is wrong.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Graphs"
 uuid = "86223c79-3864-5bf0-83f7-82e725a168b6"
-version = "1.9.0"
+version = "1.9.1"
 
 [deps]
 ArnoldiMethod = "ec485272-7323-5ecc-a04f-4719b315124d"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Graphs"
 uuid = "86223c79-3864-5bf0-83f7-82e725a168b6"
-version = "1.9.1"
+version = "1.10.0"
 
 [deps]
 ArnoldiMethod = "ec485272-7323-5ecc-a04f-4719b315124d"
--- a/src/traversals/all_simple_paths.jl
+++ b/src/traversals/all_simple_paths.jl
@@ -10,6 +10,9 @@ the graph `g` from a source vertex `u` to a target vertex `v` or iterable of tar
 The iterator's elements (i.e., the paths) can be materialized via `collect` or `iterate`.
 Paths are iterated in the order of a depth-first search.
 
+If the requested path has identical source and target vertices, i.e., if `u = v`, a
+zero-length path `[u]` is included among the iterates.
+
 ## Keyword arguments
 The maximum path length (i.e., number of edges) is limited by the keyword argument `cutoff`
 (default, `nv(g)-1`). If a path's path length is greater than or equal to `cutoff`, it is
@@ -74,6 +77,7 @@ mutable struct SimplePathIteratorState{T<:Integer}
     # (parent vertex, index of children)
     visited::Stack{T}         # current path candidate
     queued::Vector{T}         # remaining targets if path length reached cutoff
+    self_visited::Bool        # in case `u ∈ vs`, we want to return a `[u]` path once only
 end
 function SimplePathIteratorState(spi::SimplePathIterator{T}) where {T<:Integer}
     stack = Stack{Tuple{T,T}}()
@@ -81,7 +85,7 @@ function SimplePathIteratorState(spi::SimplePathIterator{T}) where {T<:Integer}
     queued = Vector{T}()
     push!(visited, spi.u)    # add a starting vertex to the path candidate
     push!(stack, (spi.u, 1)) # add a child node with index 1
-    return SimplePathIteratorState{T}(stack, visited, queued)
+    return SimplePathIteratorState{T}(stack, visited, queued, false)
 end
 
 function _stepback!(state::SimplePathIteratorState) # updates iterator state.
@@ -112,7 +116,7 @@ function Base.iterate(
         end
 
         child = children[next_child_index]
-        next_child_index′ = pop!(state.stack)[2]               # move child index forward
+        next_child_index′ = pop!(state.stack)[2]                 # move child index forward
         push!(state.stack, (parent_node, next_child_index′ + 1)) # ↩
         child in state.visited && continue
 
@@ -145,5 +149,11 @@ function Base.iterate(
                 return result, state
             end
         end
+    end
+
+    # special-case: when `vs` includes `u`, return also a 1-vertex, 0-length path `[u]`
+    if spi.u in spi.vs && !state.self_visited
+        state.self_visited = true
+        return [spi.u], state
     end
 end
--- a/src/traversals/all_simple_paths.jl
+++ b/src/traversals/all_simple_paths.jl
@@ -1,6 +1,6 @@
 """
     all_simple_paths(g, u, v; cutoff=nv(g)) --> Graphs.SimplePathIterator
-    
+
 Returns an iterator that generates all simple paths in the graph `g` from a source vertex
 `u` to a target vertex `v` or iterable of target vertices `vs`.
 
@@ -13,30 +13,32 @@ The maximum path length (i.e., number of edges) is limited by the keyword argume
 omitted.
 
 ## Examples
-```jldoctest
-julia> using Graphs
-julia> g = complete_graph(4)
+```jldoctest allsimplepaths; setup = :(using Graphs)
+julia> g = complete_graph(4);
+
 julia> spi = all_simple_paths(g, 1, 4)
- Graphs.SimplePathIterator(1 → 4)
+SimplePathIterator{SimpleGraph{Int64}}(1 → 4)
+
 julia> collect(spi)
 5-element Vector{Vector{Int64}}:
- [1, 4]
- [1, 3, 4]
- [1, 3, 2, 4]
- [1, 2, 4]
  [1, 2, 3, 4]
+ [1, 2, 4]
+ [1, 3, 2, 4]
+ [1, 3, 4]
+ [1, 4]
 ```
 We can restrict the search to paths of length less than a specified cut-off (here, 2 edges):
-```jldoctest
+```jldoctest allsimplepaths; setup = :(using Graphs)
 julia> collect(all_simple_paths(g, 1, 4; cutoff=2))
+3-element Vector{Vector{Int64}}:
  [1, 2, 4]
  [1, 3, 4]
  [1, 4]
 ```
 """
 function all_simple_paths(
-            g::AbstractGraph{T}, 
-            u::T, 
+            g::AbstractGraph{T},
+            u::T,
             vs;
             cutoff::T=nv(g)
             ) where T <: Integer
@@ -99,7 +101,7 @@ end
 Returns the next simple path in `spi`, according to a depth-first search.
 """
 function Base.iterate(
-            spi::SimplePathIterator{T}, 
+            spi::SimplePathIterator{T},
             state::SimplePathIteratorState=SimplePathIteratorState(spi)
             ) where T <: Integer
 
@@ -121,7 +123,7 @@ function Base.iterate(
             continue
         end
 
-        child = children[next_childe_index]    
+        child = children[next_childe_index]
         first(state.stack)[2] += 1 # move child index forward
         child in state.visited && continue
 
--- a/src/traversals/all_simple_paths.jl
+++ b/src/traversals/all_simple_paths.jl
@@ -117,7 +117,7 @@ function Base.iterate(
 
         child = children[next_child_index]
         next_child_index′ = pop!(state.stack)[2]                 # move child index forward
-        push!(state.stack, (parent_node, next_child_index′ + 1)) # ↩
+        push!(state.stack, (parent_node, next_child_index′ + one(T))) # ↩
         child in state.visited && continue
 
         if length(state.visited) == spi.cutoff
--- a/src/traversals/all_simple_paths.jl
+++ b/src/traversals/all_simple_paths.jl
@@ -1,8 +1,10 @@
 """
     all_simple_paths(g, u, v; cutoff=nv(g)) --> Graphs.SimplePathIterator
 
-Returns an iterator that generates all simple paths in the graph `g` from a source vertex
-`u` to a target vertex `v` or iterable of target vertices `vs`.
+Returns an iterator that generates all 
+[simple paths](https://en.wikipedia.org/wiki/Path_(graph_theory)#Walk,_trail,_and_path) in 
+the graph `g` from a source vertex `u` to a target vertex `v` or iterable of target vertices
+`vs`. A simple path has no repeated vertices.
 
 The iterator's elements (i.e., the paths) can be materialized via `collect` or `iterate`.
 Paths are iterated in the order of a depth-first search.
--- a/src/traversals/all_simple_paths.jl
+++ b/src/traversals/all_simple_paths.jl
@@ -15,7 +15,7 @@ zero-length path `[u]` is included among the iterates.
 
 ## Keyword arguments
 The maximum path length (i.e., number of edges) is limited by the keyword argument `cutoff`
-(default, `nv(g)-1`). If a path's path length is greater than or equal to `cutoff`, it is
+(default, `nv(g)-1`). If a path's path length is greater than `cutoff`, it is
 omitted.
 
 ## Examples
@@ -33,7 +33,7 @@ julia> collect(spi)
  [1, 3, 4]
  [1, 4]
 ```
-We can restrict the search to paths of length less than a specified cut-off (here, 2 edges):
+We can restrict the search to paths of length less than or equal to a specified cut-off (here, 2 edges):
 ```jldoctest allsimplepaths; setup = :(using Graphs)
 julia> collect(all_simple_paths(g, 1, 4; cutoff=2))
 3-element Vector{Vector{Int64}}:
--- a/src/traversals/all_simple_paths.jl
+++ b/src/traversals/all_simple_paths.jl
@@ -1,5 +1,6 @@
 """
-    all_simple_paths(g, u, v; cutoff=nv(g)) --> Graphs.SimplePathIterator
+    all_simple_paths(g, u, v; cutoff)  --> Graphs.SimplePathIterator
+    all_simple_paths(g, u, vs; cutoff) --> Graphs.SimplePathIterator
 
 Returns an iterator that generates all 
 [simple paths](https://en.wikipedia.org/wiki/Path_(graph_theory)#Walk,_trail,_and_path) in 
@@ -11,7 +12,7 @@ Paths are iterated in the order of a depth-first search.
 
 ## Keyword arguments
 The maximum path length (i.e., number of edges) is limited by the keyword argument `cutoff`
-(default, `nv(g)`). If a path's path length is greater than or equal to `cutoff`, it is
+(default, `nv(g)-1`). If a path's path length is greater than or equal to `cutoff`, it is
 omitted.
 
 ## Examples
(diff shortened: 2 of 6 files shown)


gdalle:
documentation seems to be failing

thchr:
> documentation seems to be failing

Should be fixed now, thanks.

src/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,147 @@
+"""
+    all_simple_paths(g, u, v; cutoff=nv(g)) --> Graphs.SimplePathIterator
+
+Returns an iterator that generates all 
+[simple paths](https://en.wikipedia.org/wiki/Path_(graph_theory)#Walk,_trail,_and_path) in 
+the graph `g` from a source vertex `u` to a target vertex `v` or iterable of target vertices
+`vs`. A simple path has no repeated vertices.
```
gdalle, line 7: Add `vs` to the method signatures

src/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,147 @@
...
+
+The iterator's elements (i.e., the paths) can be materialized via `collect` or `iterate`.
+Paths are iterated in the order of a depth-first search.
+
+## Keyword arguments
+The maximum path length (i.e., number of edges) is limited by the keyword argument `cutoff`
+(default, `nv(g)`). If a path's path length is greater than or equal to `cutoff`, it is
```
gdalle, line 14: why not default to `nv(g)-1`?

test/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,108 @@
...
+    add_edge!(g, 3, 5)
+    paths = all_simple_paths(g, 1, [4, 5])
+    @test Set(p for p in paths) == Set([[1, 2, 3, 4], [1, 2, 3, 5]])
+    @test Set(collect(paths)) == Set([[1, 2, 3, 4], [1, 2, 3, 5]])
+
+    # two paths with cutoff
+    g = path_graph(4)
+    add_vertex!(g)
+    add_edge!(g, 3, 5)
+    paths = all_simple_paths(g, 1, [4, 5]; cutoff=3)
```
gdalle, line 24: what happens if the cutoff is 2 here? do we generate `[1, 2, 3]` twice?

test/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,108 @@
...
+    @test Set(p for p in paths) == Set([[1, 2, 3, 4], [1, 2, 3, 5]])
+
+    # two targets in line emits two paths
```
gdalle, line 27: what's the use case for several targets `vs`?

test/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,108 @@
...
+    # two paths digraph with cutoff
+    g = SimpleDiGraph(5)
+    add_edge!(g, 1, 2)
+    add_edge!(g, 2, 3)
+    add_edge!(g, 3, 4)
+    add_edge!(g, 3, 5)
+    paths = all_simple_paths(g, 1, [4, 5]; cutoff=3)
+    @test Set(p for p in paths) == Set([[1, 2, 3, 4], [1, 2, 3, 5]])
+
+    # digraph with a cycle
```
gdalle, line 51: it would be nice to test a graph with a cycle such that a walk going around the cycle is clearly below the (default) cutoff, so that we see it is excluded due to not being a simple path

test/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,108 @@
...
+    add_edge!(g, 2, 4)
+    paths = all_simple_paths(g, 1, [3, 4])
+    @test Set(p for p in paths) == Set([[1, 2, 3], [1, 2, 4]])
+
+    # source equals targets
+    g = SimpleGraph(4)
+    paths = all_simple_paths(g, 1, 1)
+    @test Set(p for p in paths) == Set([])
+
+    # cutoff prones paths
```
gdalle, line 74: prunes

test/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,108 @@
...
+    add_edge!(g, 2, 3)
+    add_edge!(g, 3, 1)
+    add_edge!(g, 2, 4)
+    paths = all_simple_paths(g, 1, [3, 4])
+    @test Set(p for p in paths) == Set([[1, 2, 3], [1, 2, 4]])
+
+    # source equals targets
+    g = SimpleGraph(4)
+    paths = all_simple_paths(g, 1, 1)
+    @test Set(p for p in paths) == Set([])
```
gdalle, line 72: Is the return vector untyped if it is empty?

test/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,108 @@
...
+
+    # cutoff prones paths
+    # Note, a path lenght is node - 1
```
gdalle, line 75: length

src/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,147 @@
...
+
+## Examples
+```jldoctest allsimplepaths; setup = :(using Graphs)
+julia> g = complete_graph(4);
+
+julia> spi = all_simple_paths(g, 1, 4)
+SimplePathIterator{SimpleGraph{Int64}}(1 → 4)
+
+julia> collect(spi)
+5-element Vector{Vector{Int64}}:
```
gdalle, line 25: is this order reliable? or does it depend on internal implementation details of a stack for instance?

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Graphs"
 uuid = "86223c79-3864-5bf0-83f7-82e725a168b6"
-version = "1.9.0"
+version = "1.9.1"
```
gdalle, line 3: actually it's a new feature so semver says 1.10

gdalle (commented):
I have reviewed everything except the actual DFS, because I'm wondering if there is an exploitable overlap with #163

src/traversals/all_simple_paths.jl
thchr: Added

test/traversals/all_simple_paths.jl
thchr: We don't generate anything then: I think that's expected, no? There are no paths of length less than 3.

test/traversals/all_simple_paths.jl
thchr: I can only guess: if someone wants to explore multiple sets of final destinations, this is then possible. It's a feature also in networkx (that's the reasoning given in the earlier PR also https://github.com/JuliaGraphs/Graphs.jl/pull/20#discussion_r1157189693)

test/traversals/all_simple_paths.jl
thchr: No; this is just copied over from the old test-set: the actual output is type-stable. I can fix here though.

src/traversals/all_simple_paths.jl
thchr: It's probably an internal implementation detail. If that changes though, it will be straightforward to update the doctest.

src/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,147 @@
...
+end
+
+function Base.show(io::IO, spi::SimplePathIterator)
+    print(io, "SimplePathIterator{", typeof(spi.g), "}(", spi.u, " → ")
+    if length(spi.vs) == 1
+        print(io, only(spi.vs))
+    else
+        print(io, '[')
+        join(io, spi.vs, ", ")
+        print(io, ']')
```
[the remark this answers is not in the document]
thchr: Sure

src/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,147 @@
...
+    end
+    print(io, ')')
+    return nothing
+end
+Base.IteratorSize(::Type{<:SimplePathIterator}) = Base.SizeUnknown()
+Base.eltype(::SimplePathIterator{T}) where {T} = Vector{T}
+
+mutable struct SimplePathIteratorState{T<:Integer}
+    stack::Stack{Vector{T}} # used to restore iteration of child vertices; each vector has
```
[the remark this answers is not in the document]
thchr: Just the one type parameter :)

src/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,147 @@
...
+    # two elements: a parent vertex and an index of children
```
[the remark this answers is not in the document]
thchr: Just copying over the original implementation - as we discussed on Slack, further optimizations are welcome in the future; my goal is just to get a working implementation in.

I changed to `Tuple{T,T}` now: it makes updating the stack incrementally a bit more complicated, but probably more efficient as you say (although I would guess very small impact, given other costs).

src/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,147 @@
...
+    return SimplePathIteratorState{T}(stack, visited, queued)
+end
+
+function _stepback!(state::SimplePathIteratorState) # updates iterator state.
+    pop!(state.stack)
+    pop!(state.visited)
+    return nothing
+end
+
+# Returns the next simple path in `spi`, according to a depth-first search
```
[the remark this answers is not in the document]
thchr: I don't know: but I don't think it's worth waiting for.

My aim is just to have something across the finish line here: this is the 3rd time the PR has been made. I think that very strongly suggests we should just merge it as soon as we've established correctness and appropriate API, without further micro- or macro-optimizations.

Project.toml
thchr: Okay 👍

test/traversals/all_simple_paths.jl
thchr: All cycles are automatically excluded by the algorithm: i.e., it is unrelated to the cutoff. The cutoff was previously `typemax(Int)`.

thchr:
There is one API question we should settle: what should the set of all paths from `u` to `v = u` be? I.e., what happens when the start and end point are identical? To my mind, it should return `[[u]` which is a path of zero length. Currently, however, it returns no paths, i.e., `[Int[]]`.

I propose we change it to return `[[u]]` in this case.

test/traversals/all_simple_paths.jl
thchr: I've added a test in any case, should the algorithm change.

thchr:
I've gone ahead an implemented the proposed `u = v` special-casing, adding corresponding tests also.

test/traversals/all_simple_paths.jl
gdalle: oh right, I thought we might generate path stubs but we don't

src/traversals/all_simple_paths.jl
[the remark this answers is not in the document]
gdalle: fine by me, doing a second pass today

gdalle:
I agree with your suggestion that paths from `u` to `u` should be `[u]` and not `[]`.
This is also more coherent with the way lengths are measured in the algorithm: a path `[u]` has length `0`

src/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,159 @@
...
+The iterator's elements (i.e., the paths) can be materialized via `collect` or `iterate`.
+Paths are iterated in the order of a depth-first search.
+
+If the requested path has identical source and target vertices, i.e., if `u = v`, a
+zero-length path `[u]` is included among the iterates.
+
+## Keyword arguments
+The maximum path length (i.e., number of edges) is limited by the keyword argument `cutoff`
+(default, `nv(g)-1`). If a path's path length is greater than or equal to `cutoff`, it is
+omitted.
```
gdalle, line 19:
```suggestion
The maximum path length (i.e., number of edges) is limited by the keyword argument `cutoff`
(default, `nv(g)-1`). If a path's path length is greater than `cutoff`, it is
omitted.
```

gdalle: at least that's what I deduce from the example below

src/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,159 @@
...
+
+julia> collect(spi)
+5-element Vector{Vector{Int64}}:
+ [1, 2, 3, 4]
+ [1, 2, 4]
+ [1, 3, 2, 4]
+ [1, 3, 4]
+ [1, 4]
+```
+We can restrict the search to paths of length less than a specified cut-off (here, 2 edges):
```
gdalle, line 36:
```suggestion
We can restrict the search to paths of length less than or equal to a specified cut-off (here, 2 edges):
```

src/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,159 @@
...
+    visited::Stack{T}         # current path candidate
+    queued::Vector{T}         # remaining targets if path length reached cutoff
+    self_visited::Bool        # in case `u ∈ vs`, we want to return a `[u]` path once only
+end
+function SimplePathIteratorState(spi::SimplePathIterator{T}) where {T<:Integer}
+    stack = Stack{Tuple{T,T}}()
+    visited = Stack{T}()
+    queued = Vector{T}()
+    push!(visited, spi.u)    # add a starting vertex to the path candidate
+    push!(stack, (spi.u, 1)) # add a child node with index 1
```
gdalle, line 87:
```suggestion
    push!(stack, (spi.u, one(T))) # add a child node with index 1
```

src/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,159 @@
...
+        parent_node, next_child_index = first(state.stack)
+        children = outneighbors(spi.g, parent_node)
+        if length(children) < next_child_index
+            _stepback!(state) # all children have been checked, step back
+            continue
+        end
+
+        child = children[next_child_index]
+        next_child_index′ = pop!(state.stack)[2]                 # move child index forward
+        push!(state.stack, (parent_node, next_child_index′ + 1)) # ↩
```
gdalle, line 120:
```suggestion
        push!(state.stack, (parent_node, next_child_index′ + one(T))) # ↩
```

src/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,159 @@
...
+        end
+
+        parent_node, next_child_index = first(state.stack)
+        children = outneighbors(spi.g, parent_node)
+        if length(children) < next_child_index
+            _stepback!(state) # all children have been checked, step back
+            continue
+        end
+
+        child = children[next_child_index]
```
gdalle, line 118: this may not work for graphs in which neighbors are not indexable, but let's disregard it

src/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,159 @@
...
+            result = if child in spi.vs
+                vcat(reverse(collect(state.visited)), child)
+            else
+                nothing
+            end
+
+            # update state variables
+            push!(state.visited, child) # move to child vertex
+            if !isempty(setdiff(spi.vs, state.visited)) # expand stack until all targets are found
+                push!(state.stack, (child, 1)) # add the child node as a parent for next iteration
```
gdalle, line 143:
```suggestion
                push!(state.stack, (child, one(T))) # add the child node as a parent for next iteration
```

src/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,159 @@
...
+
+        parent_node, next_child_index = first(state.stack)
+        children = outneighbors(spi.g, parent_node)
+        if length(children) < next_child_index
+            _stepback!(state) # all children have been checked, step back
+            continue
+        end
+
+        child = children[next_child_index]
+        next_child_index′ = pop!(state.stack)[2]                 # move child index forward
```
gdalle, line 119: not a fan of this notation with the prime

src/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,159 @@
...
+        push!(state.stack, (parent_node, next_child_index′ + 1)) # ↩
+        child in state.visited && continue
+
+        if length(state.visited) == spi.cutoff
+            # collect adjacent targets if more exist and add them to queue
+            rest_children = Set(children[next_child_index:end])
```
gdalle, line 125: shouldn't there be a prime here, to avoid including `next_child_index` itself?

src/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,159 @@
...
+            else
+                nothing
+            end
+
+            # update state variables
+            push!(state.visited, child) # move to child vertex
+            if !isempty(setdiff(spi.vs, state.visited)) # expand stack until all targets are found
+                push!(state.stack, (child, 1)) # add the child node as a parent for next iteration
+            else
+                pop!(state.visited) # step back and explore the remaining child nodes
```
gdalle, line 145: should this be `_stepback!`?

src/traversals/all_simple_paths.jl
gdalle: no it's actually right, sorry

test/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,124 @@
+@testset "All simple paths" begin
+    # single path
+    g = path_graph(4)
+    paths = all_simple_paths(g, 1, 4)
+    @test Set(paths) == Set([[1, 2, 3, 4]])
+    @test Set(collect(paths)) == Set([[1, 2, 3, 4]])
```
gdalle, line 6: why the double test every time?

test/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,124 @@
...
+
+    # printing
+    @test sprint(show, paths) == "SimplePathIterator{SimpleGraph{Int64}}(1 → 4)"
+
+    # single path with cutoff
```
gdalle, line 11: complete graph

test/traversals/all_simple_paths.jl
```diff
@@ -0,0 +1,124 @@
...
+
+    # two paths
+    g = path_graph(4)
+    add_vertex!(g)
+    add_edge!(g, 3, 5)
+    paths = all_simple_paths(g, 1, [4, 5])
+    @test Set(paths) == Set([[1, 2, 3, 4], [1, 2, 3, 5]])
+    @test Set(collect(paths)) == Set([[1, 2, 3, 4], [1, 2, 3, 5]])
+
+    # two paths with cutoff
```
gdalle, line 23: in this example I'd put the second target one more hop away, so that it is indeed cut off while the first remains
same for the digraph equivalent below

gdalle (changes requested):
I'm now reasonably confident that the implementation is correct! A few minor tweaks and I can merge

test/traversals/all_simple_paths.jl
thchr: Just a result of copying the old implementation (that had special-cased the `collect` thing). Found two instance of it; kept one to ensure that `collect` keeps working as intended.

test/traversals/all_simple_paths.jl
thchr: done

test/traversals/all_simple_paths.jl
thchr: I added an example like this now.

src/traversals/all_simple_paths.jl
thchr: Yeah, let's get something that works for 99% of the cases merged.

src/traversals/all_simple_paths.jl
thchr: I don't know; I presume not since it is not in the original implementation.

If not crucial, I suggest this is left as-is.

thchr:
Thanks for the second review. I've updated everything accordingly, with the exception of any action on your `_stepback!` question: I simply don't know (and admit that I do not have the time or motivation to find out at the moment; I just want something that works and doesn't stall again).

src/traversals/all_simple_paths.jl
gdalle: okay by me

gdalle (approved):
thank you for the contribution!

mschauer:
Sorry I only see this now that the work is done. Anyway thank you!

mschauer:
Cc @mwien Graphs.jl has now path enumeration

PR #355, opened 2024-03-23
using symbols instead of structures when calling methods from ArnoldiMethod.jl

gkantsidis:
According to https://github.com/JuliaLinearAlgebra/ArnoldiMethod.jl/commit/56221508b32e2be8cc041733f32f1fd669ba1490
the authors of ArnoldiMethods.jl would prefer users to use the symbols `:LM`, `:SR`, `:LR`, `:SI`, `:LI`
instead of `LM()`, `SR()`, `LR()`, `SI()`, and `LI()` respectively.
Accordingly, I have made those adjustments and removed the unecessary now imports of `LM`, `SR`, `LR`, `SI` and `LI`.

### DIFF
--- a/src/Graphs.jl
+++ b/src/Graphs.jl
@@ -3,7 +3,7 @@ module Graphs
 using SimpleTraits
 
 ### Remove the following line once #915 is closed
-using ArnoldiMethod: LM, SR, LR, partialschur, partialeigen
+using ArnoldiMethod: partialschur, partialeigen
 using Statistics: mean
 
 # Currently used to support the ismutable function that is not available in Julia < v1.7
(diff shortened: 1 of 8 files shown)


gdalle:
Just because users can pass symbols doesn't mean they should. In particular, passing the two-letter structs (which are no longer exported but I took care of that in #136) allows type-stability, whereas the symbols probably introduce an additional runtime dispatch due to the conversion:
https://github.com/JuliaLinearAlgebra/ArnoldiMethod.jl/blob/92f50945717cb3daa1ed237897deaa40d77f9535/src/run.jl#L181-L185

gkantsidis:
Although I also prefer the structs, I interpreted their commit comment as an attempt to hide the LR()/LM()/etc from the public interface. Now that I read their comment again, I may have over-interpreted it.

gdalle:
It's ambiguous I agree :) Closing for now, let's keep the structs and if ArnoldiMethod changes again in a cleare way we'll reconsider!

PR #369, opened 2024-04-26
fix merge_vertices

### LINKED ISSUE #363: [BUG] `merge_vertices` returns an undirected graph, even if the input is directed.
`merge_vertices` returns an undirected graph, even if the input is directed.

The output graph should be the same type of graph as the input.

etiennedeg:
Fixes #363 

In the tests, there was a loop encapsulating the `merge_vertices` tests, but these tests were not using the test graph, hence I moved it out and refactored it a bit.

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -34,27 +34,27 @@ Edge 5 => 3
 ```
 """
 function complement(g::Graph)
-	gnv = nv(g)
-	h = SimpleGraph(gnv)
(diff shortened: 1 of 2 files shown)


gdalle (approved):

gdalle:
feel free to merge it @etiennedeg, according to ColPrac the person submitting the PR should be the one to merge it after review if they have the rights

Issue #370, opened 2024-04-28
dorogovtsev_mendes graph generator is running in O(n^2)


AntoineBut:
Generating a graph of size N using the[ dorogovtsev_mendes algorithm](https://arxiv.org/pdf/cond-mat/0106144.pdf#page=24) should run in $O(N)$ and not $O(N^2)$. (see [source](https://github.com/JuliaGraphs/Graphs.jl/blob/0fa5f52856f7f779e8f2ca49bc6cee26e5e19b50/src/SimpleGraphs/generators/randgraphs.jl#L1455-L1481))

I actually don't really understand why this has been implemented this way, and I can try to make a more efficient implementation.

gdalle:
that would be a welcome contribution, thanks!

PR #371, opened 2024-04-29
Faster dorogovtsev mendes

AntoineBut:
Addresses #370.

This implementation has $O(N)$ runtime, while the previous implementation had $O(N^2)$ runtime. 

The optimization comes from storing all edges added at each iteration in an indexed table, allowing to draw a random edge in $O(1)$ time, while the previous implementation had to go through (on average) half on the nodes in order to draw a random edge, which takes $O(N)$ time. 

On my computer, this implementation uses approximately 16% more memory. 

It runs faster than the previous one even on small graphs ($N = 100$).

### DIFF
--- a/src/SimpleGraphs/generators/randgraphs.jl
+++ b/src/SimpleGraphs/generators/randgraphs.jl
@@ -1494,24 +1494,29 @@ function dorogovtsev_mendes(
     n < 3 && throw(DomainError("n=$n must be at least 3"))
     rng = rng_from_rng_or_seed(rng, seed)
     g = cycle_graph(3)
+    bag_of_edges = Vector{SimpleEdge{Int}}(undef, 2*n - 3)
 
-    for iteration in 1:(n - 3)
-        chosenedge = rand(rng, 1:(2 * ne(g))) # undirected so each edge is listed twice in adjlist
-        u, v = -1, -1
-        for i in 1:nv(g)
-            edgelist = outneighbors(g, i)
-            if chosenedge > length(edgelist)
-                chosenedge -= length(edgelist)
-            else
-                u = i
-                v = edgelist[chosenedge]
-                break
-            end
-        end
+    bag_of_edges[1] = SimpleEdge(1, 2)
+    bag_of_edges[2] = SimpleEdge(1, 3)
+    bag_of_edges[3] = SimpleEdge(2, 3)
+    index = 3
+   
+    for _ in 1:n-3
+        # Choose random edge from bag
+        edge = bag_of_edges[rand(rng, 1:index)]
+        u, v = edge.src, edge.dst
 
-        add_vertex!(g)
-        add_edge!(g, nv(g), u)
-        add_edge!(g, nv(g), v)
+        # Add new vertex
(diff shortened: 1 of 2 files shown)


src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -1494,24 +1494,32 @@ function dorogovtsev_mendes(
...
+
+    bag_of_edges[1] = SimpleEdge(1, 2)
+    bag_of_edges[2] = SimpleEdge(1, 3)
+    bag_of_edges[3] = SimpleEdge(2, 3)
+    index = 3
+
+    for _ in 1:(n - 3)
+        # Choose random edge from bag
+        edge = bag_of_edges[rand(rng, 1:index)]
+        u, v = edge.src, edge.dst
```
gdalle, line 1507:
```suggestion
        u, v = src(edge), dst(edge)
```

gdalle: it is generally better to use accessors than to get the fields directly

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -1494,24 +1494,32 @@ function dorogovtsev_mendes(
...
+
+        # Add new vertex
+        add_vertex!(g) || throw(
+            DomainError(
+                "Failed to add vertex. One possible explanation is that type $(eltype(g)) cannot represent enough vertices",
+            ),
```
gdalle, line 1513: why is this necessary?

gdalle (changes requested):
Thanks for the fix, this is nearly good to go!

src/SimpleGraphs/generators/randgraphs.jl
AntoineBut: After adding this I realized that it is currently note possible to use this function to generate a graph where the node types are different than Int64. Making this possible would better fit in a dedicated issue I think (and of coherence this would need to be done with other similar generators).

Currently, only this function allows for user-defined node types : 

`function SimpleDiGraph{T}(
    nv::Integer,
    ne::Integer;
    rng::Union{Nothing,AbstractRNG}=nothing,
    seed::Union{Nothing,Integer}=nothing,
) where {T<:Integer}`

If in the future we allow this kind of freedom, this checks makes sense as it could happen that someone tries to generate a graph too large for its nodes to be represented with the given type (this probably won't happen with Int64 though).

So as of now it isn't necessary.

src/SimpleGraphs/generators/randgraphs.jl
```diff
@@ -1494,24 +1494,32 @@ function dorogovtsev_mendes(
...
+        # Choose random edge from bag
+        edge = bag_of_edges[rand(rng, 1:index)]
+        u, v = src(edge), dst(edge)
+
+        # Add new vertex
+        add_vertex!(g) || throw(
+            DomainError(
+                "Failed to add vertex. One possible explanation is that type $(eltype(g)) cannot represent enough vertices",
+            ),
+        )
```
gdalle, line 1514:
```suggestion
        add_vertex!(g)
```

gdalle (changes requested):

gdalle (approved):

PR #372, opened 2024-05-02
Avoid materializing intermediate adjacency list

simsurace:
For graphs where the adjacency is not stored, this can lead to a lot of allocations when iterating over edges.

Before PR:
```julia
julia> g
{58089, 120148} directed simple Int32 graph

julia> typeof(g)
StaticGraphs.StaticDiGraph{Int32, Int32}

julia> @time collect(Graphs.edges(g));
 21.959129 seconds (330.20 k allocations: 260.015 GiB, 27.19% gc time, 0.21% compilation time)
```

After PR:
```julia
julia> @time collect(Graphs.edges(g));
  0.000396 seconds (3 allocations: 938.844 KiB)
```


KristofferC:
Couldn't this be done in the `iterate` above as well?

gdalle (approved):
Gotta love a 12-line PR with a times-one-gazillion speedup

PR #374, opened 2024-05-04
Remove CI on nightly

### LINKED ISSUE #295: Stop running CI on nightly?
It's not obvious at all that packages should run CI on nightly, for instance because it might hide actual breaking changes in the language itself. I suggest removing the nightly CI

See this [discourse thread](https://discourse.julialang.org/t/why-do-packages-run-continuous-integration-tests-on-julia-nightly/101208/3?u=gdalle)

gdalle:
Fix #295

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,12 +13,10 @@ jobs:
   test:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
-    continue-on-error: ${{ matrix.allow_failure }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


etiennedeg:
What does the allow failure do if the tests still error when nightly errors?

gdalle:
I think it does at least allow the CI badge on README to be green, but I'm not sure of the internals. See https://discourse.julialang.org/t/ignoring-nightly-failure-for-ci-badge/98028/5?u=gdalle for a discussion

Beyond the green checks and red crosses, it also comes down to a philosophical issue of whether we should test on nightly or not. I'd be favorable to testing on the next release candidate as soon as it comes out instead, because it's likely to be more polished and e.g. not make JET or Aqua fail

etiennedeg (approved):

PR #376, opened 2024-05-08
ReverseView and UndirectedView

etiennedeg:


etiennedeg:
@gdalle Should be good for review. Do you know what's wrong with the documentation?

src/wrappedGraphs/graphviews.jl
```diff
@@ -0,0 +1,125 @@
...
+```
+"""
+struct ReverseView{T<:Integer,G<:AbstractGraph} <: AbstractGraph{T}
+    g::G
+
+    @traitfn ReverseView{T,G}(g::::(IsDirected)) where {T<:Integer,G<:AbstractGraph{T}} =
+        new(g)
+    @traitfn ReverseView{T,G}(g::::(!IsDirected)) where {T<:Integer,G<:AbstractGraph{T}} =
+        throw(ArgumentError("Your graph needs to be directed"))
+end
```
simonschoelly, line 31: Wouldn't it make more sense if this function works on all graphs? Even if it does nothing for undirected graphs.

src/wrappedGraphs/graphviews.jl
```diff
@@ -0,0 +1,125 @@
...
+ 2
+
+julia> neighbors(ug, 2)
+1-element Vector{Int64}:
+ 1
+```
+"""
+struct UndirectedView{T<:Integer,G<:AbstractGraph} <: AbstractGraph{T}
+    g::G
+    ne::Int
```
simonschoelly, line 76: Precalculating `ne` is a bit dangerous as one might change the underlying graph and then the view would be invalid. If we cache this value then we should be very clear about that in the documentation.

I think it might be worth not to cache it and then make sure that all our algorithms calculate `ne` only once.

src/wrappedGraphs/graphviews.jl
```diff
@@ -0,0 +1,125 @@
...
+    @traitfn function UndirectedView{T,G}(
+        g::::(IsDirected)
+    ) where {T<:Integer,G<:AbstractGraph{T}}
+        ne = count(e -> src(e) <= dst(e) || !has_edge(g, dst(e), src(e)), Graphs.edges(g))
```
simonschoelly, line 80:
```suggestion
        ne = count(e -> src(e) <= dst(e) && src(e) <= dst(e), Graphs.edges(g))
```
This way we can avoid the expensive call to `has_edge!`

kenahoo: This seems to be checking the same thing twice?

docs/src/core_functions/wrappedgraphs.md
```diff
@@ -0,0 +1,21 @@
...
+```@index
+Pages = ["wrappedgraphs.md"]
+```
+
+## Full docs
+
+```@autodocs
+Modules = [Graphs]
+Pages = [
+    "wrappedgraphs/graphviews.jl",
```
Krastanov:
```suggestion
    "wrappedGraphs/graphviews.jl",
```

Krastanov:
There are a number of functionality and performance improvements left suggested in the review, but all documentation and correctness comments are now addressed. I will proceed with merging this. Future PRs can address the missing features and improve performance.

PR #377, opened 2024-05-20
Use main URL for Wikipedia, not mobile URL

kenahoo:
I noticed these links to the mobile Wikipedia site, figured it would be better to be consistent across the docs.


gdalle (approved):
thanks!

PR #381, opened 2024-06-03
Optimize a bit bfs and dfs iterators

Tortar:
in particular, bfs is like 40% faster

### DIFF
--- a/src/iterators/bfs.jl
+++ b/src/iterators/bfs.jl
@@ -58,7 +58,8 @@ First iteration to visit vertices in a graph using breadth-first search.
 function Base.iterate(t::BFSIterator{<:Integer})
     visited = falses(nv(t.graph))
     visited[t.source] = true
-    return (t.source, BFSVertexIteratorState(visited, [t.source], 1, 1))
+    state = BFSVertexIteratorState(visited, [t.source], 1, 1)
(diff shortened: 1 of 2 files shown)


src/iterators/bfs.jl
```diff
@@ -75,35 +76,33 @@ Iterator to visit vertices in a graph using breadth-first search.
 """
 function Base.iterate(t::BFSIterator, state::BFSVertexIteratorState)
     graph, visited, queue = t.graph, state.visited, state.queue
-    while !isempty(queue)
-        if state.n_visited == nv(graph)
-            return nothing
-        end
```
simonschoelly, line 81: Does that have any influence on performance? Because I feel like expliectly stating that we return `nothing` in that case makes it easier to understand what is going on.

simonschoelly:
I see why `dfs` should always be faster, as there the only improvement there is using `@inbound` and the rest seems to be renaming. For `bfs` a statement like `makes it 40% faster` needs some explanation. At least on what kind of graphs - but better with the whole benchmark setup.

I am saying this, because we also had the case, that people just tested performance on graphs that did not really fit the problem. Therefore I think we should publish the whole methodology when testing for speedup

Tortar:
closing in favour of #382

PR #382, opened 2024-06-04
Fix bfs iterator for multiple source nodes

Tortar:
I noticed that the previous implementation of multi source bfs was wrong, because it didn't start with the first level nodes (also my fault :( ), this should be correct instead, and also faster than the one in #381. I see a 1.7x improvement over the previous version on a `erdos_renyi(1000000, 0.00001)` starting from a random node.

There is still a problem though, I think multi-source dfs suffers from a similar problem. But I unfortunately don't have time to fix it at the moment.

### DIFF
--- a/src/iterators/bfs.jl
+++ b/src/iterators/bfs.jl
@@ -66,6 +66,7 @@ end
 function Base.iterate(t::BFSIterator{<:AbstractArray})
     visited = falses(nv(t.graph))
     curr_level = unique(s for s in t.source)
+    sort!(curr_level)
     visited[curr_level] .= true
     state = BFSVertexIteratorState(visited, curr_level, Int[], 0, 0)
     return Base.iterate(t, state)
@@ -89,6 +90,7 @@ function Base.iterate(t::BFSIterator, state::BFSVertexIteratorState)
             end
         end
         state.curr_level, state.next_level = state.next_level, Int[]
+        sort!(state.curr_level)
         state.node_idx = 0
     end
     # we visit all nodes in this level
--- a/src/iterators/bfs.jl
+++ b/src/iterators/bfs.jl
@@ -91,6 +91,7 @@ function Base.iterate(t::BFSIterator, state::BFSVertexIteratorState)
         end
         length(state.next_level) == 0 && return nothing
         state.curr_level, state.next_level = state.next_level, empty!(state.curr_level)
+        sort!(state.curr_level)
         state.node_idx = 0
     end
     # we visit all nodes in this level
--- a/src/iterators/bfs.jl
+++ b/src/iterators/bfs.jl
@@ -77,9 +77,10 @@ end
 Iterator to visit vertices in a graph using breadth-first search.
 """
 function Base.iterate(t::BFSIterator, state::BFSVertexIteratorState)
-    state.n_visited == nv(t.graph) && return nothing
     # we fill nodes in this level
     if state.node_idx == length(state.curr_level)
+        state.n_visited += length(state.curr_level)
+        state.n_visited == nv(t.graph) && return nothing
         @inbounds for node in state.curr_level
             for adj_node in outneighbors(t.graph, node)
                 if !state.visited[adj_node]
@@ -93,7 +94,6 @@ function Base.iterate(t::BFSIterator, state::BFSVertexIteratorState)
         state.node_idx = 0
     end
     # we visit all nodes in this level
-    state.n_visited += 1
     state.node_idx += 1
     @inbounds node = state.curr_level[state.node_idx]
     return (node, state)
--- a/src/iterators/bfs.jl
+++ b/src/iterators/bfs.jl
@@ -67,7 +67,7 @@ end
 function Base.iterate(t::BFSIterator{<:AbstractArray})
     visited, added = falses(nv(t.graph)), falses(nv(t.graph))
     visited[first(t.source)] = false
-    state = BFSVertexIteratorState(visited, added, copy(t.source), Int[], 0, 0)
+    state = BFSVertexIteratorState(visited, added, [s for s in t.source], Int[], 0, 0)
     return Base.iterate(t, state)
 end
 
--- a/src/iterators/bfs.jl
+++ b/src/iterators/bfs.jl
@@ -42,8 +42,10 @@ in a `BitVector` so that to skip those nodes.
 """
 mutable struct BFSVertexIteratorState
     visited::BitVector
-    queue::Vector{Int}
-    neighbor_idx::Int
+    added::BitVector
+    curr_level::Vector{Int}
+    next_level::Vector{Int}
+    node_idx::Int
     n_visited::Int
 end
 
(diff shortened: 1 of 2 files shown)


Tortar:
Bump here

Tortar:
Hi @gdalle @simonschoelly sorry for the ping but I think this one would be good for a patch version release, if you have some time to spend reviewing it

test/iterators/bfs.jl
```diff
@@ -35,7 +35,7 @@
         end
     end
     nodes_visited = collect(BFSIterator(g2, [1, 6]))
-    @test nodes_visited == [1, 2, 3, 6, 5, 7, 4]
+    @test nodes_visited == [1, 6, 2, 3, 5, 7, 4]
```
gdalle, line 38: If these tests are sensitive to the ordering at each level, which is an implementation detail, can you rework them to make them independent?

src/iterators/bfs.jl
```diff
@@ -35,15 +35,16 @@ end
     BFSVertexIteratorState
 
 `BFSVertexIteratorState` is a struct to hold the current state of iteration
-in BFS which is needed for the `Base.iterate()` function. A queue is used to
-keep track of the vertices which will be visited during BFS. Since the queue
-can contains repetitions of already visited nodes, we also keep track of that
-in a `BitVector` so that to skip those nodes.
+in BFS which is needed for the `Base.iterate()` function. We use two vectors,
+one for the current level nodes and one from the next level nodes to visit 
+the graph. Since the queue can contains repetitions of already visited nodes,
```
gdalle, line 40: which vector is the "queue" here?

src/iterators/bfs.jl
```diff
@@ -35,15 +35,16 @@ end
...
 
 `BFSVertexIteratorState` is a struct to hold the current state of iteration
-in BFS which is needed for the `Base.iterate()` function. A queue is used to
-keep track of the vertices which will be visited during BFS. Since the queue
-can contains repetitions of already visited nodes, we also keep track of that
-in a `BitVector` so that to skip those nodes.
+in BFS which is needed for the `Base.iterate()` function. We use two vectors,
+one for the current level nodes and one from the next level nodes to visit 
+the graph. Since the queue can contains repetitions of already visited nodes, 
+we also keep track of that in a `BitVector` so that to skip those nodes.
```
gdalle, line 41:
```suggestion
we also keep track of that in a `BitVector` so as to skip those nodes.
```

src/iterators/bfs.jl
```diff
@@ -74,36 +77,24 @@ end
...
-            if !visited[node]
-                push!(queue, node)
-                state.visited[node] = true
-                state.n_visited += 1
-                return (node, state)
+    # we fill nodes in this level
+    if state.node_idx == length(state.curr_level)
+        state.n_visited += length(state.curr_level)
+        state.n_visited == nv(t.graph) && return nothing
+        @inbounds for node in state.curr_level
```
gdalle, line 84: if the indices were not checked for correctness, `@inbounds` is dangerous

src/iterators/bfs.jl
```diff
@@ -74,36 +77,24 @@ end
...
-            popfirst!(queue)
-            state.neighbor_idx = 1
         end
+        length(state.next_level) == 0 && return nothing
+        state.curr_level, state.next_level = state.next_level, empty!(state.curr_level)
+        state.node_idx = 0
     end
+    # we visit all nodes in this level
+    state.node_idx += 1
+    @inbounds node = state.curr_level[state.node_idx]
```
gdalle, line 98: same remark on `@inbounds`, better to remove it

gdalle (changes requested):

test/iterators/bfs.jl
Tortar: I'm a bit unsure if the ordering on each level could be considered an implementation detail, if this is it we could speed-up the running time by 2x by ordering the level nodes (for cache locality reasons I presume), but this is actually wrong because you want to follow strictly how bfs works, which is to look for each of the neighbors of a certain node in the previous level and just then go to the next node of the previous level

src/iterators/bfs.jl
Tortar: No queue anymore, my bad

src/iterators/bfs.jl
Tortar: it is checked at iterator creation, so I think it should be fine

gdalle: I don't think so because `state.curr_level` is a vector with variable length

src/iterators/bfs.jl
gdalle: how can we be sure that `state.node_ix in eachindex(state.curr_level)`?

test/iterators/bfs.jl
gdalle: If you consider two graph structures where the neighbors of each vertex make up the same (mathematical) set but are stored in different orders, the BFS algorithm will return different things for `node_visited`. And they will both be correct. So our tests should be agnostic to that

src/iterators/bfs.jl
Tortar: because it is sure that `state.node_ix < length(state.curr_level)` since we check that at the start of the function

test/iterators/bfs.jl
Tortar: yes I mean I think you are right that we should be agnostic to that, just to clarify what I mean:
          
             0 
          /      \
        1         2
       /  \     /  \
      3    4    5    6

given a graph like this if we start at node 0 it is okay to have e.g. `0, 1, 2, 3, 4, 5, 6` or `0, 2, 1, 5, 6, 3, 4` but not `0, 1, 2, 5, 6, 3, 4` or `0, 1, 2, 5, 3, 4, 6`.

But I think this what you are actually saying in your last comment, so we need to have tests which are okay with all acceptable versions.

gdalle: Exactly. It's a bit of a pain so I'm not making it strictly necessary for the PR to be merged, but essentially in your example we would want to check that the returned vector has the form `[.|..|....]` where the first subset is `{0}`, the second is `{1, 2}` and the third is `{3, 4, 5, 6}` but in any order

Tortar: that would be actually not okay though I think, because for example `0, 1, 2, 5, 6, 3, 4` is not acceptable

Tortar: Your definition iterates sligthly differently than what I had in mind...but it's totally okay I wanted just to understand which one was preferable and indeed parallelizing the algorithm effectively would require to drop mine. So let's go with yours, I think that we can also get a 2x speed-up by sorting with that :-)

gdalle: I'm slightly biased by a recent bachelor project I supervised on... parallel BFS ;) check out the repo of my interns https://github.com/KassFlute/ParallelGraphs.jl for a multithreaded and even BLAS-ified version of BFS that is much faster than the one here! ping @KassFlute and @AntoineBut

gdalle: Gentle ping @Tortar if you want to adjust the tests so that I can merge while it's still fresh in our minds

Tortar: should be okay now :+1:

gdalle (approved):

test/iterators/bfs.jl
gdalle: thanks!

PR #387, opened 2024-06-19
Implement `is_articulation` to check if vertex is articulation point

thchr:
This factors out the DFS step from `articulation` into a separate function `articulation_dfs!` and then uses this new function to implement `is_articulation(g, v)` which checks whether `v` is an articulation point of `g` without necessarily computing all the articulation points.

The branching on `isnothing(is_articulation_pts)` in `articulation_dfs!` is a bit inelegant - as is the fact that `articulation_dfs!` doesn't mutate if `is_articulation_pts === nothing`, but it seemed the simplest way to avoid code duplication.

Aside from these additions, I revised the doc string of `articulation(g)` a bit (e.g., it 1. previously misleadingly stated that `g` had to be connected; it does not; and 2. it changed nomenclature from "articulation point" to "cut vertex" mid-sentence).


thchr:
Timing-wise, `is_articulation` appears to have a speed-up of about a factor of ~2 in cases I looked at, compared to `v ∈ articulation(g)`:

```jl
julia> g = path_graph(5)
julia> articulation(g) # [2, 3, 4]
julia> @btime 2 ∈ articulation($g)   # 209.656 ns (8 allocations: 704 bytes)
julia> @btime is_articulation($g, 2) # 122.850 ns (4 allocations: 464 bytes)

julia> g = path_graph(15)
julia> articulation(g) # [2:14...]
julia> @btime 9 ∈ articulation($g)   # 453.853 ns (10 allocations: 2.16 KiB)
julia> @btime is_articulation($g, 9) # 245.286 ns (4 allocations: 624 bytes)
```

thchr:
Bump.

thchr:
Bump :)

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Graphs"
 uuid = "86223c79-3864-5bf0-83f7-82e725a168b6"
-version = "1.11.1"
+version = "1.11.2"
```
gdalle, line 3: merge default branch into this PR, the version has evolved since

src/biconnectivity/articulation.jl
```diff
@@ -22,74 +23,136 @@ julia> articulation(path_graph(5))
...
+end
+
+"""
+    is_articulation(g, v)
+
+Determine whether `v` is an
+[articulation point](https://en.wikipedia.org/wiki/Biconnected_component) of an undirected
+graph `g`, returning `true` if so and `false` otherwise.
+
+See also [`articulation`].
```
gdalle, line 46:
```suggestion
See also [`articulation`](@ref).
```

src/biconnectivity/articulation.jl
```diff
@@ -1,8 +1,9 @@
 """
     articulation(g)
 
-Compute the [articulation points](https://en.wikipedia.org/wiki/Biconnected_component)
-of a connected graph `g` and return an array containing all cut vertices.
+Compute the [articulation points](https://en.wikipedia.org/wiki/Biconnected_component) (also
+known as cut or seperating vertices) of an undirected graph `g` and return an array
```
gdalle, line 5:
```suggestion
known as cut or seperating vertices) of an undirected graph `g` and return a vector
```

src/biconnectivity/articulation.jl
```diff
@@ -22,74 +23,136 @@ julia> articulation(path_graph(5))
...
+
+julia> is_articulation(g, 1)
+false
+```
+"""
+function is_articulation end
+@traitfn function is_articulation(g::AG::(!IsDirected), v::T) where {T,AG<:AbstractGraph{T}}
+    s = Vector{Tuple{T,T,T}}()
+    low = zeros(T, nv(g))
+    pre = zeros(T, nv(g))
```
gdalle, line 72: If the idea is to call `is_articulation` several times in a row for different vertices, these allocations might be reusable? Maybe we should expose `articulation_dfs!`?

src/biconnectivity/articulation.jl
```diff
@@ -22,74 +23,136 @@ julia> articulation(path_graph(5))
...
+    return articulation_dfs!(nothing, g, v, s, low, pre)
+end
+
+@traitfn function articulation_dfs!(
+    is_articulation_pt::Union{Nothing,BitVector},
+    g::AG::(!IsDirected),
+    u::T,
+    s::Vector{Tuple{T,T,T}},
+    low::Vector{T},
+    pre::Vector{T},
```
gdalle, line 83: What do these arguments mean?

src/biconnectivity/articulation.jl
```diff
@@ -22,74 +23,136 @@ julia> articulation(path_graph(5))
...
+) where {T,AG<:AbstractGraph{T}}
+    if !isnothing(is_articulation_pt)
+        if pre[u] != 0
+            return is_articulation_pt
+        end
+    end
```
gdalle, line 89: Can you explain this shortcut in a comment?

gdalle (changes requested):
Thanks for the contribution, you were right to ping again

src/biconnectivity/articulation.jl
thchr: I admit, I don't recall: this was mainly a refactoring of the original code and I just kept the variable names as they were.

src/biconnectivity/articulation.jl
thchr: Sure, that'd be alright with me also. The `_dfs` part is an implementation detail though, so maybe this should just be called `is_articulation!`, taking work-arrays for `pre` and `low` then? Let me know what you think and I'll change to that.

src/biconnectivity/articulation.jl
gdalle: fair enough

src/biconnectivity/articulation.jl
gdalle: Sounds good, but then we should at the very least explain how to initialize these arguments (and ideally what they mean)

thchr: See also discussion in `https://github.com/JuliaGraphs/Graphs.jl/pull/407#discussion_r1851547909`: maybe we should just be calling these `work_buffer1` and `work_buffer2` (adding some comments to explain in the code - or reassigning them to more aptly named variables) to avoid exposing unnecessary complications and implementation details to users.

Krastanov:
@thchr , do you have any outstanding things on this PR that you would like to do before having it merged?

I updated to current master and ran the formatter and added it to the changelog.

thchr:
Thanks for picking this one up as well.

And no, there was no other things I wanted to do here. There was a request for a comment/explanation on some aspect of the code, but since this was effectively refactoring of an earlier implementation,  I didn't know the explanation off-hand and didn't have time to look more.

For my sake, this is good to go as-is.

PR #390, opened 2024-07-12
fix typo in vertices docs

cserteGT3:


gdalle (approved):
good catch, thank you for the fix!

PR #393, opened 2024-07-18
Fix tests on v1.11, run them on pre-release too

gdalle:
The new syntax `'lts'` and `'pre'` for GitHub actions adapts to the current LTS and pre-release versions. Pre-release is much more stable than nightly when it exists, so it's a good place to check for future issues

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,8 @@ jobs:
       matrix:
         version:
           - '1'
-          - '1.6'
+          - 'lts'
+          - 'pre'
         os:
           - ubuntu-latest
           - macos-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       matrix:
         version:
           - '1'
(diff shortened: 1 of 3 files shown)


gdalle:
JET complains because of traits

simonschoelly:
Isn't `v1.10` the new lts? In that case we should still keep the check for the lower bound version `v1.6` until we decide to raise that version in `Project.toml`.

gdalle:
@simonschoelly you're right, perhaps we can bump compat to 1.10 in this PR as well? It's not a breaking change because people on older versions of Julia won't even be able to download the new release of Graphs

gdalle:
Agreed, I'll open a separate PR to bump the compat

gdalle:
I made the JET tests slightly less sensitive (`:typo` pass instead of `:basic` pass) until we figure out the trait union split

gdalle:
@simonschoelly this is a very small change but it is necessary for the tests in 1.11 to actually be informative instead of always failing. If you have no objections I'd like to merge it?

simonschoelly (approved):
Ah yes of course

PR #397, opened 2024-09-05
Fix sanity checks for Eulerian trail

evanfields:
The sanity check for a graph having an Eulerian trail was incorrect: for a connected simple graph to have an Eulerian trail there must be exactly two odd degree vertices, but the existing code checks for exactly two even degree vertices. The current test happened to test on a 4-vertex graph with two even and two odd degree vertices, so the test passed.

I just flipped the check and added the relevant tests.


gdalle (approved):
thank you!

PR #405, opened 2024-10-21
Add algorithms for distance-regular graphs

pddshk:
Namely:

- is_distance_regular
- is_strongly_regular
- intersection_array
- global_parameters

[Wikipedia reference](https://en.wikipedia.org/wiki/Distance-regular_graph)

[NetworkX reference](https://github.com/networkx/networkx/blob/main/networkx/algorithms/distance_regular.py)

### DIFF
--- a/src/Graphs.jl
+++ b/src/Graphs.jl
@@ -437,7 +437,13 @@ export
     vertex_cover,
 
     # longestpaths
-    dag_longest_path
+    dag_longest_path,
+    
+    # distance-regular
+    is_distance_regular,
+    intersection_array,
+    global_parameters,
+    is_strongly_regular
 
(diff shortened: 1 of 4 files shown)


gdalle (changes requested):
Hi, thank you for this contribution and sorry for the delay!
I think these are useful functions, but before I review in detail, could you maybe:

- choose more specific names for `global_parameters` and perhaps `intersection_array` (perhaps prefix them with `distance_regular`
- test all the functions you introduce, including on cases where they return false
- whenever possible, avoid math in the docstrings (it doesn't display well in the REPL), replace the double backticks with single backticks and leverage unicode instead

Thanks in advance!

PR #407, opened 2024-11-14
Count number of connected components more efficiently than `length(connected_components(g))`

thchr:
This adds a new function `count_connected_components`, which returns the same value as `length(connected_components(g))` but substantially faster by avoiding unnecessary allocations. In particular, `connected_components` materializes component vectors that are not actually necessary for determining the number of components. 
Similar reasoning also lets one optimize `is_connected` a bit: did that also.

While I was there, I also improved `connected_components!` slightly: previously, it was allocating a new queue for every new "starting vertex" in the search; but the queue is always empty when it's time to add a new vertex at that point, so there's no point in instantiating a new vector.
To enable users who might want to call `connected_components!` many times in a row to reduce allocations further (I am one such user), I also made it possible to pass this queue as an optimization.

Finally, `connected_components!` is very useful and would make sense to export; so I've done that here.

Cc @gdalle, if you have time to review.

### DIFF
--- a/src/connectivity.jl
+++ b/src/connectivity.jl
@@ -1,26 +1,32 @@
 # Parts of this code were taken / derived from Graphs.jl. See LICENSE for
 # licensing details.
 """
-    connected_components!(label, g)
+    connected_components!(label, g, [search_queue])
 
 Fill `label` with the `id` of the connected component in the undirected graph
 `g` to which it belongs. Return a vector representing the component assigned
 to each vertex. The component value is the smallest vertex ID in the component.
 
+A `search_queue`, an empty `Vector{eltype(edgetype(g))}`, can be provided to reduce
+allocations if `connected_components!` is intended to be called multiple times sequentially.
+If not provided, it is automatically instantiated.
+
 ### Performance
 This algorithm is linear in the number of edges of the graph.
 """
-function connected_components!(label::AbstractVector, g::AbstractGraph{T}) where {T}
+function connected_components!(
+    label::AbstractVector, g::AbstractGraph{T}, search_queue::Vector{T}=Vector{T}()
+) where {T}
+    isempty(search_queue) || error("provided `search_queue` is not empty")
     for u in vertices(g)
         label[u] != zero(T) && continue
         label[u] = u
-        Q = Vector{T}()
-        push!(Q, u)
-        while !isempty(Q)
-            src = popfirst!(Q)
+        push!(search_queue, u)
+        while !isempty(search_queue)
+            src = popfirst!(search_queue)
             for vertex in all_neighbors(g, src)
                 if label[vertex] == zero(T)
-                    push!(Q, vertex)
+                    push!(search_queue, vertex)
                     label[vertex] = u
                 end
             end
@@ -129,9 +135,61 @@ julia> is_connected(g)
 true
 ```
 """
-function is_connected(g::AbstractGraph)
+function is_connected(g::AbstractGraph{T}) where T
     mult = is_directed(g) ? 2 : 1
-    return mult * ne(g) + 1 >= nv(g) && length(connected_components(g)) == 1
+    if mult * ne(g) + 1 >= nv(g)
+        label = zeros(T, nv(g))
+        connected_components!(label, g)
+        return allequal(label)
+    else
+        return false
+    end
+end
+
+"""
+    count_connected_components( g, [label, search_queue])
+
+Return the number of connected components in `g`.
+
+Equivalent to `length(connected_components(g))` but uses fewer allocations by not
+materializing the component vectors explicitly. Additionally, mutated work-arrays `label`
+and `search_queue` can be provided to reduce allocations further (see
+[`connected_components!`](@ref)).
+
+```
+julia> using Graphs
+
+julia> g = Graph(Edge.([1=>2, 2=>3, 3=>1, 4=>5, 5=>6, 6=>4, 7=>8]));
+
+length> connected_components(g)
+3-element Vector{Vector{Int64}}:
+ [1, 2, 3]
+ [4, 5, 6]
+ [7, 8]
+
+julia> count_connected_components(g)
+3
+```
+"""
+function count_connected_components(
+    g::AbstractGraph{T},
+    label::AbstractVector=zeros(T, nv(g)),
+    search_queue::Vector{T}=Vector{T}()
+) where T
+    connected_components!(label, g, search_queue)
+    return count_unique(label)
+end
+
+function count_unique(label::Vector{T}) where T
+    seen = Set{T}()
+    c = 0
+    for l in label
+        if l ∉ seen
+            push!(seen, l)
+            c += 1
+        end
+    end
+    return c
 end
 
 """
--- a/src/Graphs.jl
+++ b/src/Graphs.jl
@@ -210,6 +210,7 @@ export
 
     # connectivity
     connected_components,
+    count_connected_components,
     strongly_connected_components,
     strongly_connected_components_kosaraju,
     strongly_connected_components_tarjan,
--- a/src/Graphs.jl
+++ b/src/Graphs.jl
@@ -210,6 +210,7 @@ export
 
     # connectivity
     connected_components,
+    connected_components!,
     count_connected_components,
     strongly_connected_components,
     strongly_connected_components_kosaraju,
--- a/test/spanningtrees/boruvka.jl
+++ b/test/spanningtrees/boruvka.jl
@@ -21,14 +21,14 @@
         g1t = GenericGraph(SimpleGraph(edges1))
         @test res1.weight == cost_mst
         # acyclic graphs have n - c edges
-        @test nv(g1t) - length(connected_components(g1t)) == ne(g1t)
+        @test nv(g1t) - ne(g1t) == length(connected_components(g1t)) == count_connected_components(g1t)
         @test nv(g1t) == nv(g)
 
         res2 = boruvka_mst(g, distmx; minimize=false)
         edges2 = [Edge(src(e), dst(e)) for e in res2.mst]
         g2t = GenericGraph(SimpleGraph(edges2))
         @test res2.weight == cost_max_vec_mst
-        @test nv(g2t) - length(connected_components(g2t)) == ne(g2t)
+        @test nv(g2t) - ne(g2t) == length(connected_components(g2t)) == count_connected_components(g2t)
         @test nv(g2t) == nv(g)
     end
     # second test
@@ -60,14 +60,14 @@
         edges3 = [Edge(src(e), dst(e)) for e in res3.mst]
         g3t = GenericGraph(SimpleGraph(edges3))
         @test res3.weight == weight_vec2
-        @test nv(g3t) - length(connected_components(g3t)) == ne(g3t)
+        @test nv(g3t) - ne(g3t) == length(connected_components(g3t)) == count_connected_components(g3t)
         @test nv(g3t) == nv(gx)
 
         res4 = boruvka_mst(g, distmx_sec; minimize=false)
         edges4 = [Edge(src(e), dst(e)) for e in res4.mst]
         g4t = GenericGraph(SimpleGraph(edges4))
         @test res4.weight == weight_max_vec2
-        @test nv(g4t) - length(connected_components(g4t)) == ne(g4t)
+        @test nv(g4t) - ne(g4t) == length(connected_components(g4t)) == count_connected_components(g4t)
         @test nv(g4t) == nv(gx)
     end
 
@@ -123,14 +123,14 @@
         edges5 = [Edge(src(e), dst(e)) for e in res5.mst]
         g5t = GenericGraph(SimpleGraph(edges5))
         @test res5.weight == weight_vec3
-        @test nv(g5t) - length(connected_components(g5t)) == ne(g5t)
+        @test nv(g5t) - ne(g5t) == length(connected_components(g5t)) == count_connected_components(g5t)
         @test nv(g5t) == nv(gd)
 
         res6 = boruvka_mst(g, distmx_third; minimize=false)
         edges6 = [Edge(src(e), dst(e)) for e in res6.mst]
         g6t = GenericGraph(SimpleGraph(edges6))
         @test res6.weight == weight_max_vec3
-        @test nv(g6t) - length(connected_components(g6t)) == ne(g6t)
+        @test nv(g6t) -  ne(g6t) == length(connected_components(g6t)) == count_connected_components(g6t)
         @test nv(g6t) == nv(gd)
     end
 end


thchr:
For the doctest example of ` g = Graph(Edge.([1=>2, 2=>3, 3=>1, 4=>5, 5=>6, 6=>4, 7=>8]))`, `count_connected_components` is about twice as fast as `length∘connected_components` (179 ns vs. 290 ns). Using the buffers, it is faster still (105 ns).

src/connectivity.jl
```diff
@@ -1,26 +1,32 @@
 # Parts of this code were taken / derived from Graphs.jl. See LICENSE for
 # licensing details.
 """
-    connected_components!(label, g)
+    connected_components!(label, g, [search_queue])
```
simonschoelly, line 4: I am all for performance improvements. But I am a bit skeptical if it is worth making the interface more complicated.

Almost all graph algorithms need some kind of of work buffer, so we could have something like in al algorithms but in the end it should be the job for Julia's allocator to verify if there is some suitable piece of memory lying around. We can help it by using `sizehint!` with a suitable heuristic.

src/connectivity.jl
```diff
@@ -129,9 +135,69 @@ julia> is_connected(g)
...
+3-element Vector{Vector{Int64}}:
+ [1, 2, 3]
+ [4, 5, 6]
+ [7, 8]
+
+julia> count_connected_components(g)
+3
+```
+"""
+function count_connected_components(
```
simonschoelly, line 179: I am a bit undecided if we should call this `count_connected_components` or `num_connected_components`. Currently we have both conventions, namely `num_self_loops` and `Graphs.Experimental.count_isomorph`.

Ideally we use the same word everywhere. @gdalle Do you have an opinion on that?

src/connectivity.jl
```diff
@@ -129,9 +135,69 @@ julia> is_connected(g)
...
+function count_unique(label::Vector{T}) where {T}
+    seen = Set{T}()
+    c = 0
+    for l in label
+        if l ∉ seen
+            push!(seen, l)
+            c += 1
+        end
+    end
+    return c
```
simonschoelly, line 200:
```suggestion
    return length(Set(label))
```

src/connectivity.jl
```diff
@@ -1,26 +1,32 @@
...
+If not provided, it is automatically instantiated.
+
 ### Performance
 This algorithm is linear in the number of edges of the graph.
 """
-function connected_components!(label::AbstractVector, g::AbstractGraph{T}) where {T}
+function connected_components!(
+    label::AbstractVector{T}, g::AbstractGraph{T}, search_queue::Vector{T}=Vector{T}()
+) where {T}
+    isempty(search_queue) || error("provided `search_queue` is not empty")
```
simonschoelly, line 20: If we pass in `search_queue`  to increase performance then it gets modified anyways. So instead of requiring that it is empty we can either call `empty!` on it. Calling `empty!` does not always (or maybe even never) deallocate the memory buffer that is associated with a `Vector` - so `push!` would not do unnecessary allocations when that vector is filled again.

simonschoelly:
> For the doctest example of ` g = Graph(Edge.([1=>2, 2=>3, 3=>1, 4=>5, 5=>6, 6=>4, 7=>8]))`, `count_connected_components` is about twice as fast as `length∘connected_components` (179 ns vs. 290 ns). Using the buffers, it is faster still (105 ns).

We should not do benchmarks on such small graphs unless the algorithm has a huge complexity and is slow even on very small graphs. Otherwise the benchmark is way too noisy and also does not really reflect the situations where this library is used.

src/connectivity.jl
thchr: I agree that this will usually not be relevant; in my case it is though, and is the main reason I made the changes. I also agree that there is a trade off between performance improvements and complications of the API. On the other hand, I think passing such work buffers as _optional_ arguments is a good solution to such trade-offs: for most users, the complication can be safely ignored and shouldn't complicate their lives much.

As you say, there are potentially many algorithms in Graphs.jl that _could_ take a work buffer; in light of that, maybe this could be more palatable if we settle on a unified name for these kinds of optional buffers, so that it lowers the complications by standardizing across methods.
Maybe just `work_buffer` (and, if there are multiple, `work_buffer1`, `work_buffer2`, etc?)

src/connectivity.jl
thchr: There's also `nv(g)` for the number of vertices. Maybe just `nconnected_components`?

src/connectivity.jl
thchr: Agree, yeah, that's neater. Done.

src/connectivity.jl
thchr: That's less performant than the explicit looped version though:

```jl
julia> label_small = rand(1:3, 20)
julia> @b count_unique($label_small)
150.851 ns (4 allocs: 320 bytes) # loop
174.412 ns (4 allocs: 464 bytes) # length(Set(label))

julia> label_big = rand(1:50, 5000)
julia> @b count_unique($label_big)
23.385 μs (11 allocs: 3.312 KiB) # loop
32.719 μs (6 allocs: 72.172 KiB) # length(Set(label))

julia> label_huge = rand(1:5000, 500000)
julia> @b count_unique($label_huge)
3.499 ms (25 allocs: 192.625 KiB) # loop
4.876 ms (6 allocs: 9.000 MiB, 2.51% gc time)  # length(Set(label))
```

thchr:
> We should not do benchmarks on such small graphs unless the algorithm has a huge complexity and is slow even on very small graphs. Otherwise the benchmark is way too noisy and also does not really reflect the situations where this library is used.

What are some good go-to defaults for testing? This is a thing I'm running up against frequently, I feel: I am not sure which graphs to test against, and anything beyond small toy examples are not easily accessible via convenience constructors in Graphs.

As context, in my situation the graphs are rarely larger than 50-100 vertices; my challenge is that I need to consider a huge number of permutations of such graphs, so performance in the small-graph case is relevant to me.

src/connectivity.jl
thchr: It's indeed not very great that the `length(Set(label))` version is slower though. The reasons seems to be that `Set(itr)` is assuming that most elements in `itr` will be unique and goes ahead an `sizehint!`s the to-be-filled-in `Set` to be the full length of `itr` - but that seems very unlikely to ever be the case in this scenario: there will usually be far fewer connected components than vertices.

thchr: Actually, it is not really an "issue" in Base, per se: rather, it seems `Set` is optimized with the assumption that most things that are `push!`ed into it are new, unique things. But when that assumption doesn't apply, it is faster to check `∉` before trying to `push!`. Here, I would say it's very safe to assume that `label` will usually contain far fewer unique things than its length, so we might as well exploit that.

src/connectivity.jl
gdalle: If I had to pick I'd rather use `count` than `num` or `n` because it is a complete word

src/connectivity.jl
gdalle: If we do this then all functions should take exactly one `work_buffer` (possibly a tuple) and have an appropriate function to initialize the buffer. I think it is a major change which should be discussed separately.
What is your use case @thchr ?

gdalle: - https://github.com/JuliaGraphs/Graphs.jl/issues/410

gdalle:
> What are some good go-to defaults for testing? This is a thing I'm running up against frequently, I feel: I am not sure which graphs to test against, and anything beyond small toy examples are not easily accessible via convenience constructors in Graphs.

I have opened this issue to discuss further:

- What are some good go-to defaults for testing? This is a thing I'm running up against frequently, I feel: I am not sure which graphs to test against, and anything beyond small toy examples are not easily accessible via convenience constructors in Graphs.

src/connectivity.jl
simonschoelly: Definitely no to `nconnected_components` - `nv` and `ne` might be some exceptions as they are used all the time - but we might rename them one day.

I don't mind abbreviation from time to time, but lets go with `count_connected_components` then - after all we also have a `count` function in the Julia base.

src/connectivity.jl
simonschoelly: That's interesting. I did not know that. Btw. if try to be really efficient here - would using `BitSet` instead of `Set` be even more efficient?

src/connectivity.jl
simonschoelly: So I think if this is really important for your use case you can either
- Create a version that uses a buffer in the `Experimental` submodule. Currently we don't guarantee semantic versioning there - this allows use to remove things in the future without breaking the API.
- Or as this code is very simple you might just copy it to your own repository.


But just to clarify - your problem is not that you are building graphs by adding edges until they are connected? Because if that is the issue, there is a much better algorithm.

src/connectivity.jl
thchr: Indeed, that is a good chunk faster (but only usable for `T = Int`), even for vectors whose elements are more sparsely distributed than its length. For the above examples, it's 2-6 times faster than `Set`. I've swapped to `BitSet` for `T=Int`.

src/connectivity.jl
thchr: >Create a version that uses a buffer in the Experimental submodule. Currently we don't guarantee semantic versioning there - this allows use to remove things in the future without breaking the API.

I won't be able to find the time for factoring this out into the Experimental submodule, unfortunately.
I'm happy to e.g., add an admonition to the docstring, indicating that the work buffer arguments are unstable API which are subject to breakage though. Factoring this into a submodule, piecing it back together, and adding multiple doc-strings across modules, eventually loading this behind a `Graphs.Experimental` call is more fiddling than I'm up for.

> Or as this code is very simple you might just copy it to your own repository.

Indeed, I have and will just continue to do that, yep.

> But just to clarify - your problem is not that you are building graphs by adding edges until they are connected?  Because if that is the issue, there is a much better algorithm.

No, I'm not doing that; appreciate the check though.

---

As a general side note - and please know that I appreciate these reviews very much (!) and your efforts on what is no doubt spare time (!!) - but I wonder if the level of scrutiny and optimization that many PRs here go through is optimal: I understand the intent and the aim of making stable API and of getting good, maintainable code. But I think there's a risk of trading off too much towards these goals, at the cost of vibrancy and community engagement. From my experience here, there's room for leaning more towards "is this better than what we previously had" over "could this be even better".

PRs like this usually happen on time "stolen away" from our day-jobs, and that the odds of returning to PRs for edits, however small, go down very quickly with time; similarly, the expectation that a PR will be a multi-iteration process reduces the likelihood it will be made.

gdalle: I appreciate the feedback, and I lean in the same direction. The JuliaGraphs community calls have pretty much dried up recently and we don't have much of a community to start with, so it's in our best interest to make engagement rewarding instead of tiresome.
I'm expecting an answer this month on funding I applied to which could serve to revitalize this ecosystem, hopefully I'll have good news to share soon.

rafaqz: > but in the end it should be the job for Julia's allocator to verify if there is some suitable piece of memory lying around. 

After hitting similar problems to op I feel the need to point out that this statement is a bit problematic as a design philosophy in Julia. 

Julias garbage collector is slow, and in practice it's on us as developers not to overwork it constantly. Base julia lets use reuse workspaces in `sort` and LinearAlgebra solves, and the many in-place methods.

For me the shortest paths algorithms here are barely useable at production scale (milllions of calls) due to the allocation and garbage collector overheads. I can also just write fast custom methods of everything that our project needs, but the end point of us all doing that is useful code duplicated and scattered through field-specific tools, rather than centralized and usable to the ecosystem.

But I think there needs to be a broad policy to keep things coherent, like: mutated workspaces are either the first argument with a `!` method if they are also the return value (as I would do with `Graphs.enumerate_paths!`) or keywords like in `Base.sort`.

But given that kind of framework PRs adding workspace keywords and ! mutating versions shouldn't be too controversial.

Krastanov:
@thchr , it has been a while since the discussion on this PR has died out.

- What is your stance on this PR -- do you still care for it? If you have gone on to solve your problem in a different way, could you share what you are using instead of Graphs.jl
- Could you confirm that the only outstanding issue is the discussion about whether it is ok to introduce the use of a scratch space in the public API?

If that is indeed the only outstanding issue, I would like to keep it as it is and move forward with rebasing/testing/merging this. It would be nice to clean up the backlog of PRs.

thchr:
> @thchr , it has been a while since the discussion on this PR has died out.
> 
> * What is your stance on this PR -- do you still care for it? If you have gone on to solve your problem in a different way, could you share what you are using instead of Graphs.jl
> * Could you confirm that the only outstanding issue is the discussion about whether it is ok to introduce the use of a scratch space in the public API?
> 
> If that is indeed the only outstanding issue, I would like to keep it as it is and move forward with rebasing/testing/merging this. It would be nice to clean up the backlog of PRs.

Hi @Krastanov - thanks for having a look at this, and for your efforts here more broadly!

I still think it's a useful addition: I use it myself, but have so far simply duplicated the definitions under new names ([here](https://github.com/thchr/BandGraphs.jl/blob/main/src/count_connected_components.jl)). I have since optimized the methods a bit more (e.g., [here](https://github.com/thchr/BandGraphs.jl/blob/29b2036d3709d3810b4d70b549663e30cbeaa127/src/count_connected_components.jl#L102-L202)), but it's more complicated and probably not worthwhile porting & restarting review here.

The outstanding issue was indeed whether or not it was OK to introduce scratch space. The bigger question around that got transferred into issue #410, but no consensus was reached afaik. I think the idea of having a scratch/buffer style like what I proposed here https://github.com/JuliaGraphs/Graphs.jl/issues/410#issuecomment-2882936463 is sound, but it's not implemented here (for lack of a clear consensus in #410).

We could merge as-is, simply stating that the use of the scratch space buffers is experimental and subject to change?

Krastanov:
Thank you for getting back to this after a long wait! Let's indeed merge it. I asked my instance of claude to add a quick changelog update and to edit the docstring:

image

Could you check whether the incoming edits are fitting your preferences. If they do, I will mark this as approved and we can merge.

By the way, you probably have noticed that we are short on volunteers here. Would you be interested in being added to the github org for Graphs.jl to be notified about the occasional PRs, in case you have the bandwidth to review (once or twice a year would already be a big help)?

thchr:
Thanks for the Claude-edit; I need to play with that myself - very handy..!

I've made a minor revision to the experimental warning language: it's good to go from my side now.

thchr:
> By the way, you probably have noticed that we are short on volunteers here. Would you be interested in being added to the github org for Graphs.jl to be notified about the occasional PRs, in case you have the bandwidth to review (once or twice a year would already be a big help)?

Oh, forgot this: yeah, I'd be happy to lend a hand when I have the bandwidth and know-how.

Krastanov:
Thank you! I will add you to the github org. A lot of the volunteers are also on the julia slack in the #graphs channel. Funnily, I was just meeting with Mikkel while you were updating the PR this morning -- just learnt you are working together. Funny how small the world is.

PR #415, opened 2025-01-10
Fix error in implementation of Erdos-Gallai condition in isgraphical

### LINKED ISSUE #400: [BUG] `isgraphical` returns true for a non-graphical sequence
**Description of bug, expected and actual behavior**
`isgraphical` returns true for a non-graphical sequence, namely `[4,2,2,2,0]`. It should return false on this input.

**How to reproduce**
Call the function `isgraphical` on input `[4,2,2,2,0]`.

**Code demonstrating bug**
```julia
using Graphs

println(isgraphical([4,2,2,2])) # should print false, prints false
println(isgraphical([4,2,2,2,0])) # should print false, prints true
println(isgraphical([4,2,2,2,0]) == isgraphical([4,2,2,2])) # should definitely print true, prints false
```

**Version information**
```
Julia Version 1.10.4
Commit 48d4fd4843 (2024-06-04 10:41 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 16 × AMD Ryzen 7 5800H with Radeon Graphics
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, znver3)
Threads: 1 default, 0 interactive, 1 GC (on 16 virtual cores)   
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS =
```
Output of `]status Graphs` is `[86223c79] Graphs v1.9.0`. 

**Additional context**
I at first thought the implementation allowed self-loops, in which case the sequence can be realized by (for example) the graph on vertices a,b,c,d,e with edges a-a, a-b, b-c, a-c, a-d, d-d. 
However, [4,2,2,2] returns false (as it should if we disallow self-loops). 
The implementation should be consistent (optionally with a flag to allow or disallow self-loops, if this is a needed feature). 

If the bug is limited to cases where there is a vertex of degree zero, an easy fix would be to add the line
```julia
degs = [deg for deg in degs if deg > 0]
```
at the beginning of the function (e.g. after checking the sum of degrees is even, and before the sequence length is computed).

Syuizen:
Fixes #400

Fix `isgraphical` function to correctly handle non-graphical sequences.

# Error in current implementation

`mindeg` is computed globally at the start using `min(i, sorted_degs[i])` for all indices. However, the Erdös-Gallai condition requires dynamically calculating `min(r, sorted_degs[i])` for vertices after the current index `r`.

# What changed

* Update the Erdös-Gallai condition check to calculate the sum of the minimum of r and the degrees of the vertices.

* Add a test case in `test/connectivity.jl` to verify that `isgraphical` returns false for the sequence [4,2,2,2,0].

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/JuliaGraphs/Graphs.jl/pull/415?shareId=0624035b-53a0-44f3-a33a-03d1ce4ad531).

### DIFF
--- a/src/connectivity.jl
+++ b/src/connectivity.jl
@@ -798,19 +798,49 @@ function isgraphical(degs::AbstractVector{<:Integer})
     !isempty(degs) || return true
     # Check whether the sum of degrees is even
     iseven(sum(degs)) || return false
-    # Check that all degrees are non negative and less than n-1
+    # Compute the length of the degree sequence
     n = length(degs)
+    # Check that all degrees are non negative and less than n-1
     all(0 .<= degs .<= n - 1) || return false
     # Sort the degree sequence in non-increasing order
     sorted_degs = sort(degs; rev=true)
-    # Compute the length of the degree sequence
+    # Initialise a sum variable
     cur_sum = zero(UInt64)
+    right_deg_sum = zero(UInt64)
+    # Initalise a pointer to track the smallest index with degree greater than r
+    ptr = n
     # Check if the degree sequence satisfies the Erdös-Gallai condition
     @inbounds for r in 1:(n - 1)
         cur_sum += sorted_degs[r]
-        # Calculate the sum of the minimum of r and the degrees of the vertices
-        mid_deg_sum = sum([min(r, sorted_degs[i]) for i in (r + 1):n])
-        cond = cur_sum <= (r * (r - 1) + mid_deg_sum)
+        #  Calculate the sum of the minimum of r and the degrees of the vertices
+        min_idx = r + 1
+        while ptr >= min_idx
+            if sorted_degs[ptr] <= r
+                # left_deg_sum = sum_{ptr+1}^n d_i
+                right_deg_sum += sorted_degs[ptr]
+                # move pointer to the 1-slot left
+                ptr -= 1
+            else 
+                # the ptr points to the degree greater than r
+                break
+            end
+        end
+        # calculate min_deg_sum: sum_{r+1}^n min(r, d_i)
+        if ptr < min_idx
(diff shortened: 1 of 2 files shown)


simonschoelly:
Thank you. I think this is the second bug in this algorithm in a short while. I think we might have mixed up the indices for the summation on the right hand side of the formula.

Syuizen:
Thanks for the advice, I have implemented a better algorithm which (should) be O(n*log(n)) complexity now.

simonschoelly:
Your code looks good but we have some failing tests. I think they don't have anything to with your changes though. @gdalle do you have an idea what is going on?

gdalle:
Run JuliaFormatter and try again?

simonschoelly:
> Run JuliaFormatter and try again?

Right, that seems to fix it. I was confused because I thought the issue was in the `Experimental` submodule but that seems to be just a test that is marked as broken.

@Syuizen The issue seems to be something on line `823` - you can either fix it manually or simply run `JuliaFormatter` for example like this:

First go into the `Graphs` directory and start Julia then
```julia
]activate --temp
]add JuliaFormatter
using JuliaFormatter
format(".")
```
then commit the changes.

simonschoelly (approved):

simonschoelly:
Thanks @Syuizen . Do you nee a new tagged version immediately? Otherwise it will be released with next patch.

Syuizen:
Awesome! Thanks, Simon. No, I didn’t have any urgent needs right now. I
plan to continue making more contributions into the Graphs.jl :)

PR #417, opened 2025-02-16
minor addition to `edges` docstring

Krastanov:
Seeing that `edges` is lexicographical took me a minute longer than necessary, so I thought this small addition is warranted.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -183,7 +183,7 @@ vertices(g::AbstractGraph) = _NI("vertices")
     edges(g)
 
 Return (an iterator to or collection of) the edges of a graph.
-For `AbstractSimpleGraph`s it returns a `SimpleEdgeIter`.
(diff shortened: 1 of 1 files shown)


simonschoelly:
I am not sure that we currently guarantee this for any `AbstractSimpleGraph`. Having this as part of the abstract interface might make some implementations slow as they will have to sort the edges first.

Perhaps we should only guarantee this for `SimpleGraph` and `SimpleDiGraph`? Or is it important that this is also the case for other graphs in your case?

Krastanov:
It is guaranteed in another part of the documentation. This PR is meant to simply make that guarantee easier to find for the reader. 
image

Krastanov:
but to answer your question, no the general case is not important for me

simonschoelly (approved):

simonschoelly:
In that case it makes sense. Thanks.

PR #418, opened 2025-02-26
Graph isomorphism and canonization via NautyGraphs.jl

mxhbl:
Following the suggestion of @Krastanov, this PR adds [NautyGraphs.jl](https://github.com/mxhbl/NautyGraphs.jl) as a weakdep, which makes it possible to use [nauty](https://pallini.di.uniroma1.it) for graph isomorphism and canonization.

### Main Changes
- New function `Graphs.Experimental.canonize!(g, alg)` that permutes a graphs vertices into the canonical order defined by the algorithm `alg`. If two graphs `g1` and `g2` are isomorphic, `g1 == g2` after canonization.
- `NautyGraphsExt` package extension, which implements `canonize!` and `has_isomorph` with NautyGraphs as a backend. Other isomorphism algorithms are currently not supported, and throw an error if called with `alg=AlgNautyGraphs()`. Vertex and edge relations are currently also not supported, but could in principle be implemented by cleverly transforming the input graph.
- Since `canonize!` permutes a graph's vertices in-place, I also implemented a function `permute!(g::Graph, p)` that performs the permutation.

### Other Changes
- Calling any function in `Graphs.Experimental` with an algorithm that doesn't define an implementation method results in a stack overflow. Since the dispatch struct `AlgNautyGraphs` needs to be define inside the main Graphs module and not in the package extension, this would cause stack overflow errors if the user tries to call e.g. `has_isomorph(g, h, AlgNautyGraphs())` without loading NautyGraphs first. I changed this to manually throw method errors in the fallback methods, which leads to nicer error messages for the user.

### Tests?
I am not sure how to best test the isomorphism and canonization, both in Graphs.jl, but also in NautyGraphs.jl itself. Are there any standard test sets of graphs that are not too hard to check but also nontrivial? Should I test NautyGraphs against VF2?


src/Experimental/isomorphism.jl
```diff
@@ -87,9 +87,7 @@ function has_induced_subgraphisomorph(
     vertex_relation::Union{Nothing,Function}=nothing,
     edge_relation::Union{Nothing,Function}=nothing,
 )::Bool
-    return has_induced_subgraphisomorph(
-        g1, g2, alg; vertex_relation=vertex_relation, edge_relation=edge_relation
-    )
+    throw(MethodError(has_induced_subgraphisomorph, (g1, g2, alg)))
```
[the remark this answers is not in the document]
Krastanov: also, generally, I try to avoid throwing my own MethodErrors. Structuring the dispatch such that julia is the one the throw MethodErrors usually leads to simpler code (and then error hints can be used for more messages).

src/Experimental/isomorphism.jl
```diff
@@ -141,12 +137,14 @@ end
 Return `true` if the graph `g1` is isomorphic to `g2`.
 
 ### Optional Arguments
-- `alg`: The algorithm that is used to find the induced subgraph isomorphism. Can be only
-    `VF2()` at the moment.
+- `alg`: The algorithm that is used to find the induced subgraph isomorphism. Can be
+    `VF2()` or `AlgNautyGraphs()`, if `NautyGraphs` is installed and imported.
```
Krastanov, line 141:
```suggestion
    `VF2()` or, if `NautyGraphs` is imported, `AlgNautyGraphs()`.
```

src/Experimental/nautygraphs.jl
```diff
@@ -0,0 +1,8 @@
+"""
+    AlgNautyGraphs
+
+An empty concrete type used to dispatch to [`NautyGraphs`](@ref) isomorphism functions.
```
Krastanov, line 4:
```suggestion
A "configuration" type used to dispatch to isomorphism algorithms from the `nauty_graphs` C library,
wrapped by the `NautyGraphs.jl` package.
```

src/operators.jl
```diff
@@ -104,6 +104,18 @@ function reverse! end
     return g
 end
 
+"""
+    permute!(g, p)
+
+Permute the vertices of graph `g` in-place, according to permutation `p`. No checking is done to verify that p is a permutation.
+"""
+function permute! end
```
Krastanov, line 112: this should probably not be a new function, just a new method for the existing `Base.permute!` function

src/Experimental/isomorphism.jl
```diff
@@ -173,9 +171,40 @@ function has_isomorph(
...
+[4, 1, 2, 3]
+julia> g1 == g2
+true
+```
+### See also 
+[`has_isomorph`](@ref)
+"""
+function canonize!(g::AbstractGraph, alg::IsomorphismAlgorithm=AlgNautyGraphs())
+    throw(MethodError(canonize!, (g, alg)))
 end
```
Krastanov, line 208: this can be `function canonize! end` so that we do not manually throw `MethodError`. Then we can have a hint that detects the method error and warns that one needs to import NautyGraphs.

src/Experimental/isomorphism.jl
```diff
@@ -173,9 +171,40 @@ function has_isomorph(
...
+
+Permute the vertices of graph `g` into the canonical order defined by the algorithm `alg` and return the permutation. 
+If graphs `g1` and `g2` are isomorphic, the orders of their vertices will be equal after canonizing them with the same algorithm.
+
+### Optional Arguments
+- `alg`: The algorithm that is used to canonize the graph. Can be only be `AlgNautyGraphs()`
+   at this moment, which requires `NautyGraphs` to be installed and imported.
+
+### Examples
+```doctest.jl
```
Krastanov, line 188: huh... I did not know that you can also use `doctest.jl` as a specifier for doctest!

Krastanov (changes requested):
Thank you for setting this up! I can hopefully help with the first couple of rounds of review before this gets to the core maintainers.

This is pretty close to what I have in mind, I really appreciate you put in the time to prepare it. I think my main suggestion is that this can be a bit simpler if we remove most of the `MethodError` and `error` calls and just rely on Julia to raise the correct MethodError (that we improve with registering hints). I left a few more detailed comments below, but if something is unclear I would be happy to prototype a few more specific examples.

I put in a few suggestions for rewording the docs, but they are just personal preferences.

simonschoelly:
NautyGraphs.jl is interesting, but before you put too much effort into this PR, I don't see what we gain by making an extension in Graphs.jl? NautyGraphs.jl depends on Graphs.jl, so we can just import the symbols there and create concrete implementations there?

Krastanov:
> NautyGraphs.jl depends on Graphs.jl, so we can just import the symbols there and create concrete implementations there?

You are right about not needing the extension itself, that is my fault for misleading @mxhbl . I did not catch that the dependency already existed, rather silly of me. Apologies for that! Either way, three points to specify though:

- AlgNautyGraphs existing in this repo would make discoverability much nicer
- Error hints provided here, about the need to import NautyGraphs would also make discoverability easier
- `canonize!` coming from Graphs and just being extended by NautyGraphs, as you mentioned, would simplify organization

mxhbl:
You are right that the package extension is not required. But I do agree with @Krastanov that making `canonize!` a function in `Graphs` might be a nice addition. Especially if there are multiple algorithms that could do the canonization. I am not super familiar with VF2, but if there is a way to perform graph canonization with it I would be happy to implement a version of `canonize!` in `Graphs` that defaults to VF2 and can then be extended in `NautyGraphs`.

simonschoelly:
> You are right that the package extension is not required. But I do agree with @Krastanov that making `canonize!` a function in `Graphs` might be a nice addition. Especially if there are multiple algorithms that could do the canonization. I am not super familiar with VF2, but if there is a way to perform graph canonization with it I would be happy to implement a version of `canonize!` in `Graphs` that defaults to VF2 and can then be extended in `NautyGraphs`.

I actually asked myself the same question a few weeks ago - it looks like there is no such things for VF2 or any of its extensions such as VF2+, VF3. But these algorithms are basically just simple search algorithms that explore the tree of all possible matchings between two graphs and use some pruning heuristics to speed up the process.

The tools/algorithms listened in this [survey](https://arxiv.org/pdf/1301.1493) are:
- nauty
- traces
- bliss
- saucy
- conauto

I did not have the time to look at them yet and see if one would be easy to implement here.

There is also this [never result](https://par.nsf.gov/servlets/purl/10179675) by László Babai - but it seems to be of theoretical nature only - and I am not even sure if the results there have been definitely confirmed.

mxhbl:
Thanks for the list. I was also not able to find pre-made approaches to canonization with VF2 unfortunately.

It's my understanding that all of the algorithms you listed rely complex heuristics to achieve good performance and would take a lot of effort to reimplement here. E.g. nauty and traces each have >10,000 lines of code.

Krastanov:
@mxhbl , would you be interested in setting up a bounty (for you, for a coworker, for a junior dev/student that you know or just someone from the community) in the style of this one https://github.com/JuliaGraphs/Graphs.jl/issues/448 (but for NautyGraphs.jl)

mxhbl:
@Krastanov that sounds great, I'd be more than happy to help in any way I can. What exactly did you have in mind?

Krastanov:
I think the following is a good list of requirements, but please feel free to suggest changes or other approaches. Some things might already be available:

- `Graph(::NautyGraph)` and `NautyGraph(::Graph)` should be fast and if the internal layout permits, creating non-copying constructors should be possible
- The basic Graphs API should be supported by `NautyGraph`, basically what is shown here https://juliagraphs.org/Graphs.jl/stable/core_functions/interface/
- after that is vetted, making sure that these functions "just work" https://juliagraphs.org/Graphs.jl/stable/core_functions/core/
- some documentation for the fact that NautyGraphs is available in a way the extends the capabilities of Graphs
- and lastly, introducing the interesting capabilities of nauty graphs. This has two components:
    - if there is a function `Graphs.some_expensive_computation(::Graph)` that is also implemented in nauty graphs, then a method for it should be defined in NautyGraphs `some_expensive_computation(::Graph, NautyGraphs.NautyAlg())`
    - if there is an interesting algorithm available in NautyGraphs but not available in Graphs, then an empty function declaration should be added to Graphs (together with a good error hint) and a method for that new function goes in NautyGraphs

The very last subpoint is something that will need to be approved by the community though, as it is an actual change to Graphs.

Krastanov:
another thing to add to the list:
- use GraphsInterfaceChecker.jl in the test suite

mxhbl:
Thats sounds great. I think a lot of the Graphs API should already be supported, but test and general polish are badly needed. This would also be a great opportunity to add support for `nauty`s sparse graph format, which probably can be converted to and from a `SimpleGraph` without copies. I'd love to work on this and I think your review would be very beneficial! What is the standard way to manage/track the workflow with this kind of project? Should I open an issue in NautyGraphs?

Krastanov:
We do not really have a standard workflow for this. How about the following: copy one of the existing bounties for the other libraries, flesh it out with the TODOs you and I listed here, post it on the Graphs.jl issue tracker, and if it is something that I can comfortably request from the sponsor, I will just mark it as a bounty and mark it as reserved for you.

mxhbl:
I created an issue (#452) with the points we discussed. Let me know if it looks good to you!

Krastanov:
great! Marked as a reserved bounty

Krastanov:
Closing this as it is mostly done inside of NautyGraphs.jl now. See this comment on things left to do https://github.com/JuliaGraphs/Graphs.jl/issues/452#issuecomment-3941319021

PR #427, opened 2025-04-08
Update randgraphs.jl

javier-gallardo-saenz:
Added Hierarchical Configuration Model to randgraphs.jl

I have tested its performance for small/medium numbers of nodes and communities, but further testing might be necessary.

This implementation allows self-loops and multiple edges (which are reduced here to a simple edge), so the resulting degree sequences might not exactly matched the specified sequences. Thus, I am wondering if this function might be better suited for Multigraph.jl, but since the repository includes the Configuration Model I figured it would make sense that it included the Hierarchical Configuration Model as well (as this is basically its extension to community structure). 

Some external degree sequences cannot be realized even if they add up to an even number. To detect such sequences is not trivial, so when a external degree sequence is 'correct' (adds up to even number) but cannot be realized the function should return the closest possible realization. 

Best regards,
Javier

### DIFF
--- a/src/SimpleGraphs/generators/randgraphs.jl
+++ b/src/SimpleGraphs/generators/randgraphs.jl
@@ -1559,3 +1559,104 @@ function random_orientation_dag(
     end
     return g2
 end
+
+"""
+    random_hierarchical_configuration_model(in_degrees::Vector{Int},
+                                             ext_degrees::Vector{Int},
+                                             community_sizes::Vector{Int};
+                                             seed::Union{Nothing, Int}=nothing)
+
+Generates a random graph using the Hierarchical Configuration Model.
+This is a modified configuration model that incorporates community structure.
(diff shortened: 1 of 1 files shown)


Krastanov:
Thanks, @javier-gallardo-saenz for this submission and apologies for letting it fall between the cracks and not having it reviewed.

It seems right now there is a syntax error -- a missing `end` statement in the newly defined function. If you have the energy to get back to it, the community would certainly be grateful. For the moment I will mark this as a draft to have my review queue a bit more organized, but do not hesitate to mark it back to not-a-draft.

If you get back to this, do not hesitate to bump us repeatedly for reviews -- we are pretty short staffed on volunteers, so it might take us a bit to remember to get to this. Speaking of which, we are struggling to get enough volunteer maintainers, so let me know if you are interested in helping with reviews -- even once a year can be quite valuable.

PR #429, opened 2025-05-14
Add missing `parallel=:threads` implementations

topolarity:
These implementations are basic, but I tried to follow the patterns in the other parts of the Parallel module.

My real motivation is to be able to move the Distributed implementations into an extension, so that Graphs.jl does not depend on Distributed.

### DIFF
--- a/src/Parallel/distance.jl
+++ b/src/Parallel/distance.jl
@@ -1,20 +1,55 @@
 # used in shortest path calculations
 
 function eccentricity(
+    g::AbstractGraph,
+    vs=vertices(g),
+    distmx::AbstractMatrix{T}=weights(g);
+    parallel::Symbol=:distributed,
+) where {T<:Number}
+    return if parallel === :threads
+        threaded_eccentricity(g, vs, distmx)
+    elseif parallel === :distributed
+        distr_eccentricity(g, vs, distmx)
+    else
+        throw(
+            ArgumentError(
+                "Unsupported parallel argument '$(repr(parallel))' (supported: ':threads' or ':distributed')",
+            ),
+        )
+    end
+end
+
+function distr_eccentricity(
     g::AbstractGraph, vs=vertices(g), distmx::AbstractMatrix{T}=weights(g)
 ) where {T<:Number}
     vlen = length(vs)
     eccs = SharedVector{T}(vlen)
     @sync @distributed for i in 1:vlen
-        eccs[i] = maximum(Graphs.dijkstra_shortest_paths(g, vs[i], distmx).dists)
+        local d = Graphs.dijkstra_shortest_paths(g, vs[i], distmx)
+        eccs[i] = maximum(d.dists)
     end
     d = sdata(eccs)
     maximum(d) == typemax(T) && @warn("Infinite path length detected")
     return d
 end
 
-function eccentricity(g::AbstractGraph, distmx::AbstractMatrix)
-    return eccentricity(g, vertices(g), distmx)
+function threaded_eccentricity(
+    g::AbstractGraph, vs=vertices(g), distmx::AbstractMatrix{T}=weights(g)
+) where {T<:Number}
+    vlen = length(vs)
+    eccs = Vector{T}(undef, vlen)
+    Base.Threads.@threads for i in 1:vlen
+        d = Graphs.dijkstra_shortest_paths(g, vs[i], distmx)
+        eccs[i] = maximum(d.dists)
+    end
+    maximum(eccs) == typemax(T) && @warn("Infinite path length detected")
+    return eccs
+end
+
+function eccentricity(
+    g::AbstractGraph, distmx::AbstractMatrix; parallel::Symbol=:distributed
+)
+    return eccentricity(g, vertices(g), distmx; parallel)
 end
 
 function diameter(g::AbstractGraph, distmx::AbstractMatrix=weights(g))
--- a/src/Parallel/traversals/greedy_color.jl
+++ b/src/Parallel/traversals/greedy_color.jl
@@ -1,4 +1,31 @@
-function random_greedy_color(g::AbstractGraph{T}, reps::Integer) where {T<:Integer}
+function random_greedy_color(
+    g::AbstractGraph{T}, reps::Integer; parallel::Symbol=:distributed
+) where {T<:Integer}
+    return if parallel === :threads
+        threaded_random_greedy_color(g, reps)
+    elseif parallel === :distributed
+        distr_random_greedy_color(g, reps)
+    else
+        throw(
+            ArgumentError(
+                "Unsupported parallel argument '$(repr(parallel))' (supported: ':threads' or ':distributed')",
+            ),
+        )
+    end
+end
+
+function threaded_random_greedy_color(g::AbstractGraph{T}, reps::Integer) where {T<:Integer}
+    local_best = Vector{Graphs.Coloring{T}}(undef, reps)
+    Base.Threads.@threads for i in 1:reps
+        seq = shuffle(vertices(g))
+        local_best[i] = Graphs.perm_greedy_color(g, seq)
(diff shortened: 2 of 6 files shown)


simonschoelly:
What is the issue with using `Distrubuted`? As far as I know it is a standard library.

oscardssmith:
Multithreading typically has lower overhead since it's shared memory. Distributed is usually only worthwhile if you are trying to distribute across multiple machines.

oscardssmith:
Also, Distributed is a stdlib, but stdlibs are still dependencies, and as we have started moving stlibs out of the sysimage, they are becoming more like regular packages.

topolarity:
> Would this require also bumping the minimum requirement to julia=1.9?

There are compatibility pathways where we only switch to an extension on 1.9+ and keep the hard dependency for earlier versions

I think this PR should be ready for review now - it is passing tests for me locally

Krastanov (approved):
LGTM, but I am a bit new to the maintainer team, and this does introduce a new public feature (the `parallel` kwarg), so I will wait a bit for dissent from folks with more seniority.

topolarity:
>  a new public feature (the parallel kwarg)

FWIW, that kwarg is already there on a lot of other methods, so it's not a new design choice exactly - it was just never provided for these particular operations

simonschoelly:
> LGTM, but I am a bit new to the maintainer team, and this does introduce a new public feature (the `parallel` kwarg), so I will wait a bit for dissent from folks with more seniority.

There is some new multi threaded code here - so we probably should take a bit of time to review it - as it can lead to notoriously difficult to spot bugs.

src/Parallel/distance.jl
```diff
@@ -1,20 +1,51 @@
 # used in shortest path calculations
 
 function eccentricity(
+    g::AbstractGraph,
+    vs=vertices(g),
+    distmx::AbstractMatrix{T}=weights(g);
+    parallel=:distributed,
```
simonschoelly, line 7:
```suggestion
    parallel::Symbol=:distributed,
```

simonschoelly: I know we haven't done that in other places - but perhaps to make this code more future proof would it make sense to use a specific structure to specify the parallel policy? Then we could later add things such as the number of threads/cores used.

src/Parallel/distance.jl
```diff
@@ -1,20 +1,51 @@
...
+) where {T<:Number}
+    return if parallel === :threads
+        threaded_eccentricity(g, vs, distmx)
+    elseif parallel === :distributed
+        distr_eccentricity(g, vs, distmx)
+    else
+        error(
+            "Unsupported parallel argument '$(repr(parallel))' (supported: ':threads' or ':distributed')",
```
simonschoelly, line 15: We should also add a test for the case that someone uses the wrong argument.

simonschoelly: We repeat the same error multiple times - Perhaps we can throw a custom error type instead?

src/Parallel/distance.jl
```diff
@@ -1,20 +1,51 @@
...
+end
+
+function distr_eccentricity(
     g::AbstractGraph, vs=vertices(g), distmx::AbstractMatrix{T}=weights(g)
 ) where {T<:Number}
     vlen = length(vs)
     eccs = SharedVector{T}(vlen)
     @sync @distributed for i in 1:vlen
-        eccs[i] = maximum(Graphs.dijkstra_shortest_paths(g, vs[i], distmx).dists)
+        d′ = Graphs.dijkstra_shortest_paths(g, vs[i], distmx)
```
simonschoelly, line 26: I am a bit unsure about the usage of `d′` - it is hardly distinguishable from `d'` - in that case `'` is the adjoint operator.

test/parallel/distance.jl
```diff
@@ -7,31 +7,43 @@
...
-        @test isapprox(y, z)
-        @test @inferred(Graphs.diameter(y)) ==
-            @inferred(Parallel.diameter(g, distmx1)) ==
-            6.2
-        @test @inferred(Graphs.periphery(y)) ==
-            @inferred(Parallel.periphery(g, distmx1)) ==
-            [1, 3]
-        @test @inferred(Graphs.radius(y)) == @inferred(Parallel.radius(g, distmx1)) == 4.2
-        @test @inferred(Graphs.center(y)) == @inferred(Parallel.center(g, distmx1)) == [2]
+    for parallel in [:threads, :distributed]
```
simonschoelly, line 10: It might make sense to write the test as 
```julia
@testset "Parallel.Distance" for parallel in [:threads, :distributed]
[...]
end
```

src/Parallel/traversals/greedy_color.jl
```diff
@@ -1,4 +1,29 @@
...
+        distr_random_greedy_color(g, reps)
+    else
+        error(
+            "Unsupported parallel argument '$(repr(parallel))' (supported: ':threads' or ':distributed')",
+        )
+    end
+end
+
+function threaded_random_greedy_color(g::AbstractGraph{T}, reps::Integer) where {T<:Integer}
+    local_best = Any[nothing for _ in 1:reps]
```
simonschoelly, line 16: Is there a reason to use `Any[nothing for _ in 1:reps]` instead of
```julia
Vector{Graphs.Coloring{T}}(undef, reps)
```
or
```julia
Vector{Union{Nothing, Graphs.Coloring{T}}}(nothing, reps)
```
?

simonschoelly (changes requested):
Looks mostly good to me. I don't see any race condition

I have added some comments. They are mostly suggestions. In addition:
- We probably should write some documentation for the new `parallel` argument. Although I admit that this module is fairly undocumented anyway - and as we don't export anything yet we don't have to be so rigorous.

src/Parallel/traversals/greedy_color.jl
topolarity: Nope, I just wasn't sure if that type was always guaranteed

Let's use `Vector{Graphs.Coloring{T}}(undef, reps)`

src/Parallel/distance.jl
```diff
@@ -1,20 +1,55 @@
...
-    return eccentricity(g, vertices(g), distmx)
+function threaded_eccentricity(
+    g::AbstractGraph, vs=vertices(g), distmx::AbstractMatrix{T}=weights(g)
+) where {T<:Number}
+    vlen = length(vs)
+    eccs = Vector{T}(undef, vlen)
+    Base.Threads.@threads for i in 1:vlen
+        d = Graphs.dijkstra_shortest_paths(g, vs[i], distmx)
+        eccs[i] = maximum(d.dists)
+    end
```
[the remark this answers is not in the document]
Krastanov: That has been a common issue for me with macro-transformed `begin-end` blocks. I have never bothered delving into why it is happening though. Macro-heavy libraries like `ResumableFunctions.jl` and `ConcurrentSim.jl` suffer from that (and from other more severe macro-covereage issues).

Krastanov:
The PR Pre-Commit Bot failure is a problem with the github action, not this PR. (Probably due to this coming from a fork -- I will look into fixing the action)

topolarity:
Anything else needed from me here?

topolarity:
(polite) bump - anything still blocking merge?

Krastanov:
Thanks, @topolarity ! Sorry for the slow response, we are a bit low on volunteers right now and Simon is the only current maintainer with actual deep knowledge of the codebase.

Looking through the code, I do not see any issues, all of Simon's comments have been addressed, and any documentation requirements or API restructuring needs are reported as separate posts on the issue tracker. This seems to be complete. I will rerun the tests and merge after they pass.

Krastanov:
The patch codecov is failing, but the overall project coverage is increasing, so it seems to be good on that front too. Merging. Thank you for fixing this up!

PR #432, opened 2025-05-21
Make Julia v1.10 (LTS) min version

simonschoelly:
This PR makes Julia `v1.10` - the latest long-term support version the minimum version of this package.

Additionally
- Remove some code that was only necessary for older versions
- Remove `Compat.jl` dependency as it is not needed anymore (we can always add it back)

I noticed there are some TODO comments for changing the error types in some cases when we upgrade Julia - I left them in for now as this might be a slightly breaking case.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       matrix:
         version:
           - '1'
-          - '1.6'
+          - '1.10'
           - 'pre'
         os:
           - ubuntu-latest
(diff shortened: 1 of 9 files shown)


Krastanov (approved):
I am strongly in favor. A cursory look through the edits does not raise any issues for me, but I am not deeply familiar with the codebase yet. LGTM for what is worth.

simonschoelly:
> I am strongly in favor. A cursory look through the edits does not raise any issues for me, but I am not deeply familiar with the codebase yet. LGTM for what is worth.

Thanks. Most of the changes was the removal of the `@compat` macro - we needed that because `ismutable` was not available before Julia `v1.7`. And then these changes lowered the coverage percentage so I added a test to bring the percentage back up.

Issue #431, opened 2025-05-16
Graph property recognition algorithms


Luis-Varona:
I'd like to add the following graph property recognition algorithms to `Graphs.jl`:

* [ ] `is_cartesian_product` [(Imrich and Peterin 2007)](https://doi.org/10.1016/j.disc.2005.09.038)
* [ ] `is_chordal` [(Klein 1996)](https://doi.org/10.1137/S0097539789166880)
* [ ] `is_chordal_bipartite` + `is_strongly_chordal` [(Spinrad 1993)](https://doi.org/10.1016/0020-0190(93)90209-R)
* [ ] `is_cograph`[(Habib and Paul 2005)](https://doi.org/10.1016/j.dam.2004.01.011)
* [ ] `is_planar` + `is_outerplanar` [(Patrignani 2013)](https://cs.brown.edu/people/rtamassi/gdhandbook/chapters/planarity.pdf)

All of these are algorithms implemented in [`SageMath`](https://doc.sagemath.org/html/en/index.html) that I regularly use in my research, and which I typically have to reimplement from scratch or use my own FFI with sage to access from Julia. Given that `Graphs.jl` already includes functions such as `is_bipartite`, `is_cyclic`, `is_strongly_connected`, etc., I think (at least some of) these algorithms would be well-suited for inclusion.

Should the maintainers deem these additions appropriate, I believe each item listed above merits its own separate PR to avoid overloading any single review.

Krastanov:
Please go ahead, no need to ask for permission. These additions will be appreciated. Review might be somewhat slow, so anything you can do to simplify that step would be useful (check the dev docs, have easy to understand tests, run code autoformatting, etc). Thank you for spending some of your time on improving the library!

Krastanov:
of note, check whether NautyGraphs, VNGraphs, or IGraphs implements any of these features. Using them as a cross check in the unit tests can be very valuable.

simonschoelly:
> Please go ahead, no need to ask for permission. These additions will be appreciated. Review might be somewhat slow, so anything you can do to simplify that step would be useful (check the dev docs, have easy to understand tests, run code autoformatting, etc). Thank you for spending some of your time on improving the library!

From CONTRIBUTING.MD
> If you want to introduce a new feature, open an issue to discuss a feature before you start coding (this maximizes the likelihood of patch acceptance).

[...]

PR #434, opened 2025-06-02
Add an `is_chordal` algorithm

Luis-Varona:
We implement Tarjan and Yannakakis (1984)'s MCS algorithm, taking inspiration from the existing NetworkX implementation. Everything is done except for example doctests in `src/chordality.jl` and unit tests in `test/chordality.jl`. (This PR is part of a new suite of algorithms outlined in issue #431.)


simonschoelly:
Do you by any change know where I could find a public accessible version of that paper? My usual methods did not help.

src/chordality.jl
```diff
@@ -0,0 +1,78 @@
...
+    Reduce Acyclic Hypergraphs." *SIAM Journal on Computing* 13, no. 3 (1984):
+    566–79. https://doi.org/10.1137/0213035.
+[2] NetworkX Developers. "is_chordal." NetworkX 3.5 documentation. NetworkX,
+    May 29, 2025. Accessed June 2, 2025.
+    https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.chordal.is_chordal.html.
+
+# Examples
+TODO: Add examples
+"""
+function is_chordal(g::AbstractSimpleGraph)
```
simonschoelly, line 37:
```suggestion
function is_chordal(g::AbstractGraph)
```
We can extend this to `AbstractGraph` - we don't support multigraphs anyways currently. `AbstractSimpleGraph` is more of a technical type for historical reasons. As long as you don't modify the graph we probably don't need it.

Furthermore we can use traits to check if a graph is directed or not - for an example see here: https://github.com/JuliaGraphs/Graphs.jl/blob/c001daba3ae32a8649af5e21c0c9786d9c38e082/src/biconnectivity/articulation.jl#L23

simonschoelly: Oh I see you are using `induced_subgraph` below - that might requite a `SimpleGraph` - perhaps there is a way to avoid that? This function is very costly as it needs to allocate a new graph.

Krastanov:
[out.pdf](https://github.com/user-attachments/files/20560769/out.pdf)

The paper is attached here, courtesy of the UMass library.

src/chordality.jl
Luis-Varona: Thanks for the review, @simonschoelly&mdash;will respond in full later today/tomorrow. I will say for now that (I think?) `AbstractSimpleGraph` rules out the possibility of parallel edges, which would otherwise be computationally expensive to check, but I'm not sure what your thoughts on the matter are.

Luis-Varona:
> [out.pdf](https://github.com/user-attachments/files/20560769/out.pdf)
> 
> The paper is attached here, courtesy of the UMass library.

Wonderful, @Krastanov, thanks!

src/chordality.jl
simonschoelly: I think in order to work with other graph types defined in other packages it is better to use `AbstractGraph` - for example the graphs defined here are not subtypes of `AbstractSimpleGraph`: https://github.com/JuliaGraphs/SimpleWeightedGraphs.jl

simonschoelly: The current `AbstractGraph` interface does not support multiple edges at all - this is something that we have to tackle in the future - but for now the assumption is that all graphs are simple graphs with potential loops.

simonschoelly: So you don't need to worry about that case.

Luis-Varona:
@simonschoelly, just wanted to let you know I certainly haven't abandoned this PR. I was busy with work this past week and I'm at a conference this weekend, but I'll get back to it shortly. I've looked over your comments and it shouldn't be many more changes at all anyway, hehe.

Luis-Varona:
@simonschoelly, just changed the input type from `AbstractSimpleGraph` to `AbstractGraph`. Anything else you'd like me to change before I get around to adding tests?

Luis-Varona:
Wait—I just remembered your comment about induced subgraphs. I'll think about that and get back to you I was largely just porting over networkx's implementation, which does do this, but it probably really isn't optimal.

simonschoelly:
> @simonschoelly, just wanted to let you know I certainly haven't abandoned this PR. I was busy with work this past week and I'm at a conference this weekend, but I'll get back to it shortly. I've looked over your comments and it shouldn't be many more changes at all anyway, hehe.

Don't worry - we are quite slow with reviewing PRs here :D

Luis-Varona:
@simonschoelly @Krastanov I've fixed the issue regarding unnecessary allocations with `induced_subgraph`, instead having the `_induces_clique` helper take in a vertex subset and, in the context of the original graph, confirm that all possible edges exist. @simonschoelly, is there anything else I should change before I start on unit tests and docstring examples?

Luis-Varona:
Hi @simonschoelly, just following up. Is everything good enough for me to conclude my work on the actual logic and start on the docstring and unit tests? 🙂

Luis-Varona:
What do you think, @Krastanov? 🙂

Krastanov:
Thanks, @Luis-Varona ! No need to wait for anything else to be finished before adding tests on my end. I actually frequently refuse to review PRs before the tests are added, as I judge quality first by the tests and documentation.

I am adding some quality of life improvements to IGraphs.jl that should make it easy to test against their implementation as well. I will post it here shortly.

Krastanov:
Just to clarify: while I might have a particular style of review, that style is not shared by everyone, and please do not hesitate to request me approaching this differently.

Krastanov:
IGraphs.jl just got a new release in which you can use the following in order to test your implementation against the ones in the igraph C library https://igraph.org/c/html/latest/igraph-Structural.html#igraph_is_chordal

```
the_graph = random_regular_graph(10,5)
LibIGraph.is_chordal(IGraph(the_graph), IGNull(), IGNull(), IGNull(), IGNull())
```

I frequently use random sampling over appropriate ensembles as a way to test that two implementations give the same result. This can be useful here as well.

It can also be fun to compare performance (just make sure to not count the time to convert the graph from one format to the other).

simonschoelly:
There is also https://github.com/wangjie212/ChordalGraph

simonschoelly:
You resolved that conversation but might have overlooked what I wrote there about using the `IsDirected` trait.

Luis-Varona:
Thanks for the info, @Krastanov!

> You resolved that conversation but might have overlooked what I wrote there about using the `IsDirected` trait.

So I did, @simonschoelly. Sorry—I must've overlooked that. I expect that sort of fix shouldn't take long, anyway, so I'll try to take a crack at it later this week.

I suppose regardless of the internal workings of our version of `is_chordal`, I can also start right away on the test suite. Testing against only one of the two libraries suggested above (i.e., `ChordalGraphs.jl` _or_ `IGraphs.jl`, not both) should be sufficient, right?

**Edit:** Certainly, benchmarking our implementation against both of the above would be useful, but I'm thinking I'll leave that to the `benchmark/` folder and just use one or the other as a point of comparison in the actual test suite in `test/`.

Krastanov:
Yes, no need to test against multiple other libraries. Your own correctness tests are probably most important -- testing against another library with a bunch of random samples just provides a bit more certainty that we are not missing anything.

Luis-Varona:
Hi @Krastanov and @simonschoelly, sorry for the long silence&mdash;I plan to return to this issue sometime next week. IIRC, all that's left is just a test suite, and maybe a docs fix. 🙂

Krastanov:
Thanks @Luis-Varona

samuelsonric:
@Luis-Varona @Krastanov @simonschoelly  I have an extremely performant implementation of this function [here](https://github.com/AlgebraicJulia/CliqueTrees.jl/blob/ccb384de122d73230b39158837d97bcce9cc73ad/src/chordal_graphs.jl#L6), if that is of any interest. In particular, the MCS algorithm is best implemented using a bucket queue: see [here](https://github.com/AlgebraicJulia/CliqueTrees.jl/blob/ccb384de122d73230b39158837d97bcce9cc73ad/src/elimination_algorithms.jl#L1487).

Krastanov:
@samuelsonric , thanks for the heads up! It would be great to have access to both implementations from Graphs.jl -- this also makes testing much easier as we can just do random tests against each other.

If you or anyone else has the bandwidth, my suggestion would be to do

`is_chordal(graph)` which dispatches to `is_cordal(graph, alg::GraphsMCSAlg)` which is the algorithm implemented here and then in CliqueTrees.jl `import Graphs; Graphs.is_cordal(graph, alg::CliqueTreesMCSAlg)`. The reason I am suggesting the extension happening in CliqueTrees is because CliqueTrees already depends on Graphs. The reason I am suggesting keeping both options is because it is valuable to keep the Graphs dependencies as minimal as possible.

As for `is_chordal` vs `ischordal`, it seems we mostly use the underscore notation for public functions:

image

samuelsonric:
@Krastanov,

I am happy to do this. I can probably implement `is_cograph` as well.

Krastanov:
Thanks, Richard!

Luis-Varona:
Hi @simonschoelly and @Krastanov - I'm really sorry for the long absence. I've just gotten back to work on this PR. (Because it kept getting messy from me merging back in from `master`, I cherry-picked my old commits onto a french branch from `master`.)

@simonschoelly, I incorporated (in the latest commit) your trait dispatch suggestions; hopefully it aligns with what you had in mind. I've also added docstring examples, @Krastanov.

The next step, I suppose, is to update the test suite, and add to the documentation and changelog. (Both of those things I am to do within the day.) As for @samuelsonric's performant implementation, I'll certainly take a look.

Krastanov:
Thanks, @Luis-Varona . Indeed the main next step would be tests, including correctness tests against other libraries. No worries about the delay, we probably would not have had an active volunteer to help with review at that time anyway.

Luis-Varona:
Got it @Krastanov! Do you suggest including dispatch (re: @samuelsonric's implementation) in this PR or in a future one after the basic functionality has been finalized here?

Krastanov:
Your PR is the first one to introduce an implementation of `is_chordal` that is free of other dependencies. We can think of this as the default implementation. No need to include dispatch on other implementations -- and if we are to introduce such dispatch, it would be defined in whatever other library implements it, something like `is_chordal(..., ::OtherLibraryAlgorithm)` defined in the body of `OtherLibrary`. So no need to do anything extra here (unless you are talking about some other dispatch-related setup)

Luis-Varona:
@Krastanov, I've finished the test suite locally and an just adding some comments for clarity before pushing. I'll probably finish it either before sleeping or sometime tomorrow. :slightly_smiling_face:

(And of course, I'll update the CHANGELOG as well once I'm done with that.)

Luis-Varona:
@Krastanov (and @simonschoelly if you're still interested in this PR), everything should be good and ready for review now! :slightly_smiling_face:

Luis-Varona:
@Krastanov (and @simonschoelly if you're still interested in this PR), everything should be good and ready for review now! :slightly_smiling_face:

Krastanov:
Thank you for the implementation and apologies for the slow response.

This looks great, I have only a minor question about the docstring. The docstring says "linear time" but my naive reading is the following:

```
while ... # on each iteration we take a vertex from unnumbered and put it in numbered
    _max_cardinality_vertex(...) # complexity of this line is >= length(numbered)*length(unnumbered)
```

so we have `complexity = number of iterations * bound for length of numbered * bound for length of unnumbered`

Isn't this n^3 complexity?

Otherwise this looks good to merge to me (and we can then politely ask @samuelsonric to add dispatch to their algorithm in their package)

docs/src/algorithms/chordality.md
```diff
@@ -0,0 +1,17 @@
+# Degeneracy
```
Krastanov, line 1: Should the title be `Chordality` instead of `Degeneracy`?

Luis-Varona: Oops, I'm fairly certain I copy-pasted that from another docs file to get the same structure! I'll fix it once I'm back at my computer.

PR #435, opened 2025-06-02
Allow JuliaFormatter v2 + run formatter

simonschoelly:
Follow up to #433 

There is an issue with JuliaFormatter deleting all code that was marked with should-not-format comments so I removed these comments. See https://github.com/domluna/JuliaFormatter.jl/issues/909

And it also produces others errors - so we should probably wait with upgrading.


Krastanov:
FYI, the Formatter v2 bug is now fixed

Krastanov:
closing in favor of #436

PR #440, opened 2025-06-08
Add/line graph

lampretl:
I added the function `line_graph` (one method for undirected graphs, one for directed graphs) and several unit tests that check the correctness of the output graph on simple cases (cycle, path, star graph, and a small graph with 2  or 3 vertices and a loop). 

I also edited `runtests.jl` to use `ARGS`, so that when running 
```
using Pkg;  Pkg.activate(".");  Pkg.instantiate();  Pkg.test("Graphs", test_args=["operators","juliaformatter"]);
```
only those two tests are performed. This was necessary, because running the whole set of tests took 15min on my computer (6-core i7-8850H CPU), but doing only those two tests took <50sec, so it was easier to iterate.


.gitignore
```diff
@@ -14,3 +14,4 @@ benchmark/Manifest.toml
 /docs/src/index.md
 /docs/src/contributing.md
 /docs/src/license.md
+.aider*
```
simonschoelly, line 17: This excludes all files that start with the prefix `.aider` i.e. also a file with the name `.aiderfoo` would be hidden - do we want that?

simonschoelly:
@Krastanov There seems to be an issue with the pre-commit bot - is this the same issue that you mentioned in another PR?

lampretl:
@simonschoelly I'd appreciate any feedback. Are the unit tests sufficient, should I add or remove some of them? Do you find the implementations and docstrings ok?

Krastanov:
Same issue. Currently the bot works only on PRs from branches of this repo. PRs from forks fail. All tests pass, so that is not a blocker.

I would suggest editing the PR description to actually describe what this PR is about.

simonschoelly:
> @simonschoelly I'd appreciate any feedback. Are the unit tests sufficient, should I add or remove some of them? Do you find the implementations and docstrings ok?

I haven't had the time to look at it entirely - but I wrote a comment yesterday that somehow did not get posted:

I was wondering about the behavior for loops. Currently a graph with edges `(1 -- 1)` and `(1 --2)` while result in the line graph with the single edge `(1--1) -- (1--2)`. But should the there not also be a an edge `(1--1) - (1--1)`?

src/operators.jl
```diff
@@ -879,3 +879,88 @@ function merge_vertices!(g::Graph{T}, vs::Vector{U} where {U<:Integer}) where {T
...
+```jldoctest
+julia> using Graphs
+
+julia> g = path_graph(5);
+
+julia> lg = line_graph(g)
+{4, 3} undirected simple Int64 graph
+```
+"""
+function line_graph(g::SimpleGraph)::SimpleGraph
```
simonschoelly, line 900:
```suggestion
function line_graph(g::SimpleGraph)
```
Julia does not have true return type annotations. If you have a function `f(x)::T` then Julia will just call `convert(T, f(x))`.

src/operators.jl
```diff
@@ -879,3 +879,88 @@ function merge_vertices!(g::Graph{T}, vs::Vector{U} where {U<:Integer}) where {T
...
+        for i in 1:(n - 1), j in (i + 1):n  # iterate through pairs of edges with same endpoint
+            ei, ej = es[i], es[j]
+            m += 1
+            push!(edge_to_neighbors[ei], ej)
+            push!(edge_to_neighbors[ej], ei)
+        end
+    end
+
+    foreach(sort!, edge_to_neighbors)
+    return SimpleGraph(m, edge_to_neighbors)
```
simonschoelly, line 922:
```suggestion
    fadjlist = [Int[] for _ in 1:ne(g)]
    m = 0  # number of edges in the line-graph
    for es in vertex_to_edges
        n = length(es)
        for i in 1:(n - 1), j in (i + 1):n  # iterate through pairs of edges with same endpoint
            ei, ej = es[i], es[j]
            m += 1
            push!(fadjlist[ei], ej)
            push!(fadjlist[ej], ei)
        end
    end

    foreach(sort!, fadjlist)
    return SimpleGraph(m, fadjlist)
```
Let's call it `fadjlist` so it is clear that you use the kind of private constructor that does not really check its argument but just fills the struct.

test/runtests.jl
simonschoelly, line 1: Was this committed by accident?

simonschoelly:
So the literature is quite sparse with regards to line graphs and self-loops but I just checked the behavior of networkx and igraph on a single vertex with a self-loop.

For directed graphs they line graph for both packages result in a single vertex with a self-loop again.

For undirected graphs, networkx produces a single vertex without any loops. Igraph produces a single vertex with a self-loop.

simonschoelly (changes requested):
I think the tests are mostly good - what I noticed though that you only run them for `SimpleGraph{Int}` and `SimpleDiGraph{Int}` - just to make sure they also work correctly for other eltypes I would add a few tests for example for `SimpleGraph{UInt8}`.

.gitignore
lampretl: Umm, [aider](https://github.com/Aider-AI/aider) creates several hidden files and folders whose names start with `.aider`, so I would say: yes.

lampretl:
> I would suggest editing the PR description to actually describe what this PR is about.

Done.

test/runtests.jl
lampretl: No, I edited `runtests.jl` to be able to run just specific tests (those that I added or changed). It made iterating much faster (30sec instead of 15min).

src/operators.jl
```diff
@@ -879,3 +879,88 @@ function merge_vertices!(g::Graph{T}, vs::Vector{U} where {U<:Integer}) where {T
...
+    vertex_to_edgesout = [Int[] for _ in 1:nv(g)]
+    vertex_to_edgesin = [Int[] for _ in 1:nv(g)]
+    for (i, e) in enumerate(edges(g))
+        s, d = src(e), dst(e)
+        push!(vertex_to_edgesout[s], i)
+        push!(vertex_to_edgesin[d], i)
+    end
+
+    edge_to_neighborsout = [Int[] for _ in 1:ne(g)]
+    edge_to_neighborsin = [Int[] for _ in 1:ne(g)]
```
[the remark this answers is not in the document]
lampretl: Agreed, done.

src/operators.jl
lampretl: Agreed, done.

lampretl:
> I would add a few tests for example for `SimpleGraph{UInt8}`.

Ok, I did the test for the smallest (di)graph (2 vertices) for Int8,...,Int128,Uint8,...,UInt128.

src/operators.jl
```diff
@@ -879,3 +879,88 @@ function merge_vertices!(g::Graph{T}, vs::Vector{U} where {U<:Integer}) where {T
...
+        n = length(es)
+        for i in 1:(n - 1), j in (i + 1):n  # iterate through pairs of edges with same endpoint
+            ei, ej = es[i], es[j]
+            m += 1
+            push!(edge_to_neighbors[ei], ej)
+            push!(edge_to_neighbors[ej], ei)
+        end
+    end
+
+    foreach(sort!, edge_to_neighbors)
```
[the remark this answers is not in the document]
lampretl: Hmm, the docstring of `edges` does not say we can count on the lexicographical order of iteration. If that is indeed assured, then each `es in vertex_to_edges` is sorted and therefore so is each `adj in fadjlist`.

So, what is best to do? 
(1) Keep `sort!`, just in case? (no changes)
(2) Replace `sort!` with `@assert all(issorted(adj) for adj in fadjlist)`?
(3) Remove `sort` and count on `edges` providing lexicographic sorting?

lampretl:
> I was wondering about the behavior for loops. Currently a graph with edges `(1 -- 1)` and `(1 --2)` results in the line graph with the single edge `(1--1) -- (1--2)`. But should there not also be an edge `(1--1) - (1--1)`?

If so, by the same 'logic', shouldn't there also be an edge `(1--2) - (1--2)` in lg? I prefer not spamming self-loops in `lg`.

> So the literature is quite sparse with regards to line graphs and self-loops but I just checked the behavior of `networkx` and `igraph` on a single vertex with a self-loop.

> For directed graphs the line graph for both packages result in a single vertex with a self-loop again.

> For undirected graphs, networkx produces a single vertex without any loops. Igraph produces a single vertex with a self-loop. 

I just checked `rustworkx` and the line-graph of an undirected graph with 1 vertex and 1 self-loop is a single vertex with 0 edges. It seems line graph is not even implemented there for digraphs. 

I don't know. I'll do as you suggest. 

Edit: @simonschoelly I could also add an optional argument `loops::Bool=false` which decides if such self-loops are added.

.gitignore
simonschoelly: Ok the lets keep it for now

Manifest.toml
simonschoelly, line 1: We don't commit the Manifest.toml - and I am a bit confused how that could happen because it actually is in that .gitignore file above.

lampretl:
> > I was wondering about the behavior for loops. Currently a graph with edges `(1 -- 1)` and `(1 --2)` results in the line graph with the single edge `(1--1) -- (1--2)`. But should there not also be an edge `(1--1) - (1--1)`?
> 
> If so, by the same 'logic', shouldn't there also be an edge `(1--2) - (1--2)` in lg? I prefer not spamming self-loops in `lg`.
> 
> > So the literature is quite sparse with regards to line graphs and self-loops but I just checked the behavior of `networkx` and `igraph` on a single vertex with a self-loop.
> 
> > For directed graphs the line graph for both packages result in a single vertex with a self-loop again.
> 
> > For undirected graphs, networkx produces a single vertex without any loops. Igraph produces a single vertex with a self-loop.
> 
> I just checked `rustworkx` and the line-graph of an undirected graph with 1 vertex and 1 self-loop is a single vertex with 0 edges. It seems line graph is not even implemented there for digraphs.
> 
> I don't know. I'll do as you suggest.
> 
> Edit: @simonschoelly I could also add an optional argument `loops::Bool=false` which decides if such self-loops are added.

@simonschoelly What is your decision regarding self-loops in a line-graph? Shall I add an optional argument, like `loops::Int=0`, whose value determines: `0` $\Rightarrow$ we add no self-loops; `1` $\Rightarrow$ we add loops to `lg` from loops in `g`; `2` $\Rightarrow$ we add loops to `lg` from every edge in `g`? 

Also, what is your decision regarding `sort!/issorted` of `fadjlist` (the 3 options I described above?

I'd like to finish this PR in a timely manner.

elvcastelo:
> @simonschoelly What is your decision regarding self-loops in a line-graph? Shall I add an optional argument, like `loops::Int=0`, whose value determines: `0` ⇒ we add no self-loops; `1` ⇒ we add loops to `lg` from loops in `g`; `2` ⇒ we add loops to `lg` from every edge in `g`?

Sorry for interrupting the discussion :)

Maybe it should follow from the definition of line graphs, as it make easier to understand the output. For the undirected case, we have a single loopless vertex. Then the options would not be necessary (unless for some specific reason?).

SageMath seems to follow this convention (https://doc.sagemath.org/html/en/reference/graphs/sage/graphs/line_graph.html#module-sage.graphs.line_graph) when mentioning «Loops are not adjacent to themselves».

For example, consider the P_3 {{1, 2}, {2}, {2, 3}} with a self-loop in the middle vertex. The intersection graph of the edges is a triangle.

Krastanov:
Hi, @lampretl ! Apologies for letting the momentum here die out. I would like to try to help to get this merged. It seems almost ready. A couple of comments:

- Could you undo the changes you made to runtests.jl? It helps to keep it more standardized compared to other julia packages. The julia ecosystem itself should do more work on figuring out how to do selective package runs (maybe we should switch to TestItems), but doing it bespokely different in each package is a bad precedent.
- If you end up creating flag options in your API, do not encode the flags into numbers, rather use Symbols. E.g. `option=:some_descriptive_symbol`. 
- We can just merge this without extra options and worry about extra options in future PRs. Maybe following what @elvcastelo said and using the conventions from Sage. But generally, as long as we do not need a breaking change in the future to incorporate more detailed options, we can just merge things early.

I will mark this as a draft for now to keep my review queue a bit more organized, but do not hesitate to convert it back if you have the energy and desire to finish it. Thank you for sticking around, especially given the slow pace at which we get to do reviews!

lampretl:
@elvcastelo Hey, no worries, you are not interrupting at all :). But I don't understand your comment. How do you suggest self-loops in `lg` are treated?

@Krastanov Give me a few days, I will refresh my memory about this PR and edit it as necessary. Hopefully, I'll get replies a bit faster so the code can be finished and merged into main in a timely manner. : )

Krastanov:
Thanks, Leon! We are struggling to get enough volunteer reviewers, so please do not hesitate to "bump" with reminders to review -- these are actually appreciated. Apologies for having to ask you to do so.

Relatedly, if you feel like it, do not hesitate to provide reviews to existing PRs -- those are greatly appreciated.

henrik-wolf:
To me it seems the main definition on wikipedia assumes that we are dealing with simple graphs, so no self loops are allowed anyway. We could just do the same here, and throw an error when the given graph has self loops.

Similarly, the line graph L(G) has to be simple as well, otherwise the definition
```
each vertex of L(G) represents an edge of G
two vertices of L(G) are adjacent if and only if their corresponding edges share a common endpoint in G.
```
would have self loops all over, as any edge (i-j) has a common endpoint with itself. (The question whether a self loop in G should cause a self loop in L(G) boils down to whether we check adjacency of an edge with itself. Doing this selectively only for self loops seems very arbitrary to me.)

changing the definition to `two DISTINCT vertices of L(G) [...]` will get rid of this problem. This is now equivalent to constructing the graph as the intersection graph of the edge sets which explicitly excludes the (trivially non-empty) intersection of a set with itself.

That way, a loop (or any edge for that matter...) is never adjacent to itself, but is adjacent to any edge going in or out of the looped vertex. I think this should generalize nicely to the directed case as well.

lampretl:
@Krastanov Hey, better late than never. I copy-pasted the latest commit from the Graphs.jl main branch into my fork, then improved my implementation of `line_graph`:

- I undid the changes to `runtests.jl`, as you asked.
- I added 3 options to deal with self-loops in `lg`, as `Symbol`s, like you asked.

Please have a look at the latest changes, and if satisfied, merge into main.

P.S. Upon looking at the diffs, there are way too many (all the changed code since june 2025). I only edited 
src/Graphs.jl
src/operators.jl
test/operators.jl.
src/SimpleGraphs/simplegraph.jl 
Is there a better workflow for this? Should I have deleted my PR and fork, made a new one from the latest Graphs.jl main, then inserted my code and made a PR?

PR #437, opened 2025-06-03
autoformat bot in CI

Krastanov:
Each commit is reasonably selfcontained:

- add an autoformat bot that will create a commit in this same PR with formatting fixes
- if a PR is pushed to, stop the CI running on old commits, to avoid burning our free credits
- use a more recent simpler version of the cache action
- add a misformat error on purpose to test the bot
- add a PR template explaining the presence of the bot
- the automated linter commit that was generated by the bot, fixing my on-purpose injected mistake

@rafaqz @simonschoelly , let me know if this fits your needs, I would like to merge it soon so we can remove one of the burdens putting off new contributors.

PR #442, opened 2025-06-10
turn off the precommit CI action

Krastanov:
Temporarily undo most of #437, to complete the erroneously skipped review and to fix the fact that the CI does not run when the PR is coming from a fork.

The concurrency fixes to the existing actions is kept in as well as the general-purpose precommit configuration.


simonschoelly (approved):
Thanks. I am not sure how to handle this, this has also been an issue in the past when I tried to push into someones PR.

PR #462, opened 2025-10-07
fix example code

karlwessel:
Just removes an superfluous parenthesis from some example code.


nsajko:
CI would probably pass if this PR was rebased on top of current `master`. See merged PR #463.

Krastanov (approved):
Thanks!

PR #469, opened 2025-11-01
Fix typo in traversal docs

juliohm:
I believe this is a typo, but didn't check the source code. It doesn't make sense to have a default cost of 1.0 when the user assigns the value 0.0.

### DIFF
--- a/docs/src/first_steps/paths_traversal.md
+++ b/docs/src/first_steps/paths_traversal.md
@@ -10,7 +10,6 @@ Any graph traversal will traverse an edge only if it is present in the graph. Wh
 
 1. distance values for undefined edges will be ignored;
 2. any unassigned values (in sparse distance matrices) for edges that are present in the graph will be assumed to take the default value of 1.0;
-3. any zero values (in sparse/dense distance matrices) for edges that are present in the graph will instead have an implicit edge cost of 0.0.
 
 ## Graph traversal
 


Krastanov:
In this piece of code you seem to be correct https://github.com/JuliaGraphs/Graphs.jl/blob/master/src/traversals/maxadjvisit.jl#L37-L49

However, the mere presence of this third point makes sense only if your correction is wrong. If your correction is right, we should just remove this line or at least restructure it. Otherwise, with the correction it reads "when you do A, surprisingly A does happen".

@gdalle , `git blame` says you wrote this line in the docs. Could you comment on this edit? I try to be careful with [Chesterton's fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence)

gdalle:
> @gdalle , git blame says you wrote this line in the docs. Could you comment on this edit? I try to be careful with [Chesterton's fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence)

This must be from when I formatted the whole repo with JuliaFormatter (and didn't know about `git blame` yet). Sorry I can't help out here, I don't know the traversal code base well. For shortest paths what @juliohm wrote would be true indeed

juliohm:
> we should just remove this line

I agree it is best to just remove the line. I can't think of an algorithm where this behavior would make sense. I am not fresh on graph algorithms at the moment, but am almost sure the change from 0 to 1 weight by default is wrong in the context of path traversals.

Krastanov:
Apologies for the delay in merging! Thanks for the cleanup!

PR #466, opened 2025-10-24
Update deprecated IntDisjointSets name

m-fila:
DataStructures v19.0 renamed `IntDisjointSets` to `IntDisjointSet` and deprecated `IntDisjointSets`

PR #471, opened 2025-11-06
Fix warning from DataStructures.jl

juliohm:
This is a really basic warning, but I think it didn't get fixed because the tests are annoyingly using JuliaFormatter which is not compatible with the latest version of DataStructures.jl?

Could someone please jump in and fix the tests?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -22,7 +22,7 @@ GraphsSharedArraysExt = "SharedArrays"
 [compat]
 ArnoldiMethod = "0.4"
 Distributed = "1"
-DataStructures = "0.18, 0.19"
+DataStructures = "0.19"
 Inflate = "0.1.3"
 LinearAlgebra = "1"
 Random = "1"
--- a/src/Graphs.jl
+++ b/src/Graphs.jl
@@ -8,7 +8,7 @@ using Statistics: mean
 
 using Inflate: InflateGzipStream
 using DataStructures:
-    IntDisjointSets,
+    IntDisjointSet,
     PriorityQueue,
     dequeue!,
     dequeue_pair!,
--- a/src/graphcut/karger_min_cut.jl
+++ b/src/graphcut/karger_min_cut.jl
@@ -23,7 +23,7 @@ function karger_min_cut(g::AbstractGraph{T}) where {T<:Integer}
     nvg < 2 && return zeros(Int, nvg)
     nvg == 2 && return [1, 2]
 
-    connected_vs = IntDisjointSets(nvg)
+    connected_vs = IntDisjointSet(nvg)
     num_components = nvg
 
     for e in shuffle(collect(edges(g)))
--- a/src/spanningtrees/boruvka.jl
+++ b/src/spanningtrees/boruvka.jl
@@ -15,7 +15,7 @@ function boruvka_mst end
 @traitfn function boruvka_mst(
     g::AG::(!IsDirected), distmx::AbstractMatrix{T}=weights(g); minimize=true
 ) where {T<:Number,U,AG<:AbstractGraph{U}}
-    djset = IntDisjointSets(nv(g))
+    djset = IntDisjointSet(nv(g))
     # maximizing Z is the same as minimizing -Z
     # mode will indicate the need for the -1 multiplication
     mode = minimize ? 1 : -1
--- a/src/spanningtrees/kruskal.jl
+++ b/src/spanningtrees/kruskal.jl
@@ -12,7 +12,7 @@ function kruskal_mst end
 @traitfn function kruskal_mst(
     g::AG::(!IsDirected), distmx::AbstractMatrix{T}=weights(g); minimize=true
 ) where {T<:Number,U,AG<:AbstractGraph{U}}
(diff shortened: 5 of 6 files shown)


Krastanov:
The JuliaFormatter test dependency is compat bound to its old version 1.0 because there were problems with 2.0. It seems 1.0 clashes with DataStructures 0.19.

If the problems with 2.0 are fixed, we can rerun the formatter and remove that bound. That is the blocker to making DataStructures 0.19 work.

Krastanov:
duplicate of PR https://github.com/JuliaGraphs/Graphs.jl/pull/466

juliohm:
Maybe the question is: why JuliaFormatter is a test dependency in a package
like Graphs.jl? Can't we get rid of it?

Em qui., 6 de nov. de 2025, 11:44, Stefan Krastanov <
***@***.***> escreveu:

Krastanov:
We can not delete it -- Graphs.jl enforces formatting as a part of its tests. Aggressive quality assurance like this is important for large projects with almost zero maintainers, like Graphs.jl. At least that was the belief of the couple of volunteers that were maintaining this tool in the last couple of years.

Of course, if someone volunteers to be an active maintainer, they would be able to change that policy.

juliohm:
I used to be one of the core maintainers of LightGraphs.jl before it was converted into Graphs.jl. Back in the days I don't recall any test that depended on formatting.

Could you please clarify the current situation with a specific test that depends on JuliaFormatter?

Krastanov:
formatting attempted again now in https://github.com/JuliaGraphs/Graphs.jl/pull/472

Krastanov:
> I used to be one of the core maintainers of LightGraphs.jl before it was converted into Graphs.jl. Back in the days I don't recall any test that depended on formatting.
> 
> Could you please clarify the current situation with a specific test that depends on JuliaFormatter?

Currently I am the only maintainer and I do not have the standing to make changes without input from the community and from more volunteers. I appreciate your dev work from the past, and if you have the bandwidth to get more involved again, that would certainly be appreciated.

The Formatter checks were instituted before I started volunteering to maintain by the previous two caretakers. The argument in favor is that if you do not have enough maintainers you need harsh quality assurance to avoid bitrot. I am weakly in favor of that approach, but will not be unilaterally changing it without other active maintainers on board.

The tests is here: https://github.com/JuliaGraphs/Graphs.jl/blob/9a7a6df10a34a7a7d02c46f38d7eeaa8b9cdea67/test/runtests.jl#L168

juliohm:
I believe this test can be removed in favor of GitHub Actions that check formatting issues. We did set this up across all our currently maintained packages. See Meshes.jl for example, this action: https://github.com/JuliaGeometry/Meshes.jl/blob/master/.github/workflows/FormatPR.yml

Every pull request runs a formatting check and the bot makes suggestions to fix issues. No code is merged otherwise.
The format is specified in this file: https://github.com/JuliaGeometry/Meshes.jl/blob/master/.JuliaFormatter.toml

>  I appreciate your dev work from the past, and if you have the bandwidth to get more involved again, that would certainly be appreciated.

I can try to help. For some reason I lost write access to Graphs.jl after the conversion from LightGraphs.jl.

Krastanov:
added you to the team -- you should have access again

Krastanov:
Closing in favor of #472 where the formatting issue is addressed. I cherrypicked the commit from here.

I like the idea of adding the FormatPR.yml workflow, but lets keep the tests too, to avoid accidentally merging things without properly formatted code.

juliohm:
> I like the idea of adding the FormatPR.yml workflow, but lets keep the tests too, to avoid accidentally merging things without properly formatted code.

The GitHub Action I mentioned will automatically open a PR with fixes related to formatting, so the main branch will always adhere to the specified code style.

I still think that we should not add JuliaFormatter as a test dependency. This is not a dependency that is testing the feature-set of the package, but rather the quality of the code in the repository. For the latter, GitHub Actions are more appropriate and avoid version conflicts like this one we saw that was holding unrelated updates.

Can I implement the GitHub Action for formatting code and remove the format tests to eliminate JuliaFormatter from the test dependencies?

Krastanov:
If no one else complains about this, I consent, as long as the PRs that the formatter action makes are towards the original PR branch, not to master. If the formatter PRs are to master, that would result in a very messy git history with a ton of formatting commits making git blame much harder to use.

The one thing I am strongly opposed to is frequent formatting commits in the master branch.

JoshuaLampert:
Just to clarify any misunderstandings because I came across the discussion here: The format GitHub action, @juliohm linked above does not create any PRs (that was the previous version of the action we had in Meshes.jl, @juliohm). It runs as a separate check in PRs and fails if the PR does not adhere to the formatting rules. It also suggests review comments to fix formatting (at least for PRs not coming from forks). I agree with @juliohm that it would be a better practice using a GitHub Action to check formatting rather than adding JuliaFormatter.jl as a test dependency.

Krastanov:
Thanks, Joshua, that clarifies the situation well for me. I am in favor, especially if someone has the bandwidth to make that change.

JoshuaLampert:
Ok, I opened #476.

PR #472, opened 2025-11-06
try formatter v2 again and bump DataStructures.jl to avoid warnings

Krastanov:


src/Experimental/Traversals/bfs.jl
```diff
@@ -1,4 +1,5 @@
-import Base.Sort, Base.Sort.Algorithm
+using Base.Sort: Sort
+using Base.Sort.Algorithm: Algorithm
```
m-fila:
```suggestion
import Base.Sort
import Base.Sort.Algorithm
```

Reformatting broke these imports

m-fila:
Sorry I didn't follow up in #466
I'm not sure about the stability of v2 yet. Some projects (SciML) seem to already moved to it. Yet the tooling doesn't seem to fully support it. I often use vscode which is stuck with v1, https://github.com/julia-vscode/julia-vscode/issues/3819,  and then I end up constantly reformatting code between v2-v1-v2 which sometimes breaks something on the way

Krastanov:
Thanks for looking into this! I have not thought about it much but I am still in favor of figuring out a way that still sticks to using Formatter v2.

Krastanov:
This should be ready for review now. Please do not squash if you merge, so that we can have the single re-formatting commit hidden from git-blame in the future.

Krastanov:
Thanks for the fix! I do not know how I missed something with so many loud warnings. That is now cherrypicked and the precommit config file is also removed.

I will wait for tests to pass without warnings and will merge. After that I will setup a git blame exclusion file so that the formatting commit does not pollute git blame.

JoshuaLampert:
Would you mind creating a new release including this PR, such that the warnings are fixed?

Krastanov:
Thanks for the reminder, I will do so in a couple of minutes.

JoshuaLampert:
Great, thanks a lot!

PR #482, opened 2025-12-20
Optimization: iFUB algorithm for unweighted graph diameter

LoveLow-Global:
Implement iFUB algorithm for diameter calculation of unweighted graphs, `SimpleGraph` and `SimpleDiGraph`

Below is a comparison with the original diameter calculation logic, ran on my desktop.

```
using Graphs

g = erdos_renyi(10000, 0.005) # 10k nodes, sparse

# New implentation
@time d = diameter(g) # 0.910222 seconds (372.05 k allocations: 6.602 MiB)

# Original Method
original_diameter(g) = maximum(eccentricity(g)) 
@time original_diameter(g) # 41.471079 seconds (650.01 k allocations: 13.515 GiB, 3.59% gc time)
```

### DIFF
--- a/src/distance.jl
+++ b/src/distance.jl
@@ -95,7 +95,7 @@ end
 Given a graph and optional distance matrix, or a vector of precomputed
 eccentricities, return the maximum eccentricity of the graph.
 
-An optimizied BFS algorithm (iFUB) is for unweighted graphs, both in [undirected](https://www.sciencedirect.com/science/article/pii/S0304397512008687) 
+An optimizied BFS algorithm (iFUB) is used for unweighted graphs, both in [undirected](https://www.sciencedirect.com/science/article/pii/S0304397512008687) 
 and [directed](https://link.springer.com/chapter/10.1007/978-3-642-30850-5_10) cases.
 
 # Examples
@@ -109,6 +109,8 @@ julia> diameter(path_graph(5))
 4
 ```
 """
+function diameter end
+
 diameter(eccentricities::Vector) = maximum(eccentricities)
 
 diameter(g::AbstractGraph) = diameter(g, weights(g))
--- a/test/distance.jl
+++ b/test/distance.jl
@@ -81,7 +81,6 @@
 
         NUM_SAMPLES = 50 # Adjust this to change test duration
 
-        Random.seed!(42)
         for i in 1:NUM_SAMPLES
             # Random unweighted Graphs
             n = rand(10:1000) # Small to Medium size graphs
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Graphs"
 uuid = "86223c79-3864-5bf0-83f7-82e725a168b6"
-version = "1.13.2"
+version = "1.13.3"
 
 [deps]
 ArnoldiMethod = "ec485272-7323-5ecc-a04f-4719b315124d"


Krastanov:
Thank you! On first look this looks great, but it will take a bit to get a proper review.

In the meantime, we should probably have some more rigorous tests for it added. (1) Tests comparing against large graphs with known diameters (e.g. by construction); and (2) tests comparing against the old slow implementation for random graphs. Could you add such tests, maybe something that adds an additional 60 seconds to the tests, but with a "number of random samples" parameter that I can change locally and run for a few hours on my machine?

LoveLow-Global:
Hello Krastanov, thanks for the comment. I have added (1) comparison against graphs with known diameters and the (2) comparison with old slow implementation for random graphs, under test/distance.jl. You can change NUM_SAMPLES there. Also, please note that it is my first time contributing to an existing Julia package, so please tell me if there is any changes to be done.

Krastanov:
Thanks, this looks great! I will try to review this in the next few days.

One other thing that I need to check lately -- if you used an LLM-like tool, could you disclose that, which tool, and what the prompt was? I do not see anything in your PR that makes me think that was the case, and it is not like LLM tools are bad (they help a lot), but they do require approaching the review differently and having this extra info helps.

LoveLow-Global:
Thanks for the feedback! I have used Gemini 3.0 for code generation. This optimization method was originally used in one of my researches, but I needed to change the style of the code to be similar to Graphs.jl. Therefore, I used Gemini to first change the code into a "Graphs.jl" format. Then I manually edited the codes again.

Krastanov:
Do you happen to have a reference of where this algorithm was introduced first / who named it iFUB, etc? Or a textbook about it?

Krastanov:
Pardon the formatter error -- it seems it is due to an incorrect security configuration on our end, not due to your code.

LoveLow-Global:
Hello, sorry for the late response. The iFUB algorithm was first introduced in the following research papers.

**[Undirected Graphs]**
Crescenzi, P., Grossi, R., Habib, M., Lanzi, L., Marino, A.: On Computing the Diameter of Real-World Undirected Graphs. Presented at Workshop on Graph Algorithms and Applications (Zurich–July 3, 2011) and selected for submission to the special issue of Theoretical Computer Science in honor of Giorgio Ausiello in the occasion of his 70th birthday (2011)

**[Directed Graphs]**
Crescenzi, P., Grossi, R., Lanzi, L., Marino, A. (2012). On Computing the Diameter of Real-World Directed (Weighted) Graphs. In: Klasing, R. (eds) Experimental Algorithms. SEA 2012. Lecture Notes in Computer Science, vol 7276. Springer, Berlin, Heidelberg. https://doi.org/10.1007/978-3-642-30850-5_10

Thank you. Please let me know if there is anything more needed.

test/distance.jl
```diff
@@ -44,6 +46,63 @@
...
+        for i in 1:NUM_SAMPLES
+            # Random unweighted Graphs
+            n = rand(10:1000) # Small to Medium size graphs
+            p = rand() * 0.1 + 0.005 # Sparse to medium density
+
+            # Undirected Graphs
+            g = erdos_renyi(n, p)
+            ccs = connected_components(g)
+            largest_component = ccs[argmax(length.(ccs))]
+            g_lscc, _ = induced_subgraph(g, largest_component)
```
Krastanov: While we should keep these tests of largest connected components, we should also have a set of tests that do not strip to just the largest connected subgraph, because this hides an inconsistency with the current master's behavior. One can argue that your current implementation has a more natural behavior, but even if that is true, it is a breaking change.

On master we have:

```
julia> diameter(Graph(2)) # a disconnected graph
┌ Warning: Infinite path length detected for vertex 1
└ @ Graphs ~/.julia/packages/Graphs/lNprZ/src/distance.jl:78
┌ Warning: Infinite path length detected for vertex 2
└ @ Graphs ~/.julia/packages/Graphs/lNprZ/src/distance.jl:78
9223372036854775807 # this is typemax(indextype)
```

If you do not like this behavior on current master and you have an argument for why it needs to be changed, we can discuss it though.

src/distance.jl
```diff
@@ -91,10 +91,14 @@ end
 """
     diameter(eccentricities)
     diameter(g, distmx=weights(g))
+    diameter(g::Union{SimpleGraph, SimpleDiGraph})
 
 Given a graph and optional distance matrix, or a vector of precomputed
 eccentricities, return the maximum eccentricity of the graph.
 
+For unweighted `SimpleGraph` and `SimpleDiGraph`, an optimized BFS algorithm
+(iFUB) is used to avoid computing eccentricities for all vertices.
```
Krastanov: Could you add nicely formatted references (the two papers you shared) to the docstring?

test/distance.jl
```diff
@@ -44,6 +46,63 @@
             @test @inferred(center(z)) == center(g, distmx2) == [2]
         end
     end
+
+    @testset "$(typeof(g))" for g in test_generic_graphs(a3)
```
Krastanov: The name of this testset is a bit cryptic to me. Could you elaborate it so that it is more immediately understandable if a test failure is reported?

Krastanov:
And one last thing: let's make sure `diameters(Graph())` works (i.e. it works on empty graphs), which currently fails.

src/distance.jl
```diff
@@ -107,10 +111,117 @@ julia> diameter(path_graph(5))
...
 diameter(eccentricities::Vector) = maximum(eccentricities)
+
 function diameter(g::AbstractGraph, distmx::AbstractMatrix=weights(g))
     return maximum(eccentricity(g, distmx))
 end
 
+function diameter(g::Union{SimpleGraph,SimpleDiGraph})
+    if nv(g) <= 1
+        return 0
+    end
```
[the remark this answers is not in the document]
LoveLow-Global: Yes, I have noticed that it should have been `typemax()`. I have edited that, thanks for pointing this out.

src/distance.jl
LoveLow-Global: Yes, I added the references.

src/distance.jl
```diff
@@ -91,10 +91,15 @@ end
 """
     diameter(eccentricities)
     diameter(g, distmx=weights(g))
+    diameter(g::Union{SimpleGraph, SimpleDiGraph})
 
 Given a graph and optional distance matrix, or a vector of precomputed
 eccentricities, return the maximum eccentricity of the graph.
 
+An optimizied BFS algorithm (iFUB) is used for unweighted graphs, both [undirected](https://semeval.inria.fr/2012/printemps/theme1/teams/gang/62.pdf)
```
Krastanov: This links seems to be someone's personal page or some informal repository. Could you switch the link to a more official resources, something that has better chances of still existing in 20 years?

src/distance.jl
```diff
@@ -91,10 +91,15 @@ end
 """
     diameter(eccentricities)
     diameter(g, distmx=weights(g))
+    diameter(g::Union{SimpleGraph, SimpleDiGraph})
```
Krastanov: If I am reading this correctly, your algorithm supports any graph that implements the Graphs.jl API. E.g. LEMONGraph/IGraph/NautyGraph/VNGraph are all supported, because they have the `outneighbors`/`inneighbors`/`vertices` functions.

We should not artificially exclude them (that is not really in the spirit of Multiple Dispatch languages like Julia).

For an unweighted graph we automatically have `weights(g) === DefaultDistance(g)`.

So the new methods should probably be:

- `diameter(g, ::DefaultDistance)`  your implementation
- `diameter(g) = diameter(g, weights(g))`
- `diamgter(g, distmx)` the already existing method but with the default `=weights(g)` removed so that the dispatch tree works

Krastanov (changes requested):
Thanks!

I think there are only two minor things left:
- making sure this works on any graph, not just on specific graph types (after all other graph types might support the same API)
- making sure the doc links are to archival pages that probably will exist for multiple decades

src/distance.jl
LoveLow-Global: I have edited the codes, I hope it is alright now. Thanks for pointing this out, I missed this point.

LoveLow-Global:
Let me know if there's anything else needed from my side to get this merged. Thank you.

Krastanov:
Looks good! I pushed a couple of minor changes:

- bumping the version so we can make a release immediately after merging
- a minor typo in the docstring, and attaching the docstring to the function as a whole instead of attaching it to a specific method
- removing the hardcoded seed (just to make sure tests explore more each time they run) -- for reproducibility, the test framework reports the RNG state in case of a failure

This should be merged and released shortly. Thank you for this contribution, it is very valuable!

PR #489, opened 2026-02-19
outdated option in the codecov workflow file

Krastanov:


nsajko:
It is not exactly a typo. Rather the `file` option was deprecated in favor of `files` with v5 of the codecov action.

PR #490, opened 2026-02-21
Update CI workflows: AirspeedVelocity benchmarks, fix downgrade

Krastanov-agent:
Comment from Stefan, the human: A bunch of the CI workflows were not working reliably. Benchmarks just do not run (airspeed velocity seems to be the cleanest way to run them these days), invalidation checks were outdated and not useful, and the downgrade action is missing a config option which would lead to problems when we change the julia version we are testing against.

Below is the autogenerated description from the agent:

## Summary
- Delete `postbench.yml` and `runbenchandupload.yml`, replace with a single AirspeedVelocity-based `benchmark.yml` workflow (matching QuantumSavory pattern)
- Add `julia_version` parameter to `julia-downgrade-compat@v2` action in `downgrade.yml`

## Test plan
- [ ] Verify benchmark workflow triggers correctly on PRs
- [ ] Verify downgrade CI passes with the new `julia_version` parameter

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Krastanov:
it is normal that the benchmark is not running here, as we are changing the workflow file

Issue #231, opened 2023-02-23
Community detection algorithms


tecosaur:
There seems to be a dearth of community detection algorithms in the current Graphs.jl. Hopefully this can be remedied in the future, and to that end I thought I'd list the ones that NetworkX currently has for comparison:

- [ ] Kernighan–Lin bipartition
- [ ] K-Clique
- [ ] Greedy modular communities
- [ ] Lukes exact optimal weighted tree partitioning
- [ ] Label propagation
- [ ] Louvain Community Detection
- [ ] Fluid Communities
- [ ] Girvan–Newman

gdalle:
- [ ] Spectral clustering

gdalle:
@olegfafurin

leerosenthalj:
I am still interested in implementing Girvan-Newman! I need a week before I can make steady progress, but would be happy to take a crack at it.

gdalle:
Sounds good!

PR #502, opened 2026-03-08
ECG Community Detection implementation

ryandewolfe33:
Another community detection algorithm, see #231 

Ensemble Clustering for Graphs (ECG) uses many runs of Louvain (see #488 ) to improve performance and stability.

### DIFF
--- a/src/community/ecg.jl
+++ b/src/community/ecg.jl
@@ -1,0 +1,136 @@
+"""
+    ecg(g; γ=1, ensemble_size::Integer=16, min_edge_weight=0.05, distmx::AbstractArray{<:Number}=weights(g), max_moves::Integer=1000, max_merges::Integer=1000, move_tol::Real=10e-10, merge_tol::Real=10e-10, rng=nothing, seed=nothing)
+
+Community detection using ensemble clustering for graphs (ECG). Weights the edges based on the
+proportion of time the endpoints are in the same cluster of a Louvain without merges before running
+a final Louvain to detect communities.
+
+### Optional Arguments
+- `distmx=weights(g)`: distance matrix for weighted graphs
+- `ensemble_size=16`: the number of no merge Louvains in the ensemble
+- `min_edge_weight`: the minimum edge weight passed to the final Louvain (to retain the original topology).
+- `γ=1.0`: where `γ > 0` is a resolution parameter. Higher resolutions lead to more
+    communities, while lower resolutions lead to fewer communities. Where `γ=1.0` it
+    leads to the traditional definition of the modularity.
+- `max_moves=1000`: maximum number of rounds moving vertices before merging for each Louvain.
+- `max_merges=1000`: maximum number of merges in the final Louvain.
+- `move_tol=10e-10`: necessary increase of modularity to move a vertex in each Louvain.
+- `merge_tol=10e-10`: necessary increase of modularity in the move stage to merge in the final Louvain.
+- `rng=nothing`: rng to use for reproducibility. May only pass one of rng or seed.
+- `seed=nothing`: seed to use for reproducibility. May only pass one of rng or seed.
+
+### References
+- [Valérie Poulin and François Théberge. Ensemble Clustering for Graphs: Comparisons and Applications. Applied Network Science, 4:4 (2019)][https://doi.org/10.1007/s41109-019-0162-z]
+
+
+# Examples 
+```jldoctest
+julia> using Graphs
+
+julia> barbell = blockdiag(complete_graph(3), complete_graph(3));
+
+julia> add_edge!(barbell, 1, 4);
+
+julia> ecg(barbell)
+6-element Vector{Int64}:
+ 1
+ 1
+ 1
+ 2
+ 2
+ 2
+
+julia> ecg(barbell, γ=0.01)
+6-element Vector{Int64}:
+ 1
+ 1
+ 1
+ 1
+ 1
+ 1
+```
+"""
+function ecg(
+    g::AbstractGraph{T};
+    γ=1.0,
+    ensemble_size::Integer=16,
+    min_edge_weight::Real=0.05,
+    distmx::AbstractArray{<:Number}=weights(g),
+    max_moves::Integer=1000,
+    max_merges::Integer=1000,
+    move_tol::Real=10e-10,
+    merge_tol::Real=10e-10,
+    rng::Union{Nothing,AbstractRNG}=nothing,
+    seed::Union{Nothing,Integer}=nothing,
+) where {T}
+    rng = rng_from_rng_or_seed(rng, seed)
+    if nv(g) == 0
+        return T[]
+    end
+    ensemble_weights = ecg_weights(
+        g;
+        γ=γ,
+        ensemble_size=ensemble_size,
+        distmx=distmx,
+        max_moves=max_moves,
+        move_tol=move_tol,
+        rng=rng,
+    )
+    weights =
+        (1-min_edge_weight)*ensemble_weights +
+        min_edge_weight * adjacency_matrix(g, Float64)
+    return louvain(
+        g;
+        γ=γ,
+        distmx=weights,
+        max_moves=max_moves,
+        max_merges=max_merges,
+        move_tol=move_tol,
+        merge_tol=merge_tol,
+        rng=rng,
+    )
+end
+
+"""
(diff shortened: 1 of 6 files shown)


src/community/ecg.jl
```diff
@@ -0,0 +1,136 @@
...
+function ecg(
+    g::AbstractGraph{T};
+    γ=1.0,
+    ensemble_size::Integer=16,
+    min_edge_weight::Real=0.05,
+    distmx::AbstractArray{<:Number}=weights(g),
+    max_moves::Integer=1000,
+    max_merges::Integer=1000,
+    move_tol::Real=10e-10,
+    merge_tol::Real=10e-10,
```
LoveLow-Global: Just a small note, but could you change the `10e-10` values to `1e-9`? They represent the same number, but the latter is the standard scientific notation.

test/community/ecg.jl
```diff
@@ -0,0 +1,129 @@
...
+    for g in test_generic_graphs(empty)
+        r = @inferred ecg_weights(g)
+        @test c == r
+    end
+
+    # Undirected loops
+    loops = complete_graph(2)
+    add_edge!(loops, 1, 1)
+    add_edge!(loops, 2, 2)
+    c = sparse([
```
LoveLow-Global, line 42: I believe the maximum possible weight is `1.0` for this algorithm, in general. Therefore, I think the outputs being `2.0` is a bug, as the undirected check in `ecg_weights` currently double-counts self-loops.

src/community/ecg.jl
```diff
@@ -0,0 +1,136 @@
...
+    ensemble_weights = ecg_weights(
+        g;
+        γ=γ,
+        ensemble_size=ensemble_size,
+        distmx=distmx,
+        max_moves=max_moves,
+        move_tol=move_tol,
+        rng=rng,
+    )
+    weights =
```
LoveLow-Global, line 94: I noticed an algorithmic detail issue from the weight calculation. According to Equation (1) in the Poulin & Théberge paper (which I found on line 23 of this file), the ensemble co-association adjustment must only be applied to edges within the 2-core of the graph. Any edge outside the 2-core should default to $w_*$.

Currently, if I am not mistaken, it applies the interpolation globally to every edge.
I believe you have to conditionally apply this, using `core_number(g)`.

LoveLow-Global (changes requested):
Thanks for working on this! Adding ECG is a great feature for the community detection.

I've left a few comments. The two main things to check on are ensuring the ensemble weighting only applies to the 2-core of the graph  and a small bug with undirected self-loops.

As this is my first review after being a reviewer here, I may have not given the best review possible. I apologize in advance for this issue, please let me know if there is anything to be improved from my side.

src/community/ecg.jl
ryandewolfe33: Done :)

test/community/ecg.jl
ryandewolfe33: Your belief about the edge weights is correct. The diagonal of the adjacency matrix of an undirected graph holds twice the weight of a loop (so that the degree each vertex is the sum of the row/column).

It's only the case for undirected graphs, there is the same test on a directed graph (lines 78-85) where the diagonal is 1.0.

src/community/ecg.jl
ryandewolfe33: Thanks for catching this. Without this step, the implementation (and intuition) works nicely for weighted (parallel edges considered as weighted) and directed graphs, but the 2-core is not so obvious in these cases. Also the `core_number(g)` function currently does not support self loops.

My proposed (and now implemented) solution is to have a `min_weight_outside_2core` parameter that calls `core_number(g)` and decreases all edge weights with at least of vertex with core number < 2 to the minimum weight. This way it can be bypassed for any of the edge cases above, and will throw an error if the `core_number` function is not supported for the graph type.

However I'm not super happy with this solution, and would be open to more discussion about the edge cases.

test/community/ecg.jl
LoveLow-Global: You are right about the adjacency matrix for undirected graphs preserving the degree sum. My apologies for missing that. Thanks for the reply.

src/community/ecg.jl
LoveLow-Global: I think your currect solution is good enough at the moment, I personally cannot think of a better way. I did ask LLMs for any ideas on this, and it did reply as the following.

To keep the API clean, what if we replaced `min_weight_outside_2core` with a simple boolean flag, like `core_weighting::Bool = !is_directed(g)`?
This way:
1. It automatically skips the 2-core logic for directed graphs to avoid ambiguity.
2. For undirected graphs with self-loops (which currently breaks `core_number`), we can either document that `core_weighting=false `must be passed, or handle it internally by passing a loop-free copy of the graph to `core_number` just for the mask.

Please note that this is just a suggestion from a LLM. Your ideas and thoughts have the priority for sure. If you want to move forward with the currect approach, I will do some final checks and merge if there are no other problems.

ryandewolfe33: I think it's clearer to avoid silently apply/not apply the parameter depending on the input, so I'd prefer the current implementation. I've added a bit of documentation and an error message in the function that to better explain the parameter and allowable input combinations.

ryandewolfe33:
Thanks for the helpful review! It's only my ~3rd pull request so I don't feel qualified to say much about your process, but it seems good to me.

src/community/ecg.jl
LoveLow-Global: During my double-check, I noticed the function name `ecg_weights(g; γ, {and so on})` below does not match the `ensemble_weights(g; c, {and so on})` here in the docstring. Can you take a look on this please? Thank you.

LoveLow-Global (changes requested):

LoveLow-Global:
And thank you the kind comments, I really appreciate it! I will try to improve more in the future.

src/community/ecg.jl
ryandewolfe33: Yes thanks for catching. I've updated both doc strings; the style of ecg doc string was inconsistent with the rest of the community detection files.

LoveLow-Global (approved):

LoveLow-Global:
Thank you so much for the great contribution. I believe it is ready to be merged!

LoveLow-Global:
@Krastanov Hello, I wonder if you can check on the workflow before merging, as well as if it is ready to be merged, as it is my first time reviewing a PR and I want to make it sure. Thank you!

Krastanov:
Thanks both, for the contribution and for the review!

I want to briefly discuss to questions of API design before we merge it (it otherwise seems ready for merge):

- What are the pros and cons of `ensemble_clustering` vs `ecg` as names? `ensemble_clustering` seems a bit more legible to me, especially to someone outside of the field.
- Is the `*_weights` function meant to be public? Do other clustering algorithms implemented here have corresponding `*_weights` functions or something equivalent?

And a question for future work: Given that we now have a few different types of clustering algorithms, does it make sense to start discussing how these can be organized together, e.g. by some general `Clustering` return type and some `clustering(g, ::AlgorithmType)` API?

ryandewolfe33:
Hi Krastanov, thanks for the comments.

- My interpretation is that `ensemble_clustering` is more of a meta-clustering paradigm, whereas `ecg` is the name of this specific ensemble method (with louvain the specific reweighting). I could see a future with a generic ensemble method that uses a different or many internal algorithms.

- I'm not sure, I'd be open to changing it to private. I do know that the weights returned can be quite useful (in layouts for example, see [this notebook](https://github.com/ftheberge/graph-partition-and-measures/blob/master/ECG_extras.ipynb) for some ideas).

-------------------------

I think discussion around a standardized type makes sense, but I'm not sufficiently comfortable with Julia to lead that. I will say that everything right now uses arrays with entries corresponding to community id as the return type, which I think is fine for partitions (every vertex gets exactly one community). Most popular algorithms use partitions (louvain, leiden, infomap, label propagation, some spectral things) so I think it would be fine to standardize around a partition as a result.
 
There are generalizations with outliers (a vertex can have no communities), or overlaps (a vertex can have multiple communities), or fuzzy membership (e.g. 50% in each of two communities) where the return type would have to be different, but at that point it could make sense as it's own package.

I don't have plans for adding any more community detection algorithms right now, so maybe we can move this to an issue where it's more visible and better for a longer discussion.

Krastanov:
Thanks, Ryan! Thanks Jihyung for the review!

PR #517, opened 2026-07-23
Secure AirspeedVelocity PR benchmark workflow

Krastanov-agent:
## Summary
- run the AirspeedVelocity benchmark workflow on `pull_request` instead of `pull_request_target`
- keep the workflow token read-only with `contents: read`
- write benchmark results to the job summary via `job-summary: true` instead of posting PR comments
- keep the branch filter as `[master, main]` so this file can be reused across repos

This avoids the classic `pull_request_target` pwn request problem: the old workflow could execute PR-controlled benchmark/package code while holding base-repository privileges.

cc @Krastanov

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,16 +1,15 @@
 name: Benchmarks
+
 on:
-  pull_request_target:
+  pull_request:
     branches: [master, main]
+
 permissions:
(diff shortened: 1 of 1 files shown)


Krastanov:
The previous version of the benchmark CI script was dangerous -- fixing this so that the runner can not be misused by external contributors.

PR #518, opened 2026-07-31
Add KADABRA betweenness centrality approximation algorithm

DerSchmachtin:
### Description
This Pull Request introduces the **KADABRA** (K-ADaptive Approximation of Betweenness centRAlity) algorithm to `Graphs.jl`. 

Computing exact betweenness centrality can be prohibitively expensive on very large graphs. KADABRA provides a fast and highly scalable randomized approximation of the betweenness centrality, as well as an efficient method to find the top-k nodes with the highest betweenness. 

The implementation includes two main exported functions:
- `kadabra_centrality`: Approximates the betweenness centrality for all vertices.
- `kadabra_top_k`: Returns the top `k` vertices with the highest betweenness centrality.

**Reference:**
> Borassi M., Natale E. (2016) KADABRA is an ADaptive Algorithm for Betweenness via Random Approximation. In: ESA 2016. [https://dl.acm.org/doi/10.1145/3284359]

### A note on the top-k allocation

In top-k mode KADABRA sizes each vertex's confidence interval from its rank gaps before sampling begins. Here the implementation follows the **paper** (Section 5.2) rather than the authors' C++ reference, which NetworKit's `KadabraBetweenness` also copies — so a reviewer comparing against either will find a deliberate difference rather than a porting bug:

- The reference pairs the two rank gaps opposite to the way its own stopping test consults them, and leaves `lambda_L(v_1)` unconstrained — pinning the top vertex's lower budget, the only one its own test uses, at the minimum.
- Its tie-collapse rule is never applied to the pair `(v_k, v_{k+1})`, the single pair the external-exclusion test is built around, which deadlocks both vertices once their gap falls below `2*err`.

The allocation is a heuristic for *where* to spend the `delta` budget, not part of the `(err, delta)` guarantee, so each variant remains a valid approximation and they differ only in how many samples they need. Measured over 4 SNAP graphs × `k ∈ {3, 5, 10, 100}` × 3 seeds at `err = 1e-4`, the paper's allocation needed `0.80 ± 0.18` times as many samples as the reference's, at identical top-k accuracy — the same overlap with the exact ranking and the same Kendall τ over the top k to three decimals. Collapsing the boundary pair additionally removes the cases where a top-k query cost more than computing every centrality. See the number of samples in comparison to 'k=0' below. `k = 0` is unaffected: it does not enter this code path.

topk_allocation

I am separately seeking confirmation from the KADABRA authors on whether the reference's ordering was deliberate (I'm pretty sure it was overlooked), and will report back here. In the meantime I am happy to switch to the reference's behaviour instead if you would prefer bug-for-bug compatibility with NetworKit.

### Checklist
- [x] Added `kadabra.jl` to `src/centrality/`
- [x] Added corresponding test suite in `test/centrality/kadabra.jl` and registered it in `runtests.jl`
- [x] Code has been formatted using `JuliaFormatter.jl` according to BlueStyle
- [x] Added documentation strings and exported the functions in `Graphs.jl`
- [x] Ensured RNG scoping follows `Graphs.jl` conventions using standard library imports

All tests pass locally. Let me know if there are any changes or further optimizations you would like me to make!

### Benchmarks & Performance
To verify the performance, I ran some benchmarks comparing this Julia implementation against the original, highly optimized C++ version from the authors. 

**Hardware Setup:**
All benchmarks were executed on an **AMD Ryzen Threadripper 3960X 24-Core Processor** (3.80 GHz, 48 threads) with **125 GiB of RAM** running Ubuntu 22.04 LTS.

The Julia implementation is competitive compared to the Original C++ Implementation.
Here are the results (using `k=0`, `delta=0.1` and `epsilon=0.0001`) on a few test instances taken from SNAP:

image

### DIFF
--- a/src/centrality/kadabra.jl
+++ b/src/centrality/kadabra.jl
@@ -1,0 +1,1130 @@
+# kadabra.jl
+
+"""
+    estimate_diameter(g::AbstractGraph)
+
+Computes an upper bound on the diameter of the graph using the AllCCUpperBound technique
+(Borassi et al. 2015). Highly efficient for both directed and undirected graphs.
+"""
+function estimate_diameter(g::AbstractGraph)
+    n = nv(g)
+    n == 0 && return 0.0
+    n == 1 && return 0.0
+
+    # 1. Compute strongly/connected components
+    sccs = is_directed(g) ? strongly_connected_components(g) : connected_components(g)
+    n_components = length(sccs)
+
+    # Map each vertex to its component index
+    cc = zeros(Int, n)
+    for (i, component) in enumerate(sccs)
+        for v in component
+            cc[v] = i
+        end
+    end
+
+    # 2. Compute pivots for each component
+    # The pivot vertex is the vertex maximizing the sum of the out-degree and the in-degree.
+    pivots = zeros(Int, n_components)
+    for i in 1:n_components
+        component = sccs[i]
+        best_v = component[1]
+        best_deg = outdegree(g, best_v) + indegree(g, best_v)
+        for v in component
+            deg = outdegree(g, v) + indegree(g, v)
+            if deg > best_deg
+                best_v = v
+                best_deg = deg
+            end
+        end
+        pivots[i] = best_v
+    end
+
+    # 3. Compute SCC adjacency graph (DAG of components)
+    cc_adj = [Set{Int}() for _ in 1:n_components]
+    for u in 1:n
+        for v in outneighbors(g, u)
+            if cc[u] != cc[v]
+                push!(cc_adj[cc[u]], cc[v])
+            end
+        end
+    end
+
+    # 4. Helper BFS to compute eccentricity of pivot in its SCC.
+    ecc_dist = fill(-1, n)
+
+    function compute_ecc_in_scc(start::Int, backward::Bool)
+        fill!(ecc_dist, -1)
+        q = Int[]
+        push!(q, start)
+        ecc_dist[start] = 0
+
+        head = 1
+        while head <= length(q)
+            u = q[head]
+            head += 1
+
+            neighbors = backward ? inneighbors(g, u) : outneighbors(g, u)
+            for v in neighbors
+                if ecc_dist[v] == -1 && cc[v] == cc[u]
+                    ecc_dist[v] = ecc_dist[u] + 1
+                    push!(q, v)
+                end
+            end
+        end
+
+        return isempty(q) ? 0 : ecc_dist[q[end]]
+    end
+
+    # 5. Compute forward and backward eccentricities of pivots in their SCCs
+    ecc_f_pivots_scc = zeros(Float64, n_components)
+    ecc_b_pivots_scc = zeros(Float64, n_components)
+    for i in 1:n_components
+        ecc_f_pivots_scc[i] = compute_ecc_in_scc(pivots[i], false)
+        ecc_b_pivots_scc[i] = compute_ecc_in_scc(pivots[i], true)
+    end
+
+    # 6. DP to compute bounds across components (memoized DFS).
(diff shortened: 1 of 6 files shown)


LoveLow-Global:
Hi there, sorry for the delay in responding. I was very busy thoughout the last couple of weeks, and haven't had the time to look into the new PRs. I will try to review this within a week. Thank you!!

src/centrality/kadabra.jl
```diff
@@ -0,0 +1,1166 @@
...
+    absolute::Bool,
+    bet::Vector{Float64},
+    err_l::Vector{Float64},
+    err_u::Vector{Float64},
+)
+    n_tracked = length(top_k_nodes)
+
+    for i in 1:n_tracked
+        v = top_k_nodes[i]
+        # clamp fängt asynchrone Auslesefehler ab und schützt vor DomainErrors
```
LoveLow-Global: Hi, I believe this is German. Could you change this into English please?

src/centrality/kadabra.jl
```diff
@@ -0,0 +1,1166 @@
...
+- `parallel::Bool`: If true, sample with `Threads.nthreads()` tasks; if false, use a single sampling stream (default: true).
+- `rng::Union{AbstractRNG,Nothing}`: Seed source for reproducible runs (default: `nothing`, i.e. the global RNG). With `parallel=false` the output is bit-identical across runs with the same seed. With `parallel=true` each worker's sampling stream is seeded deterministically, but how many samples each worker contributes before the shared stopping condition fires depends on thread scheduling, so estimates vary slightly between runs (within the `err` guarantee). Use `parallel=false` when exact reproducibility is required.
+
+# Returns
+A `NamedTuple` with fields:
+- `centralities::Vector{Float64}`: the betweenness estimate per vertex.
+- `lower_bounds::Vector{Float64}`, `upper_bounds::Vector{Float64}`: per-vertex confidence-interval endpoints, scaled by the same `normalize` convention.
+- `n_samples::Int`: total shortest-path pairs sampled, including the burn-in phase.
+- `omega::Float64`: the worst-case sample budget implied by `err`, `delta`, and the estimated diameter.
+- `tau::Int`: the burn-in sample count.
```
LoveLow-Global: In the docstrings here, `omega::Float64` and `tau::Int` exists, but the returned NamedTuple doesn't seem to have them. I wonder if you intended to have omega and tau in the code for diagnosing, or just forgot to edit the docstrings. I don't have a clear recommendation here, but I do think keeping omega and tau may be slightly better.

LoveLow-Global:
Hello, sorry for the very long delay. I was quite busy with other work for a while.

I wonder if you used AI-generated code here, could you share if you used LLMs? The overall code looks good, however as far as I know, the code should be reviewed differently when there is a significant usage of LLMs in code generation, which I do not know well. [Please check this comment](https://github.com/JuliaGraphs/Graphs.jl/pull/482#issuecomment-3679369883)

DerSchmachtin:
No worries, thanks for taking a look!

Yes, I used Claude (Opus 4.8 and 5, via Claude Code) for a lot of the code. It was a series of sessions over a few months while working for an uni lab project on KADABRA. The agent had direct access to the original C++ reference implementation throughout, so it was working from an established baseline (where I found a couple of places that don't match the paper, which this implementation fixes).

A main part of the uni project was a comparison to a different algorithm for betweenness centrality, so I benchmarked them extensively and in the process looked pretty heavily at the code and validated the algorithm.
The two divergence points from the C++ version (top-k allocation and burn-in normalization) were deliberate decisions driven by those empirical measurements, both of which are documented in the PR description.

I'm also writing up a report on the comparison (deadline is the 30th of September). It's not public yet, but I can send it to you if it helps with the review.

PR #526, opened 2026-09-03
Throw error when calling `cycle_basis` on a directed graph

### LINKED ISSUE #524: [BUG] `cycle_basis` is implemented/documented only for undirected graphs, but silently accepts directed ones and may run indefinitely
**Description of bug**
`cycle_basis` is implemented/documented only for undirected graphs. When calling `cycle_basis` on a directed graph, it is silently accepted, but might run indefinitely.

**How to reproduce**
This will hang and most likely eventually be SIGKILLED:
```julia
using Graphs

function directed_cycle()
    g = SimpleDiGraph(3)
    add_edge!(g, 1, 2)
    add_edge!(g, 2, 3)
    add_edge!(g, 3, 1)
    return g
end

g = directed_cycle()
cycle_basis(g)
```

**Expected behavior**
Throwing an error on directed input. It would be fine if it still works on directed graph types that happen to be undirected, such as this one, which happens to give the correct result:
```julia
function undirected_triangle()
    g = SimpleDiGraph(3)
    for (a, b) in ((1, 2), (2, 3), (3, 1))
        add_edge!(g, a, b)
        add_edge!(g, b, a)
    end
    return g
end

g = undirected_triangle()
basis = cycle_basis(g) # [[1, 3, 2]]
```
but was able to detect the failure of undirectedness and then throw.

**Actual behavior**
For some inputs, it hangs and is eventually killed by the OS.

**Version information**
```
Julia Version 1.12.7
Commit 6d172b025e4 (2026-08-15 08:05 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: macOS (arm64-apple-darwin25.5.0)
  CPU: 12 × Apple M3 Pro
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, apple-m3)
  GC: Built with stock GC
Threads: 1 default, 1 interactive, 1 GC (on 6 virtual cores)
Environment:
  JULIA_CONDAPKG_BACKEND = Null
  JULIA_PYTHONCALL_EXE = /Users/simone/Documents/flock/flock-product/packages/sheepdog/.venv/bin/python3
  JULIA_EDITOR = code
```
 - `  [86223c79] Graphs v1.14.0`

**Additional context**
The source code was originally merged without a guard and without documenting the limitation to undirected graphs, even though it states to be inspired by NetworkX, which (now, haven't checked whether it did at the time) simply throws on directed graphs. The docstring was later amended to mention the restriction to undirected graphs, but a guard was never introduced.

simsurace:
Closes #524

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ We follow SemVer as most of the Julia ecosystem. Below you might see the "breaki
 - `is_articulation(g, v)` for checking whether a single vertex is an articulation point
 - The iFUB algorithm is used for faster diameter calculation and now supports weighted graph diameter calculation
 - ECG community detection algorithm
+- `cycle_basis` now throws an error when it is called on a directed graph
 
 ## v1.14.0 - 2026-02-26
 - **(breaking)** `neighbors`, `inneighbors`, and `outneighbors` now return an immutable `FrozenVector` instead of `Vector`
--- a/src/cycles/basis.jl
+++ b/src/cycles/basis.jl
@@ -30,8 +30,7 @@ julia> cycle_basis(g)
 ### References
 * Paton, K. An algorithm for finding a fundamental set of cycles of a graph. Comm. ACM 12, 9 (Sept 1969), 514-518. [https://dl.acm.org/citation.cfm?id=363232]
 """
-function cycle_basis(g::AbstractGraph, root=nothing)
-    is_directed(g) && throw(ArgumentError("cycle_basis is not defined for directed graphs"))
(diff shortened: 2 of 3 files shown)


henrik-wolf:
Would it make sense to use the `IsDirected` trait for this?

```julia
@traitfn function cycle_basis(g::AbstractGraph::(!IsDirected), root=nothing)
```

simsurace:
I wouldn't mind changing this. I guess the only implication is the type of error thrown?

henrik-wolf:
I think so. The `MethodError` is slightly less informative, but consistent with the rest of the package.

PR #507, opened 2026-03-30
feat : Add VNGraphs.jl stubs for chromatic_number and edge_chromatic_number

mahmudsudo:
Declares chromatic_number and edge_chromatic_number in Graphs.jl with helpful error hints directing users to install VNGraphs.jl for the fast C-backed very_nauty implementations.

Includes unit tests asserting the stubs throw the expected errors.
VNGraph pr : https://github.com/JuliaGraphs/VNGraphs.jl/pull/26

closes #448

Issue #448, opened 2025-07-24
A reliable idiomatic wrapper for the C library `very_nauty` [$400]


Krastanov:
# A reliable idiomatic `very_nauty` wrapper [$400]

The JuliaGraphs/VNGraphs.jl package already exists and provides a simple low-level wrapper for the C library. A lot of this work probably will not be directly in the Graphs.jl repository.

- [ ] Fast conversion and/or views between Graphs.jl types and the C structures of very_nauty (slow converters already exist for simple graphs)
- [ ] Complete Graph API support for the VNGraph types (so that VNGraph types can be used in all already existing Graph.jl algorithms that do not peek behind the API)
- [ ] Dispatch from operations defined in Graphs/GraphsMatching/GraphsOptim to very_nauty implementation. For instance if there is a pre-existing `Graphs.some_interesting_property(::AbstractGraph)` there should now be a new method defined in VNGraphs `some_interesting_property(g::AbstractGraph, ::VNAlgorithm)` that dispatches to the C implementation. It should convert the `g` argument to VNGraph if necessary.
- [ ] If there is an algorithm defined in very_nauty that does not exist yet in Graphs.jl, it should be declared in Graphs.jl (just a function with docs but no methods), together with an error hint that VNGraphs is necessary. 
- [ ] Proper tests and documentation.
- [ ] PRs on this topic have to be submitted with clean git histories and well compartmentalized for ease of review. 

**Required skills**: familiarity with the Graphs.jl API and understanding of the Julia-C interface

**Reviewer**: any Graphs.jl member with merge rights

**Duration**: 3 months

#### Payout procedure:

The Funding for these bounties comes from the National Science Foundation and from the NSF Center for Quantum Networks. The payouts are managed by the NumFOCUS foundation and processed in bulk once every two months. If you live in a country in which NumFOCUS can make payments, you can participate in this bounty program.

[Click here for more details about the bug bounty program.](https://github.com/QuantumSavory/.github/blob/main/BUG_BOUNTIES.md)

<details>
<summary><strong>Bug bounty logistic details</strong> (click to expand)</summary>

To claim exclusive time to work on this bounty either post a comment here or message [skrastanov@umass.edu](mailto:skrastanov@umass.edu) with:

- your name
- github username
- **(optional)** a brief list of previous pertinent projects you have engaged in

[...]

PR #529, opened 2026-09-07
Declare chromatic_number and edge_chromatic_number

singhharsh1708:
Declares `chromatic_number` and `edge_chromatic_number` in the Graphs.jl namespace with no methods, so that packages computing them share one name instead of each defining its own. This is the "declare in Graphs.jl any algorithm that very_nauty has and Graphs.jl does not" item of #448.

It applies Krastanov's review on #507 point by point, and the original PR and its idea are mahmudsudo's. Bare `function f end` declarations rather than methods that throw, so nothing is invalidated when a package extends them and no error hint is needed while they stay private. Not exported, no documentation page, and no new tests, since they are private. The signature is `chromatic_number(g, alg)`, and the docstring says that per-algorithm settings such as a time limit are fields of `alg` rather than arguments.

They are in a new `src/coloring.jl`, the file name Krastanov suggested on #507, rather than in `src/traversals/greedy_color.jl`. That file is pulled into the Traversals page by `@autodocs`, which renders private docstrings by default, so putting them there would have published exactly what he asked to keep out of the documentation. No docs page lists `coloring.jl` and `checkdocs` is `:public`, so the build neither renders nor complains about them. `clique_number` in `src/community/cliques.jl` uses the same `function f end` declaration form.

Verified that both names are defined, are not in `names(Graphs)`, have zero methods, carry a docstring and raise `MethodError` when called: `Aqua.test_all` passes 10 of 10, and JuliaFormatter 2.2.0 reports the file already formatted in blue style.

The docstring and the changelog name VNGraphs.jl as the first package intended to provide methods rather than the only possible one; it can only add them once this name exists, which is the step after JuliaGraphs/VNGraphs.jl#37.

This covers one checkbox of #448 rather than the whole bounty, so it does not close the issue. The rest of that work is in VNGraphs.jl, split one concern per PR, #33 through #37.

AI Disclosure: Claude assisted with the code. Testing and review by @singhharsh1708.


singhharsh1708:
Overlap I should flag rather than let you find: #528 declares external algorithms through a general mechanism, an `EXTERNAL_ALGORITHMS` registry plus a `@declare_external` macro plus a `MethodError` hint registered in `__init__`, which is much closer to the sketch in your #456 review than what this PR does. If that lands, the two declarations here become one `@declare_external` line each, and I am happy to refile them that way or to close this in its favour.

The one piece that does not carry over is the hint itself. Your review on #507 said to declare the name without an implementation and that there was no need to worry about adding error hints while it stays private, which is why there is none here and why nothing is exported; #528 registers hints and exports the name it declares. That looks like one decision worth making once for both PRs rather than separately in each.

